PR #253, opened 2014-03-27
Add inner_html method to elements

aderyabin:


titusfortner:
@aderyabin - how was this generated? I do not see a javascript fragment for it in Selenium.

PR #286, opened 2014-10-31
ensure driver is in desired context to check for stale elements

titusfortner:
If the driver context is inside a frame, it will fail on an assert_not_stale call, even if the element is actually there. Specs for this are here - https://github.com/watir/watirspec/pull/50

### DIFF
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -387,6 +387,7 @@ module Watir
         true
       end
     end
+    alias_method :assert_not_stale, :assert_exists
 
     def reset!
       # no-op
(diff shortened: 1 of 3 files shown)


titusfortner:
I just rebased and updated this code. The last submission worked, but felt messy. This should be much more straightforward.

Essentially Element#locate and Element#assert_not_stale are the two paths for #assert_exists, so each needs to traverse up the chain and then take actions on the way back down; including switching into the parent's frame context if the parent is an IFrame.

I removed the extra logic in the IFrame#assert_exists and a few other things that are redundant with this refactor, and tweaked a few things that should make them more clear.

jarib:
Thanks!

PR #291, opened 2014-11-15
Improve stale element handling

titusfortner:
Element#assert_not_stale as currently implemented is useful for more static interactions with the page (refreshes and the like)
With more dynamic updates to the DOM, we sometimes have elements going stale between the #assert_not_stale call and the wire call that uses the element object.

These updates: 
1) Reduce the number of places where wire calls are made directly and push the implementation to the Element superclass where possible. 
2) Re-look up the element if it is stale, but still exists. If it passes the #assert_not_stale, then becomes stale, there is no reason to not retry instead of letting the test fail.

### DIFF
--- a/lib/watir-webdriver/elements/checkbox.rb
+++ b/lib/watir-webdriver/elements/checkbox.rb
@@ -17,14 +17,7 @@ module Watir
     #
 
     def set(bool = true)
-      assert_exists
-      assert_enabled
-
-      if @element.selected?
-        @element.click unless bool
-      else
-        @element.click if bool
-      end
+      set? == bool ? assert_enabled : click
     end
 
     #
@@ -34,13 +27,11 @@ module Watir
 
     def set?
       assert_exists
-      @element.selected?
+      element_call { @element.selected? }
     end
 
     #
     # Unsets checkbox.
-    #
-    # Same as +set(false)+
     #
 
     def clear
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -84,7 +84,7 @@ module Watir
 
     def text
       assert_exists
-      @element.text
+      element_call { @element.text }
     end
 
     #
@@ -95,7 +95,7 @@ module Watir
 
     def tag_name
       assert_exists
-      @element.tag_name.downcase
+      element_call { @element.tag_name.downcase }
     end
 
     #
@@ -119,17 +119,19 @@ module Watir
       assert_exists
       assert_enabled
 
-      if modifiers.any?
-        assert_has_input_devices_for "click(#{modifiers.join ', '})"
+      element_call do
+        if modifiers.any?
+          assert_has_input_devices_for "click(#{modifiers.join ', '})"
 
-        action = driver.action
-        modifiers.each { |mod| action.key_down mod }
-        action.click @element
-        modifiers.each { |mod| action.key_up mod }
+          action = driver.action
+          modifiers.each { |mod| action.key_down mod }
+          action.click @element
+          modifiers.each { |mod| action.key_up mod }
 
-        action.perform
-      else
-        @element.click
+          action.perform
+        else
+          @element.click
+        end
       end
 
       run_checkers
@@ -147,7 +149,7 @@ module Watir
       assert_exists
       assert_has_input_devices_for :double_click
 
-      driver.action.double_click(@element).perform
+      element_call { driver.action.double_click(@element).perform }
       run_checkers
     end
 
@@ -163,7 +165,7 @@ module Watir
       assert_exists
       assert_has_input_devices_for :right_click
 
-      driver.action.context_click(@element).perform
+      element_call { driver.action.context_click(@element).perform }
       run_checkers
     end
 
@@ -179,7 +181,7 @@ module Watir
       assert_exists
       assert_has_input_devices_for :hover
 
-      driver.action.move_to(@element).perform
+      element_call { driver.action.move_to(@element).perform }
     end
 
     #
@@ -197,9 +199,11 @@ module Watir
       assert_exists
       assert_has_input_devices_for :drag_and_drop_on
 
-      driver.action.
-             drag_and_drop(@element, other.wd).
-             perform
+      element_call do
+        driver.action.
+               drag_and_drop(@element, other.wd).
+               perform
+      end
     end
 
     #
@@ -217,9 +221,11 @@ module Watir
       assert_exists
       assert_has_input_devices_for :drag_and_drop_by
 
-      driver.action.
-             drag_and_drop_by(@element, right_by, down_by).
-             perform
+      element_call do
+        driver.action.
+               drag_and_drop_by(@element, right_by, down_by).
+               perform
+      end
     end
 
     #
@@ -250,13 +256,9 @@ module Watir
     #
 
     def value
-      assert_exists
-
-      begin
-        @element.attribute('value') || ''
-      rescue Selenium::WebDriver::Error::InvalidElementStateError
-        ""
-      end
+      attribute_value('value') || ''
+    rescue Selenium::WebDriver::Error::InvalidElementStateError
+      ''
     end
 
     #
@@ -267,12 +269,12 @@ module Watir
     #   #=> "link_title_2"
     #
     # @param [String] attribute_name
-    # @return [String]
+    # @return [String, nil]
     #
 
     def attribute_value(attribute_name)
       assert_exists
-      @element.attribute attribute_name
+      element_call { @element.attribute attribute_name }
     end
 
     #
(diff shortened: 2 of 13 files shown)


lib/watir-webdriver/elements/element.rb
```diff
@@ -570,6 +585,15 @@ def assert_is_element(obj)
       end
     end
 
+    def element_call
+      yield
+    rescue Selenium::WebDriver::Error::ObsoleteElementError, UnknownObjectException
+      raise UnknownObjectException unless Watir.always_locate?
```
jarib: A few points here:

1. In what cases will the yield raise an `UnknownObjectException`? Is that rescue actually necessary?
2. It's always better to rescue `StaleElementReferenceError`, which is the [new name](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/error.rb#L191) for `ObsoleteElementError`. 
3. I don't think it's ideal to raise `UnknownObjectException` if `Watir.always_locate?` is false. We're "erasing" both the error message and the specificity of the `StaleElementReferenceError`, so the user has no way of knowing whether this was caused by an element going stale or if it never existed. This will break backwards compatibility for people (like me) who have code that relies on stale element exceptions. I think we should either (a) re-raise the original exception or (b) create a new Watir error that inherits from `StaleElementReferenceError` - thus ensuring backwards compatibility without leaking WebDriver errors.

lib/watir-webdriver/elements/element.rb
```diff
@@ -197,9 +213,9 @@ def drag_and_drop_on(other)
       assert_exists
       assert_has_input_devices_for :drag_and_drop_on
 
-      driver.action.
-             drag_and_drop(@element, other.wd).
-             perform
+      element_call { driver.action.
+          drag_and_drop(@element, other.wd).
+          perform }
```
jarib: Please use `do ... end` here.

lib/watir-webdriver/elements/element.rb
```diff
@@ -217,9 +233,9 @@ def drag_and_drop_by(right_by, down_by)
       assert_exists
       assert_has_input_devices_for :drag_and_drop_by
 
-      driver.action.
-             drag_and_drop_by(@element, right_by, down_by).
-             perform
+      element_call { driver.action.
+          drag_and_drop_by(@element, right_by, down_by).
+          perform }
```
jarib: Please use `do ... end` here.

titusfortner:
I agree with all of the suggested changes; thank you.
I added an always_locate_spec to clarify the 3 key differences between having that value turned on and off. 
I fixed up the spec for element_call to be more clear about the particular use case this PR addresses (element goes stale between assert_exists and the desired wire call).

jarib:
I think this is OK now since it's possible to turn off the new behaviour using `Watir.always_locate = false`. How do you feel @p0deje and @jarmo?

lib/watir-webdriver/elements/checkbox.rb
```diff
@@ -17,14 +17,7 @@ class CheckBox < Input
...
     def set(bool = true)
-      assert_exists
-      assert_enabled
-
-      if @element.selected?
-        @element.click unless bool
-      else
-        @element.click if bool
-      end
+      set? == bool ? assert_enabled : click
```
p0deje: I'm not sure I understand this change correctly.

1. When checkbox is set and user tries to set it - `#assert_enabled` is called.
2. When checkbox is set and user tries to unset it - `#assert_enabled` is not called.

Since previously `#assert_enabled` was always called, I believe this logic is not correct.

lib/watir-webdriver/elements/element.rb
```diff
@@ -197,9 +199,11 @@ def drag_and_drop_on(other)
       assert_exists
       assert_has_input_devices_for :drag_and_drop_on
 
-      driver.action.
-             drag_and_drop(@element, other.wd).
-             perform
+      element_call do
+        driver.action.
```
p0deje: Please, use the same style it was before:

```ruby
element_call do
  driver.action.
         drag_and_drop(@element, other.wd).
         perform
end
```

lib/watir-webdriver/elements/element.rb
```diff
@@ -217,9 +221,11 @@ def drag_and_drop_by(right_by, down_by)
       assert_exists
       assert_has_input_devices_for :drag_and_drop_by
 
-      driver.action.
-             drag_and_drop_by(@element, right_by, down_by).
-             perform
+      element_call do
+        driver.action.
```
p0deje: Same as up about style.

lib/watir-webdriver/user_editable.rb
```diff
@@ -23,10 +20,7 @@ def set(*args)
     #
 
     def append(*args)
```
p0deje: Since it just send_keys, I believe it can be just aliased to `#send_keys`.

spec/always_locate_spec.rb
```diff
@@ -0,0 +1,48 @@
+require File.expand_path('watirspec/spec_helper', File.dirname(__FILE__))
+
+describe '#always_locate?' do
```
p0deje: Let's change it to:

```ruby
describe 'Watir' do
  describe '#always_locate?` do
```

lib/watir-webdriver/elements/checkbox.rb
titusfortner: So normally I'd follow the same approach as in Radio:
click unless set?
This spec fails, because it wants to ensure enabled, even if we don't click anything:
https://github.com/watir/watirspec/blob/master/radio_spec.rb#L234

p0deje: Oh, I see now: `Element#click` checks if element is enabled too. Okay then.

titusfortner: and to clarify #click does an assert_enabled call

lib/watir-webdriver/elements/element.rb
titusfortner: So the one before definitely had the period in the wrong spot (bottom line instead of top). This one is ok, though, or is the issue something else? Thanks.

p0deje: Aligned differently

p0deje:
Since `Watir.always_locate` defaults to `true` these changes will be default for all people who don't know about it or never tried changing it. With that said, I think we can merge it as is but I'd like to spend more time testing it before release (probably will just use new version on my test suite and watch the results closely).

lib/watir-webdriver/elements/element.rb
titusfortner: Ah. That isn't how the style guide shows it, but I can do it that way.

lib/watir-webdriver/user_editable.rb
titusfortner: We can't alias because we have to require user_editable before elements/text_area & text_field

titusfortner:
Great suggestions. I rebased from master and made the changes you suggested. I've run it with the various combinations against our rather varied test suite at Blackbaud; but definitely let me know if you find any other issues.

This should only affect a relatively rare race condition:

1) An action is getting taken on an element (#exists? and #assert_exists & #assert_not_stale methods, which should be what are polled to determine staleness are not affected)
2) The element is looked up or verified not stale in the #assert_exists call
3) The element goes stale before the wire call happens
4) The wire call taking the action then throws a stale exception

We actually run with always_locate? == false as well right now, so we're probably going to have to switch to the default setting if we want to fix our intermittent failures as a result of this issue.

lib/watir-webdriver/elements/element.rb
p0deje: We don't have specific style guide - just preserving existing style :)

lib/watir-webdriver/user_editable.rb
p0deje: Oh, I see. That wasn't obvious .

titusfortner: Nope. That was a trial and error determination. :)

jarib:
@p0deje We'll definitely leave some time for you to test it out. We have to wait for a proper release of selenium-webdriver 2.45 anyway.

Thanks for the effort everyone, especially Titus for not getting frustrated when progress was slow! :+1:

jarib:
And let's do `0.7.0` for the next release, since it has quite a few changes. 

@titusfortner: Perhaps you could summarize the things you've done in `CHANGES.md`?

titusfortner:
@jarib - Thanks! No worries on the slowness; I appreciate your thorough responses more, whenever they happen. Plus in several cases it gave me enough time to figure out how I could do something better than what I'd originally submitted. :)

p0deje:
@jarib @titusfortner I've just run Toptal test suite against master and had few test failures. From quick glance, it looks like it's somehow related to insufficient waiting for element presence/visibility/staleness, but I'll need some time to figure this out.

p0deje:
@jarib @titusfortner So, after adding few valid waitings everything is back to normal. From my side, it looks great!

jarib:
Great.

PR #299, opened 2014-12-06
Allow frame collections to properly handle selectors

titusfortner:
Current implementation for FrameCollection & IFrameCollection uses ElementCollection#elements to obtain a subset of frames. This is being done properly, but #to_a takes the indexes of this subset and applies it to the overall list of frames it finds on the page. So, the current implementation only works if the matching frames are the first ones.

### DIFF
--- a/lib/watir-webdriver/elements/iframe.rb
+++ b/lib/watir-webdriver/elements/iframe.rb
@@ -55,11 +55,23 @@ module Watir
   class IFrameCollection < ElementCollection
 
     def to_a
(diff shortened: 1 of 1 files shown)


titusfortner:
So, the reason you can't  instantiate a Frame Collection with an element is that a Frame Collection is actually a collection of FramedDriver instances, and you can only create IFrame objects with just an element. There might be something that can be done along these lines in the future, but it would be a significant refactor.

For now, this solution is better than what we have currently and should be ok to merge as is.

jarib:
I've merged this + the related specs.

Issue #163, opened 2012-09-03
Inconsistencies locating textareas by text_field :value between ie and ff and chrome.


arthibald:
The following code returns "true, true" in ie but "true false" in ff and chrome:

``` ruby
require 'watir-webdriver'

b = Watir::Browser.new :ff # :chrome, :ie

b.goto 'http://www.velnetsupport.co.uk/parrots/FormMail/example_form.html'

b.text_field(:name => 'realname').set 'Tom Jones'
puts b.text_field(:value => /om\ Jon/x).exists?

b.text_field(:name => 'message').set 'John Jones'
puts b.text_field(:value => /ohn\ Jon/x).exists?

b.close
```

changing the second to last line to b.textarea instead of b.text_field produces all trues in these browsers. 

watir-webdriver-0.6.1
selenium-webdriver-2.25.0
ie9, firefox 15.0, chrome 21.0 on windows 7

PR #300, opened 2014-12-06
Loosen text_field type requirements

### LINKED ISSUE #295: Unable to set the value of an HTML5 type="date" input field
First off... I LOVE this gem :)

I am using Ruby 1.9.2 and watir-webdriver-0.6.11 on Centos 5.8. Does this version of watir-webdriver understand html5 date types? Watir seems to have no issues with type="email", but doesn't fare as well on type="date" tags.

I am trying to set a type="date" field using text_field, but watir-webdriver is unable to locate the input element.
```ruby
require 'watir-webdriver'
 
phantom = Watir::Browser.new(:phantomjs)
phantom.goto('http://postmaster.untd.com/blocked.html')
 
phantom.text_field(:name => 'organisation').set 'My Company'
phantom.text_field(:name => 'name').set 'John Smith'
phantom.text_field(:name => 'email').set 'jsmith@mydomain.com'
phantom.text_field(:name => 'email_unsubscribe').set 'bounce@mydomain.com'
phantom.textarea(:name => 'description').set 'bulk email sender'
phantom.textarea(:name => 'mail_del_failure').set 'blocked'
phantom.textarea(:name => 'ipaddresses').set '12.34.56.78'
```
However, when i try to do this:
```ruby
phantom.text_field(:name => 'txtDate').set Time.now.strftime("%m/%d/%Y")
```
It fails to locate the field:
```ruby
Watir::Exception::UnknownObjectException: unable to locate element, using {:name=>"txtDate", :tag_name=>"input or textarea", :type=>"(any text type)"}
```
The input field from the page source html looks like this:
```
<div class="form-group">
  <label for="" class="control-label">Date when the error was seen (mm/dd/yyyy) <span class="important">*</span></label>
  <input type="date" class="form-control" placeholder="mm/dd/yyyy" name="txtDate" maxlength="10">
</div>
```
If I ask my Watir instantiated phantomjs browser object if the input field in question exists, it tells me it does:
```ruby
irb(main):050:0> phantom.input(:name => 'txtDate').exists?
=> true
``` 
My goal here is set the date (in mm/dd/yyyy format) on the <input> in question, but I'm not sure how to do that with watir-webdriver.

Thanks!

p0deje:
This fixes #295 and also allows to use `#text_field` for a bunch of other input types.

@jarib I'm not pushing this directly since there might be some reason for you to restrict the types. Feel free to merge otherwise.

Specs are in watir/watirspec#53

### DIFF
--- a/lib/watir-webdriver/locators/text_field_locator.rb
+++ b/lib/watir-webdriver/locators/text_field_locator.rb
@@ -1,7 +1,7 @@
 module Watir
   class TextFieldLocator < ElementLocator
 
-    NON_TEXT_TYPES     = %w[file radio checkbox submit reset image button hidden datetime date month week time datetime-local range color]
+    NON_TEXT_TYPES = %w[file radio checkbox submit reset image button hidden range color]
     # TODO: better way of finding input text fields?
     NEGATIVE_TYPE_EXPR = NON_TEXT_TYPES.map { |type| "%s!=%s" % [XpathSupport.downcase('@type'), type.inspect] }.join(' and ')
 


p0deje:
I've just thought that maybe it's possible to generate the necessary methods from https://html.spec.whatwg.org/#htmlinputelement spec (see type attribute table). In some cases (e.g. Time/Date/Range) we can enforce related Ruby object as argument. In other cases, fallback to plain string.

What do you guys think?

titusfortner:
Why was the decision made to deprecate textarea from text_field? We wanted to restrict text_field usage to just input elements? Will this be a non-backward compatible change from the watir-classic API?

p0deje:
> Why was the decision made to deprecate textarea from text_field? We wanted to restrict text_field usage to just input elements?

Exactly.

> Will this be a non-backward compatible change from the watir-classic API?

Guess it should be done through watirspec.

titusfortner:
PR #392 makes this change to adopt the first option here. I want to continue the conversation here, though. From a user's standpoint I think we should be making using Watir more permissive not less. If you can type text into it, from a user's standpoint it is a text_field. 

I see the first option as the easiest to maintain, but the least user-friendly. I don't see a reason for the second and the third to be mutually exclusive. I believe we should continue to support text_field for text_area, and could also provide support for using the applicable Ruby objects in places.

p0deje:
Originally, we deprecated locating of textareas in `#text_field` due to inconsistencies in value/text (see #163) and because we duplicate the logic of existing `#textarea`. I still think we should remove this feature for the same reasons and also to make `#text_field` a bit more fast (no more tag name check).

Regarding this PR, I believe it's not hard to generate `*_field` methods from spec (unless we want to have custom argument types for them), but it might be confusing to the user to figure out the correct method to use from dozen of them. At the same time, having one `#input` is more elegant and consistent with other Watir API.

Or we can just loosen `#text_field` type restrictions just like this PR proposed.

p0deje:
Merged in 4233610.

PR #303, opened 2015-01-02
provide textarea deprecation warning when element located by id

titusfortner:
Current deprecation warning is short-circuited if located by id.


jarib:
Thanks!

PR #305, opened 2015-01-14
update changelog

titusfortner:
I think this is everything since 0.6.11. Anything important I missed?


jarib:
Looks good, thanks!

PR #307, opened 2015-01-15
Only re-look up elements when they aren't stale

titusfortner:
This essentially changes always_locate to locate_when_necessary

The Watir.always_locate? setting currently conflates whether to cache and whether to fail when stale. [Reading through](http://watirmelon.com/2011/08/02/watir-webdriver-obselete-element-error/) the [background conversations](https://groups.google.com/forum/#!topic/watir-general/WhwpPvzHWmM/discussion) for implementing always_locate, it seems the important issue is that it not throw an exception when the element goes stale. With my test suite I saw a performance improvement of 15-20%. So, huge performance bump without affecting the desired behavior.

### DIFF
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -495,16 +495,16 @@ module Watir
   protected
 
     def assert_exists
-      reset! if Watir.always_locate?
+      begin
+        assert_not_stale if @element ||= @selector[:element]
+      rescue UnknownObjectException => ex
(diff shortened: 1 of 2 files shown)


p0deje:
For the record, I've been trying this branch on my test suite and haven't noticed any performance increase. I guess I don't relookup elements at all :)

titusfortner:
@p0deje - Did you see a significant difference between #always_locate toggled on & off before this commit?

p0deje:
Neither as well.

PR #311, opened 2015-02-02
differentiate Element#visible? and Element#present?

titusfortner:


titusfortner:
This PR is a spec update only. I'm still figuring out git submodules; I might not want to be referencing a watirspec commit that isn't yet in master (The intent is for Travis CI to see and run it). Even if this works, I expect I will need to rebase it before merging.

p0deje:
@titusfortner I believe you can just push such things on your own, we usually use pull requests for things that require discussion/review.

titusfortner:
Hey Alex, thanks. I was using it to better understand how the process we use works with coveralls and travis ci and the watirspec submodule. I'll see about getting watirspec access as well to facilitate commits like these.

p0deje:
I’ve added you full access to watirspec.

I also think that there is no point in having separate teams for different repos, but that’s a topic to discuss at watir-development.=

titusfortner:
close/open to fix watirspec submodule pointing issue

titusfortner:
trying again

titusfortner:
Well, it won't let me reopen since I did a force push on the branch in my local repo (good to know I guess). So, I just merged it instead.

PR #312, opened 2015-02-04
change behavior of a wait with a timeout of zero

titusfortner:
This fixes [this failing spec](https://github.com/watir/watirspec/pull/60)


p0deje:
This seems correct, but I'm thinking it should be fixed in `Watir::Wait`. For example, this change should be done by any other custom timer implementation (e.g. https://github.com/p0deje/watir-timecop) + I feel like this is not timer responsibility.

titusfortner:
@p0deje - Is this what you were thinking? I updated the specs as well.

PR #317, opened 2015-02-17
addresses race conditions with slow loading frames

titusfortner:
For #299 we supported selectors for frames by doing a lookup of all the frame elements and all the frame elements matching the provided selector. I'm hitting a race condition where the frame hasn't finished loading on the page. With the current code the all_elements variable returns a size of zero, but the ElementCollection#elements call returns with a size of one.  

The subset of matching iframes should never be greater than all the iframes on the page.

### DIFF
--- a/lib/watir-webdriver/elements/iframe.rb
+++ b/lib/watir-webdriver/elements/iframe.rb
@@ -48,6 +48,7 @@ module Watir
   class IFrameCollection < ElementCollection
 
     def to_a
+      element_indexes = elements.map { |el| all_elements.index(el) }
       element_indexes.map { |idx| element_class.new(@parent, tag_name: @selector[:tag_name], :index => idx ) }
     end
 
@@ -57,14 +58,12 @@ module Watir
 
     private
 
-    def element_indexes
-      all_elements = locator_class.new(
+    def all_elements
+      locator_class.new(
           @parent.wd,
-          {tag_name: @selector[:tag_name]},
+          { tag_name: @selector[:tag_name] },
           element_class.attribute_list
       ).locate_all
-
-      elements.map { |el| all_elements.index(el) }
     end
 
   end # IFrameCollection


p0deje:
I do not 100% understand what's going, can you somehow show example to reproduce the problem or write a spec for it?

From what I got, case when frame hasn't finished loading (i.e. not in DOM?) should be handled by test, isn't it?

titusfortner:
For example, my test does something like: 
browser.table(id: 'table_id').iframes.find { |iframe| iframe.element(id: 'element_id').exists? }

The iframe is inside a wysiwyg element that takes a long time to fully load.
The code as it is currently written:
1) searches frames matching the selector
2) searches for all frames
3) maps the arrays to assign the index value of the matching frames to entire list of frames

When the component loads quickly, both #1 & #2 steps result in an array with a single iframe and everything maps as expected

When the component loads very slowly, both #1 & #2 steps result in empty arrays because the iframe is not found, yet. This can be dealt with in the test code using when present logic.

The problem is when the wysiwyg hasn't yet finished loading in step #1 (so empty array), but finishes loading in before step #2 (array with single element). 

when all_elements == []
then
elements.map { |el| all_elements.index(el) } => [nil]

Which results in the to_a method returning an iframe with an index of nil, which then causes errors in the lookup later. There is no easy way to deal with this in just the test code. I can't access the selector of the element to see if an index key is defined and verify it isn't nil.

Since the subset of the set should never be larger than the set, it makes sense to address the issue here. If the condition does not apply, no looping happens and everything is as it was.

p0deje:
Ok, I now understand the case, but I believe we can do something like this:

```diff
diff --git a/lib/watir-webdriver/elements/iframe.rb b/lib/watir-webdriver/elements/iframe.rb
index 5caed9f..2d305ff 100644
--- a/lib/watir-webdriver/elements/iframe.rb
+++ b/lib/watir-webdriver/elements/iframe.rb
@@ -64,7 +64,12 @@ module Watir
           element_class.attribute_list
       ).locate_all

-      elements.map { |el| all_elements.index(el) }
+      # In case first lookup `all_elements` returns empty array,
+      # but second lookup `#elements` returns non-empty empty array
+      # (i.e. frame has loaded between these two calls),
+      # index will return nil. This is a race condition that should
+      # be gracefully handled.
+      elements.map { |el| all_elements.index(el) }.compact
     end

   end # IFrameCollection
```

Or actually locate all_elements inside map:


```diff
diff --git a/lib/watir-webdriver/elements/iframe.rb b/lib/watir-webdriver/elements/iframe.rb
index 5caed9f..b5bded6 100644
--- a/lib/watir-webdriver/elements/iframe.rb
+++ b/lib/watir-webdriver/elements/iframe.rb
@@ -62,9 +62,14 @@ module Watir
           @parent.wd,
           {tag_name: @selector[:tag_name]},
           element_class.attribute_list
-      ).locate_all
-
-      elements.map { |el| all_elements.index(el) }
+      )
+
+      # In case first lookup `all_elements` returns empty array,
+      # but second lookup `#elements` returns non-empty empty array
+      # (i.e. frame has loaded between these two calls),
+      # index will return nil. This is a race condition that should
+      # be gracefully handled.
+      elements.map { |el| all_elements.locate_all.index(el) }
     end

   end # IFrameCollection
```

I believe those two implementations will have the same result but won't require introducing additional `begin ... while` loop and 2 additional wire calls (i.e. calling `#size` on each of elements set).

p0deje:
We also need to come up with spec for this (guess it can only be implemented with stubs).

titusfortner:
Oh good, I like both of those better than the while loop, especially the 2nd one.
I'll write up a spec for it and push it tomorrow.

Thanks!

titusfortner:
Refactor is slightly different from your example, but the same idea. Made it easier to write the test implementing it this way. Let me know if both of these pull requests work for you, and if you wouldn't mind, let me make sure I can push directly to the project and update the submodule properly.

p0deje:
Changes seem great! I'm leaving this as is so you can push directly.

Issue #211, opened 2013-07-19
can not return false when I using exist?


william99wallace:
I find my issue has already been raised:
https://github.com/watir/watir-webdriver/issues/152

but I didn't see the solution,because dear @jarib can not reproduced the issue
so, let me continue 
my code :
b.frame(:id ,"xxx").button(:id, "xxx").exists? 
it will retrun Error Message just like Selenium::WebDriver::Error::StaleElementReferenceError: switchFrame execution failed;Element does not exist in cache
but if my code is 
b.button(:id, "xxx").exists? 
it can retrun "false"！！and that's all I need！！！I just need a "false"！！
 I don't know why it knows the element does not exist, and didn't bring me "false"...
I know it can bring me "true"...I think it's something wrong cause by frame
sorry about my english anyway...

PR #321, opened 2015-03-11
Ensure not stale

titusfortner:
#286 fixed the issue with #211, but introduced a bug with a race condition for stale elements with multiple hierarchy elements. When the elements go stale, the current code does not re-lookup the top level element when Watir.always_locate? == true, and does not reset the entire chain of elements (only the top element) when Watir.always_locate? == false. This fixes the submitted specs and @bret made my previous implementation more readable and made the previous assert_not_stale better parallel the behavior or assert_exist.

### DIFF
--- a/spec/element_spec.rb
+++ b/spec/element_spec.rb
@@ -75,6 +75,20 @@ describe Watir::Element do
 
       expect(browser.div(:id, 'text')).to_not exist
     end
+
+    it "returns appropriate value when an ancestor element becomes stale" do
+      stale_element = browser.div(id: 'top').div(id: 'middle').div(id: 'bottom')
+      expect(stale_element.present?).to be true # look up and store @element for each element in hierarchy
+
+      grandparent = stale_element.instance_variable_get('@parent').instance_variable_get('@parent').instance_variable_get('@element')
+
+      # simulate element going stale during lookup
+      allow(grandparent).to receive('enabled?') { raise Selenium::WebDriver::Error::ObsoleteElementError }
+
+      browser.refresh
+      expect(stale_element.present?).to be Watir.always_locate?
+      expect(stale_element.present?).to be true
+    end
   end
 
   describe "#element_call" do
--- a/spec/element_spec.rb
+++ b/spec/element_spec.rb
@@ -87,7 +87,6 @@ describe Watir::Element do
 
       browser.refresh
       expect(stale_element.present?).to be Watir.always_locate?
-      expect(stale_element.present?).to be true
     end
   end
 
--- a/spec/html/removed_element.html
+++ b/spec/html/removed_element.html
@@ -13,5 +13,12 @@
   <body>
     <button id='remove-button' onclick="removeText()">Remove text</button>
     <div id="text">Text</div>
+
+    <div id='top'>Top Element>
+      <div id='middle'>Middle Element>
+        <div id='bottom'>Bottom Element</div>
+      </div>
+    </div>
+
   </body>
 </html>
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -387,7 +387,7 @@ module Watir
         true
       end
     end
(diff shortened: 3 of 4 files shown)


spec/element_spec.rb
```diff
@@ -75,6 +75,20 @@
 
       expect(browser.div(:id, 'text')).to_not exist
     end
+
+    it "returns appropriate value when an ancestor element becomes stale" do
+      stale_element = browser.div(id: 'top').div(id: 'middle').div(id: 'bottom')
+      expect(stale_element.present?).to be true # look up and store @element for each element in hierarchy
+
+      grandparent = stale_element.instance_variable_get('@parent').instance_variable_get('@parent').instance_variable_get('@element')
```
p0deje: Can't we just call `#parent`?

spec/element_spec.rb
```diff
@@ -75,6 +75,20 @@
...
+
+      # simulate element going stale during lookup
+      allow(grandparent).to receive('enabled?') { raise Selenium::WebDriver::Error::ObsoleteElementError }
+
+      browser.refresh
+      expect(stale_element.present?).to be Watir.always_locate?
+      expect(stale_element.present?).to be true
```
p0deje: These expectation seems so weird. Why so?

spec/html/removed_element.html
```diff
@@ -13,5 +13,12 @@
   <body>
     <button id='remove-button' onclick="removeText()">Remove text</button>
     <div id="text">Text</div>
+
+    <div id='top'>Top Element>
+        <div id='middle'>Middle Element>
```
p0deje: HTML code is indented with 2 spaces, not 4.

spec/element_spec.rb
titusfortner: Yeah, we have 2 different things named parent.

@parent, is a private instance variable of Watir::Element denoting the element/frame/browser that is the ancestor.
Watir::Element#parent is a method that calls execute_atom :getParentElement

It would be great to name the private one differently, like ancestor or something, but I didn't want to go there for this.

spec/element_spec.rb
titusfortner: It has to do with the fact that Watir.always_locate? == false is weird. :)
The current behavior/expectation for calling #exist? on a stale element in that condition is to return false the first time and true the second time.

p0deje: Okay, I feel like I'm totally lost. Why should it be `false` in the first place and `true` in the second?

spec/html/removed_element.html
titusfortner: Thanks, I updated.

spec/element_spec.rb
titusfortner: When Watir.always_locate? is false, and an element exists, but is stale... Calling #exist? on the element will return false, and in the process nil out the cached @element instance variable. Because it nils it out, subsequent calls will re-look it up and return true.

p0deje: Shouldn't calling `exists?` on stale element raise `Selenium::WebDriver::Error::StaleElementReferenceError` if `Watir.always_locate?` is false (or it's raised only when element is interacted with)?

p0deje: @titusfortner Ping me in IRC, maybe it will be faster to just discuss there.

titusfortner: Is this the only blocker to merging this? We can remove this line it if it is more confusing than helpful. :)

bret: If we remove this line, can we also remove all the code that is there to support it? That is where the real confusion is.

titusfortner: My last commit here removes all the extra reset! calls. All of the previous tests are passing. So, we can consider the previous behavior to be undocumented and unnecessary. We have the Watir.always_locate? turned on, so this won't affect us at Blackbaud, either way.

p0deje:
:+1: feel free to merge.

PR #325, opened 2015-03-24
Fix issue with rest-client update with running watir-spec

titusfortner:
Looks like a new update to this gem might be breaking our tests. Doing a PR to verify, will un-comment this line as the fix if this PR fails on Travis.

### DIFF
--- a/watir-webdriver.gemspec
+++ b/watir-webdriver.gemspec
@@ -32,5 +32,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "activesupport", "~> 3.0" # for pluralization during code generation
   s.add_development_dependency "pry"
   s.add_development_dependency "coveralls"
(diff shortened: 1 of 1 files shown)


titusfortner:
Previous CI build with failing:
https://travis-ci.org/watir/watir-webdriver/builds/55699168

Gives this error:
/home/travis/build/watir/watir-webdriver/spec/watirspec/lib/server.rb:40:in `run!': undefined method `run' for HTTP:Module (NoMethodError)

Then times out after 50 minutes trying to start the server

p0deje:
:+1:

titusfortner:
I don't understand our spec implementation code. Do we need to change something to make it compatible with the latest rest-client version?

p0deje:
It's being done with Sinatra - I look into this.

p0deje:
Fixed in master.

Issue #324, opened 2015-03-20
Drop support of Ruby 1.8 and 1.9


p0deje:
Both are dead. My suggestion is to update the code for 2.0 standards and update Travis configuration to run the specs on Ruby >= 2.0.

@watir/watir @watir/watir-webdriver-team What do you think?

zeljkofilipin:
:+1:

titusfortner:
Makes sense to me.
What code would need to be updated?

p0deje:
Mostly style changes (e.g. hashrockets/lambda).

I also suggest to change old watir location args (two args) in favor of hashes in watirspec:

```ruby
browser.div(:id, 'one') # old
browser.div(id: 'one')  # new
```

titusfortner:
Then I'm completely in favor of this. As someone who started using Ruby with 1.9 the (:id, 'one') is hard to look at. :)

zeljkofilipin:
RuboCop can convert hashes from old syntax to the new one automatically.

p0deje:
Yeah, I was planning to use Rubicon for that.

zeljkofilipin:
@p0deje rubicon?

p0deje:
@zeljkofilipin Sorry, didn't get it

p0deje:
@zeljkofilipin Oh, that was some weird autocorrect. It's "Rubocop" :)

p0deje:
Work has started in `ruby_2` branch. I'm planning to migrate watir-webdriver first, after that switch to watirpsec.

titusfortner:
Fixed by #327

PR #327, opened 2015-03-30
Drop support of Ruby 1.8 and Ruby 1.9

p0deje:
Per conversation in #324

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 rvm:
-  - 1.9.3
-  - 2.0.0
+  - 2.0
+  - 2.1
 notifications:
   recipients:
     - p0deje@gmail.com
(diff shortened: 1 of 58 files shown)


p0deje:
@titusfortner Guess that's pretty much it, can you go over the changes and let me know if you're fine with them. I'm not particularly sure we need `%i` shorthand, but since we're using `%w`, I've decided to do the same.

Blackbaud-TitusFortner:
Excellent work, I'm excited to be able to use 2.0 syntax. I want to run this against our suite of tests just to verify there isn't an issue that our watirspecs don't catch. I'll do that today.

p0deje:
@Blackbaud-TitusFortner Let me know when you're done, only waiting for it to merge.

titusfortner:
Passed all my tests, looks great. Do we need to run both 2.0 & 2.1 on Travis, or is there any down side to checking both?

p0deje:
I think having both `2.0` and `2.1` is good while `2.0` is still supported.

PR #331, opened 2015-04-03
Optimize regexp selectors when using xpath

marxarelli:
Regexp selectors are highly inefficient as they require additional round
trip requests (to retrieve attribute values) for each element that
matches the simplified selector. This optimization limits the number of
requests needed by converting leading and trailing runs of literals in
each regular expression to to an xpath `contains` predicate. Note that
problematic regular expressions such as case insensitive patterns and
those containing '|' are not converted.

### DIFF
--- a/lib/watir-webdriver/locators/element_locator.rb
+++ b/lib/watir-webdriver/locators/element_locator.rb
@@ -17,6 +17,17 @@ module Watir
 
     WILDCARD_ATTRIBUTE = /^(aria|data)_(.+)$/
 
+    # Regular expressions that can be reliably converted to xpath `contains`
+    # expressions in order to optimize the locator.
+    #
+    CONVERTABLE_REGEXP = %r{
+      \A
+      ([^\[\]\\^$.|?*+()]*) # leading literal characters
+      [^|]*?                # do not try to convert expressions with alternates
+      ([^\[\]\\^$.|?*+()]*) # trailing literal characters
+      \z
+    }x
+
     def initialize(wd, selector, valid_attributes)
       @wd               = wd
       @selector         = selector.dup
@@ -146,6 +157,15 @@ module Watir
         raise Error, "internal error: unable to build WebDriver selector from #{selector.inspect}"
       end
 
+      if how == :xpath && can_convert_regexp_to_contains?
+        rx_selector.each do |key, value|
+          next if key == :tag_name || key == :text
+
+          predicates = regexp_selector_to_predicates(key, value)
+          what = "(#{what})[#{predicates.join(' and ')}]" unless predicates.empty?
+        end
+      end
+
       elements = parent.find_elements(how, what)
       elements.__send__(method) { |el| matches_selector?(el, rx_selector) }
     end
@@ -230,6 +250,22 @@ module Watir
       end
 
       rx_selector
+    end
+
+    def regexp_selector_to_predicates(key, re)
+      return [] if re.casefold?
+
+      match = re.source.match(CONVERTABLE_REGEXP)
+      return [] unless match
+
+      lhs = lhs_for(key)
+      match.captures.reject(&:empty?).map do |literals|
+        "contains(#{lhs}, #{XpathSupport.escape(literals)})"
+      end
+    end
+
+    def can_convert_regexp_to_contains?
+      true
     end
 
     def assert_valid_as_attribute(attribute)
--- a/lib/watir-webdriver/locators/text_area_locator.rb
+++ b/lib/watir-webdriver/locators/text_area_locator.rb
@@ -3,6 +3,10 @@ module Watir
 
     private
 
+    def can_convert_regexp_to_contains?
+      false
+    end
+
     def normalize_selector(how, what)
       # We need to iterate through located elements and fetch
       # attribute value using WebDriver because XPath doesn't understand
--- a/spec/element_locator_spec.rb
+++ b/spec/element_locator_spec.rb
@@ -192,7 +192,7 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, "div").and_return(elements)
         else
-          expect_all(:xpath, ".//div").and_return(elements)
+          expect_all(:xpath, "(.//div)[contains(@class, 'oob')]").and_return(elements)
         end
 
         expect(locate_one(tag_name: "div", class: /oob/)).to eq elements[1]
@@ -207,7 +207,7 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, "div").and_return(elements)
         else
-          expect_all(:xpath, ".//div").and_return(elements)
+          expect_all(:xpath, "(.//div)[contains(@class, 'foo')]").and_return(elements)
         end
 
         selector = {
@@ -260,9 +260,8 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, 'div[dir="foo"]').and_return(elements)
         else
-          expect_all(:xpath, ".//div[@dir='foo']").and_return(elements)
+          expect_all(:xpath, "(.//div[@dir='foo'])[contains(@title, 'baz')]").and_return(elements)
         end
-
 
         selector = {
           tag_name: "div",
@@ -282,7 +281,7 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, 'div').and_return(elements)
         else
-          expect_all(:xpath, ".//div").and_return(elements)
+          expect_all(:xpath, "(.//div)[contains(@data-automation-id, 'bar')]").and_return(elements)
         end
 
 
@@ -316,7 +315,6 @@ describe Watir::ElementLocator do
         expect_all(:tag_name, "label").and_return([])
         expect(locate_one(tag_name: "div", label: /foo/)).to be_nil
       end
-
     end
 
     it "finds all if :index is given" do
@@ -432,7 +430,7 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, "div").and_return(elements)
         else
-          expect_all(:xpath, ".//div").and_return(elements)
+          expect_all(:xpath, "(.//div)[contains(@class, 'oob')]").and_return(elements)
         end
 
         expect(locate_all(tag_name: "div", class: /oob/)).to eq elements.last(3)
@@ -448,7 +446,7 @@ describe Watir::ElementLocator do
         if Watir.prefer_css?
           expect_all(:css, 'div[dir="foo"]').and_return(elements)
         else
-          expect_all(:xpath, ".//div[@dir='foo']").and_return(elements)
+          expect_all(:xpath, "(.//div[@dir='foo'])[contains(@title, 'baz')]").and_return(elements)
         end
 
         selector = {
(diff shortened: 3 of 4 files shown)


Blackbaud-TitusFortner:
Interesting, I'm not that familiar with regex in xpath, but we use a lot of it in our test suite. 

What is the reason for not doing buttons and text_areas?

I'll pull this commit and run it on my local tests to see how it fares.

marxarelli:
AFAIK, support for regular expressions directly in the xpath expression would vary by browser—only xpath 2.0 supports them via the [`matches` function](http://www.w3.org/TR/xpath-functions/#func-matches). For that reason, I opted to just reduce runs of literal characters at the beginning and end of the regexp to `contains` expressions—supported by xpath 1.0 and up. It doesn't fully approximate the regexp selector but rather reduces the set of elements that need to be further tested by the regexp directly.

It's mainly to optimize for cases where there are many elements to test, such as `browser.a(href: /^foo/)` on a page that contains many `a` elements. Instead of having to retrieve and test attributes for each element selected by `.//a`, it only has to do so for those selected by `.//a[contains(@href, 'foo')]`.

I couldn't figure out how to cleanly implement the optimization for buttons because of the complexity of the selector that it generates—it would need to match both by a `button`'s text value and an `input`'s `value` attribute. I forget why I disabled it for textareas. I may have just gotten lazy. :)

p0deje:
That is a pretty neat trick, though we have to keep an eye on the passed regex very carefully, handling all the possible special chars.

I think this really needs a lot of specs to ensure nothing is broken in the lookup. I'd love to see that PR merged, we just need to do an extensive testing covering everything around the Ruby regex implementation.

Also, see my comments to code.

Regarding the `matches`, as @marxarelli mentioned it's only available in XPath 2.0 which is supported by no browsers at the moment, so we can't use it.

lib/watir-webdriver/locators/text_area_locator.rb
```diff
@@ -3,6 +3,10 @@ class TextAreaLocator < ElementLocator
 
     private
 
+    def convert_regexp?
```
p0deje: Let's rename it to `can_convert_regexp_to_contains?`.

lib/watir-webdriver/locators/element_locator.rb
```diff
@@ -233,6 +242,21 @@ def delete_regexps_from(selector)
       rx_selector
     end
 
+    def regexp_selector_to_predicates(key, re)
+      return [] if re.casefold?
+
+      # convert the head and trailing runs of literals to `contains` expressions
+      match = re.source.match(/^([^\[\]\\^$.|?*+()]*)[^|]*?([^\[\]\\^$.|?*+()]*)$/)
```
p0deje: I believe it makes sense to:

1. Extract this regexp to consant/method.
2. Split the regexp to multiple lines with comments for each meaningful part.
3. Use `\A\z` instead of `^$`

lib/watir-webdriver/locators/element_locator.rb
```diff
@@ -147,6 +147,15 @@ def wd_find_by_regexp_selector(selector, method = :find)
         raise Error, "internal error: unable to build WebDriver selector from #{selector.inspect}"
       end
 
+      if how == :xpath && convert_regexp?
+        rx_selector.each do |key, value|
+          unless key == :text || key == :tag_name
```
p0deje: Avoid using unless with 2 or more conditions.

```ruby
rx_selector.each do |key, value|
  next if key == :text || key == :tag_name
  # code
end
```

lib/watir-webdriver/locators/element_locator.rb
```diff
@@ -233,6 +242,21 @@ def delete_regexps_from(selector)
       rx_selector
     end
 
+    def regexp_selector_to_predicates(key, re)
+      return [] if re.casefold?
```
p0deje: I'm not exactly sure, but I believe case-insensitivity can be achieved by using [translate()](https://github.com/watir/watir-webdriver/blob/master/lib/watir-webdriver/xpath_support.rb#L15-17].

p0deje:
After thinking more about that, I believe we can do really cool stuff like this:

```ruby
browser.div(class: /foo/)        #=> //div[contains(@class, 'foo')]
browser.div(class: /foo|bar/)    #=> //div[contains(@class, 'foo') or contains(@class, 'bar')]
browser.div(class: /(foo|bar)z/) #=> //div[contains(@class, 'fooz') or contains(@class, 'barz')]
browser.div(class: /fooz?/)      #=> //div[contains(@class, 'foo') or contains(@class, 'fooz')]
```

This can be extended to a very big list, though in some cases (e.g. `\d` or `\w`) we would probably face browser limitations to XPath expression length :)

In any way, I think the whole `regex -> xpath` engine worths separate class (to simplify testing) or even a gem (something like [regexp-examples](https://github.com/tom-lord/regexp-examples)).

lib/watir-webdriver/locators/element_locator.rb
p0deje: Text is already handled with `normalize-space()`, so why can't we use it?

p0deje: `:tag_name` can be handled with `self::` in some cases:

```
//*[self::span or self::div]
```

marxarelli:
I'm a little hesitant to implement those more complex conversions simply because I think the margin of error is going to be pretty wide. This was mean purely as an optimization to avoid WebDriver round trips in cases where a really basic regexp is being used.

I'll go ahead and implement your other suggestions and we can go from there.

lib/watir-webdriver/locators/element_locator.rb
marxarelli: I'm not sure why I excluded matching `:text`, actually. I'll have another look.

lib/watir-webdriver/locators/element_locator.rb
marxarelli: Originally I did that but then realized `XpathSupport.downcase` only works for ASCII characters. Since this is an optimization, I didn't want to tackle too many edge cases at the risk of introducing bugs.

marxarelli:
In addition to your suggestions, I've implemented a feature flag that's opt in for now. Perhaps once this sees more use in the real world, it can easily be changed to opt out.

p0deje:
Great work! I still believe we should work on regex -> xpath conversion engine, but I agree the current changes are already valuable.

My vote is :+1: to get this merged and released along with other changes (as soon as selenium-webdriver 2.45.1 is released). Waiting for @titusfortner thumbsup.

@marxarelli Thanks a lot,

titusfortner:
My suite passed with this commit, but I hit some performance issues that might not be related to this PR. I'm going to run another test tomorrow.

titusfortner:
Oh, a toggle was added. No wonder I didn't see any performance difference. I'm rerunning my tests.

Regardless of the results, I'm against adding more configuration toggles. Let's figure out the best way to implement things and do it that way. If there are bugs, we fix them. As it stands, the Travis CI results aren't valid because the toggle is defaulted to off.

lib/watir-webdriver/locators/element_locator.rb
```diff
@@ -146,6 +159,15 @@ def wd_find_by_regexp_selector(selector, method = :find)
         raise Error, "internal error: unable to build WebDriver selector from #{selector.inspect}"
       end
 
+      if how == :xpath && can_convert_regexp_to_contains?
+        rx_selector.each do |key, value|
+          unless key == :tag_name
```
titusfortner: As is, this fails most of our tests due to [the way Watir currently matches regex for text](https://github.com/watir/watir-webdriver/blob/master/lib/watir-webdriver/locators/element_locator.rb#L178).

For instance, our tests often do:
```browser.body(text: /Any text on the page/).present? # ==> true```
where
```browser.body(text: 'Any text on the page').present? # ==> false```
and 
```browser.p(text: 'Any text on the page').present? # ==> true```

This PR forces the text to match the particular tag.

One possible fix is to change this line to: 
```next if key == :tag_name || key == :text``` 

I'll take a look at how much performance increase we get from this change without text, and let you know tomorrow.
I know this would greatly reduce the effectiveness of this PR, but I do like the flexibility (if not the performance) of the current behavior to do things like:
```desired_row = browser.tr(text: /Text in First Column/)```
```desired_row.link(text: 'Edit')```

Comment (marxarelli) [inline lib/watir-webdriver/locators/element_locator.rb]: Sounds good. I had originally omitted conversion for text-node matching as well. I don't remember if that was why, but it's simple enough to reimplement.

As for performance increases, I suspect you'll see the most gain when driving a browser across higher latency networks, at least that's where [our impetus](https://phabricator.wikimedia.org/T92613) for making this optimization came from.

**Comment by Blackbaud-TitusFortner** [comment]
Interesting. With my 300,000 wire call test suite I'm getting about a 30% increase in findElement calls, along with the associated switchToFrame, getWindowHandles & getCurrentWindowHandle, and a 40% decrease in getElementAttributeCalls. Since attribute calls make up about 3 times as many calls as the other 4 calls put together, I'm seeing a total of about 4% decrease in total wire calls, and about the same amount of decrease in running time. 

So, my vote is to take out the toggle and turn this feature on for everyone for non-text attributes.

**Comment by marxarelli** [comment]
:+1: for being such a responsive upstream!

I've implemented the skipping of text nodes and refactored the specs a bit to reflect the baseline behavior as well as some edge cases related to regexp conversion.

**Comment by titusfortner** [comment]
@p0deje - PR looks good to me with this update. Thanks @marxarelli

File: spec/element_locator_spec.rb [hunk 6]
```diff
@@ -192,7 +192,7 @@
         if Watir.prefer_css?
           expect_all(:css, "div").and_return(elements)
         else
-          expect_all(:xpath, ".//div").and_return(elements)
+          expect_all(:xpath, "(.//div)[contains(@class, 'oob')]").and_return(elements)
```
p0deje: Why do you need to wrap div in `()`?

spec/element_locator_spec.rb
```diff
@@ -459,6 +457,68 @@
...
+          expect(Watir).to receive(:prefer_css?).and_return(false)
+        end
+
+        it "converts a leading run of regexp literals to a contains() expression" do
+          elements = [
+            element(tag_name: "div", attributes: { class: "foo" }),
+            element(tag_name: "div", attributes: { class: "foob" })
+          ]
+
+          expect_all(:xpath, "(.//div)[contains(@class, 'fo')]").and_return(elements.last(1))
```
p0deje: I feel like this is not the correct behavior. Why does it lookup `fo`? Will it iterate over these elements after?

p0deje:
I've left a couple of questions, but I think we can merge this and I'll start refactoring `ElementLocator` and working on regexp-to-xpath engine right after `ruby_2` is fully merged.

spec/element_locator_spec.rb
marxarelli: I introduced the `()` in case the base xpath did any unions, etc. In other words, it can be applied to expressions like `.//div | .//span`, becoming `(.//div | .//span)[contains(...)]` instead of `.//div | .//span[contains(...)]`. My xpath is pretty rusty, though.

spec/element_locator_spec.rb
marxarelli: Looks like I messed up the implementation for that example (should be `and_return(elements)` or include a non-matching element) but, yes, it should still iterate over each matching element and do the regexp comparison.

marxarelli:
Rebased and cleaned up that one example. Thanks!

spec/element_locator_spec.rb
p0deje: Okay, that makes sense.

p0deje:
@marxarelli Thanks a lot!

PR #333, opened 2015-04-17
make elements respond to methods they forward

daneandersen:
Implement respond_to_missing? so elements respond to the methods they forward.

### DIFF
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -607,6 +607,10 @@ module Watir
         super
       end
     end
+    
+     def respond_to_missing?(meth, include_private = false)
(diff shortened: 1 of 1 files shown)


titusfortner:
I'm in favor of this, with a [couple](https://github.com/bbatsov/ruby-style-guide#no-bang-bang) of [caveats](https://github.com/bbatsov/ruby-style-guide#no-and-or-or)

daneandersen:
I understand the reason for the || vs or, though I prefer using or for flow control and || for boolean logic to make it more intent revealing. The !! however I think is useful. This is a predicate method that should return a true or false, not a nil or a truthy Fixnum. Granted, this is a private method and unlikely that anyone would end up surprised by its return value for that reason. 

Should I make those changes then?

p0deje:
I don't really care for Rubocop `!!` style. It's more important to ensure the method returns boolean value.

PR #337, opened 2015-04-23
fix bug for multiple element matching same text

titusfortner:
This is the behavior I expect for a text match. One instance of text == one element matched. It requires a few spec updates in: https://github.com/watir/watirspec/pull/65


titusfortner:
Closing based on conversation in watir/watirspec#65

PR #339, opened 2015-04-27
Support elements generation from SVG spec

### LINKED ISSUE #332: SVG tags
What do you think about adding support for SVG tags?

The issue was raised in a [question on stack overflow](http://stackoverflow.com/questions/29633485/how-to-retrieve-values-from-a-rect-html-tag-using-watir)
SVG is listed in the html5 spec as valid, and includes [these tags](http://www.w3schools.com/svg/svg_inhtml.asp)

p0deje:
Fixes #332.

This pull requests brings the following on table:

1. Refactors existing generator to separate `HTML` and `SVG` implementations. The generator code itself has not been really changed, it was only slightly DRYed.
2. The same for `html:*` and `svg:*` Rake tasks.
3. Includes autogenerated elements from SVG specification.

Generators are not perfect for now:

1. if IDL has attribute that uses dictionary, it is ignored
2. interfaces, which IDLs are implemented in different spec are ignored (a lot of SVG elements)
3. need heavy refactoring

P.S. @titusfortner The diff is really noisy because I had to move some files around along with changes inside of them, but it would be great if you could go over them.


titusfortner:
This is great. I'm super busy this week, but I'm planning on taking a few days off next week to get caught up on things, including these pull requests. Thanks!

p0deje:
@titusfortner Any update on this? It's taking almost a month already and it'd be great to merge this as soon as possible so we could start testing it.

lib/watir-webdriver/browser.rb
```diff
@@ -41,7 +41,7 @@ def start(url, browser = :firefox, *args)
 
     def initialize(browser = :firefox, *args)
       case browser
-      when Symbol, String
+      when ::Symbol, String
```
titusfortner: What does this :: do differently?

titusfortner:
Sorry this took so long. Most of this is over my head/out of my wheel house. From what I do understand, I like what you've done. 

My only suggestion is naming html.rb & svg.rb to html_element.rb and svg_element.rb both to distinguish them in the folder as 'special' files, and to following the convention of the file name being the snake case of the class name.

Either way, this is really cool. Thanks for your work on it. It just needs a rebase to merge.

lib/watir-webdriver/browser.rb
p0deje: `Symbol` now resolves to `Watir::Symbol`, not `Object::Symbol`. That's why I added `::`.

p0deje:
I can rename files to `html_elements` and `svg_elements` because there are a lot of classes there, so the class = filename convention can't be applied here.

PR #348, opened 2015-06-22
implement scrolling method

### LINKED ISSUE #316: Extend .fire_event to allow for additional parameters needed for new html5 events
I propose that in order to better support level3 events such as wheel and mouse events, that we enhance the fire_event method to accept an optional hash in addition to the event name.  The hash could contain needed parameter name, value pairs that need to be populated for the proper firing of events like onwheel

Background / additional info:

New 'level3'  events relating to mouse wheel scrolling (http://www.w3.org/TR/DOM-Level-3-Events/#events-wheelevents) need 'delta' parameters provided to indicate the direction and amount of movement of the mouse wheel.  Without being able to provide this parameter, firing a 'wheel' event has basically zero effect as the JS code receiving the event needs the parameter to make sense of the event.  The 'standard' event is now defined as 'wheel' (onwheel) and is documented pretty well by mozilla here https://developer.mozilla.org/en-US/docs/Web/Events/wheel.  As you can see it allows for multiple parameters to indicate the scrolling movement of the wheel (x,y,z,unit)

The Wheel event replaces some older events (different in different browsers) documented reasonably well at http://www.javascriptkit.com/javatutors/onmousewheel.shtml (which has a nice little inline demo) While only having one or two parameters, even these older deprecated events basically don't work without being able to provide the 'delta', and folks testing for browser compatibility with older browsers may still need to be able to fire these events and include parameters when doing so. 

In addition to wheel events where we need to indicate direction+distance along at least one axis, there are also level3 mouse events which have parameters for things like alt, ctrl, shift keys being held down, which button(s) are pressed, etc.  Being able to include those parameters allows us to easily accommodate creating mouse events such as a drag and drop with a key like ctrl or alt held down, or something strange like a shift-left-mousedown.

One reason users need to be able to fire such events are Custom scrollbox controlls like the jQuery "slimScroll" control (see  http://rocha.la/jQuery-slimScroll for live examples you can try to script against) These controls are mainly manipulated by wheel events, or drag and drop of a div that is rendered as a scrollbar.  In some cases accessing the scrollbar div is very difficult (if autohide is enabled) and it would be far simpler to be able to just fire mousewheel events instead dragging the dammed bar.  However this is not currently possible because fire_event does not allow for any additional parameters to indicate the 'delta' value(s).

titusfortner:
I think this should be sufficient to close #316 
I used the same method name as in [cheezy's page objects](https://github.com/cheezy/page-object/blob/master/lib/page-object/platforms/selenium_webdriver/element.rb#L279) for consistency

### DIFF
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -245,6 +245,17 @@ module Watir
     end
 
     #
+    # Scrolls page until element is in view
+    #
+    # @example
+    #   browser.div(id: "scroll_to").scroll_into_view
+    #
+
+    def scroll_into_view
+      wd.location_once_scrolled_into_view
+    end
+
+    #
     # Returns value of the element.
     #
     # @return [String]


p0deje:
Few notes on this:

1. Jari [mentioned](https://github.com/watir/watir-webdriver/issues/206#issuecomment-18464074) that `Element#location_once_scrolled_into_view` is not reliable and might be removed.
2. Why not just use [watir-scroll](https://github.com/p0deje/watir-scroll)?
3. Changes break doctests.

titusfortner:
1) According to the spec, something like "scrollIntoView" will be implemented: https://w3c.github.io/webdriver/webdriver-spec.html#h-click

2) I hadn't seen watir-scroll before, but like many things, I like to avoid using Javascript in my tests where possible. Even if it doesn't work in all cases, for the specific use case mentioned in the linked issue this implementation did work. As such, I think it is useful to have that exposed directly without needing to access the WebDriver Element method.

3) I'll look into the doctests.

p0deje:
1) But that is a part of internal API, not public one.

2) It just the same `scrollIntoView` when called with `Watir::Element` (e.g. FirefoxDriver [does the same thing](https://github.com/SeleniumHQ/selenium/blob/master/javascript/firefox-driver/js/utils.js#L690-L695)).

titusfortner:
Tabling this discussion for now.

Issue #351, opened 2015-07-01
run_checkers and alerts


titusfortner:
Theoretically clicking an alert should run the error checkers since it can result in a reloaded page. Current implementation of Alerts is independent of the Browser, so it can't easily be triggered. 

My code is currently doing this workaround, but it shouldn't need a separate method:

``` 
def click_with_alert
    Page.browser.without_checkers { click }
    Page.browser.alert.ok
    Page.browser.run_checkers
end
```

p0deje:
Do you mean we should run checkers after `Alert#ok`? I guess it's pretty easy to implement.

Blackbaud-TitusFortner:
We'd have to make a significant change to how we implement Alerts, but it shouldn't be too complicated. I'm just adding it to the list of things to do. :) I should be able to get to it later this summer.

p0deje:
Guess we need to extract checkers to `Watir::Checkers` which implements `#run`, `#add`, etc. Then we can pass it to `Alert` constructor.

p0deje:
In master now.

PR #352, opened 2015-07-02
Extract Checkers into separate API and run them after alert ok/close

p0deje:
Related to #351 

This is a quick and working solution to the problem of sharing checkers between. We can actually implement it way easier by passing `Watir::Browser` instance to `Watir::Alert` constructor and just calling `browser#run_checkers` later.

I kinda like this solution more because it remove checkers responsibility from `Watir::Browser` and make API more OO. Though, this means we have to deprecate current API and update it watir-classic too.

@titusfortner @jarmo @jkotests Please take a look. If you don't like this solution, I can easily revert and forget about it.

Specs are in https://github.com/watir/watirspec/pull/67

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 ### Unreleased
 
-* Deprecated checkers. Use `Browser#after_hooks` instead.
+* New AfterHooks API that deprecates Checkers:
+  * Use `Browser#after_hooks#add` instead of `Browser#add_checker`
+  * Use `Browser#after_hooks#delete` instead of `Browser#disable_checker`
+  * Use `Browser#after_hooks#run` instead of `Browser#run_checkers`
+  * Use `Browser#after_hooks#without` instead of `Browser#without_checkers`
 
 ### 0.8.0 (2015-06-26)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+### Unreleased
+
+* Deprecated checkers. Use `Browser#after_hooks` instead.
+
 ### 0.8.0 (2015-06-26)
 
 * Ruby 1.8 is no longer supported. Ruby 1.9 still works, but not supported as well (#327)
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -291,12 +291,12 @@ module Watir
     end
 
     #
-    # @deprecated Use `Watir::Checkers#disable` instead
+    # @deprecated Use `Watir::Checkers#delete` instead
     #
 
     def disable_checker(checker)
-      warn 'Browser#disable_checker is deprecated. Use Browser#checkers#disable instead.'
-      @checkers.disable(checker)
+      warn 'Browser#disable_checker is deprecated. Use Browser#checkers#delete instead.'
+      @checkers.delete(checker)
     end
 
     #
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -10,6 +10,7 @@ module Watir
     include Waitable
 
     attr_reader :driver
+    attr_reader :checkers
     alias_method :wd, :driver # ensures duck typing with Watir::Element
 
     class << self
@@ -49,9 +50,9 @@ module Watir
         raise ArgumentError, "expected Symbol or Selenium::WebDriver::Driver, got #{browser.class}"
       end
 
-      @error_checkers = []
+      @checkers = Checkers.new(self)
       @current_frame  = nil
-      @closed         = false
+      @closed = false
     end
 
     def inspect
@@ -74,7 +75,7 @@ module Watir
       uri = "http://#{uri}" unless uri =~ URI.regexp
 
       @driver.navigate.to uri
-      run_checkers
+      @checkers.run
 
       url
     end
@@ -199,7 +200,7 @@ module Watir
 
     def refresh
       @driver.navigate.refresh
-      run_checkers
+      @checkers.run
     end
 
     #
@@ -281,79 +282,39 @@ module Watir
     end
 
     #
-    # Adds new checker.
-    #
-    # Checkers are generally used to ensure application under test does not encounter
-    # any error. They are automatically executed after following events:
-    #   1. Open URL
-    #   2. Refresh page
-    #   3. Click, double-click or right-click on element
-    #
-    # @example
-    #   browser.add_checker do |page|
-    #     page.text.include?("Server Error") and puts "Application exception or 500 error!"
-    #   end
-    #   browser.goto "www.watir.com/404"
-    #   "Application exception or 500 error!"
-    #
-    # @param [#call] checker Object responding to call
-    # @yield Checker block
-    # @yieldparam [Watir::Browser]
+    # @deprecated Use `Watir::Checkers#add` instead
     #
 
     def add_checker(checker = nil, &block)
-      if block_given?
-        @error_checkers << block
-      elsif checker.respond_to? :call
-        @error_checkers << checker
-      else
-        raise ArgumentError, "expected block or object responding to #call"
-      end
+      warn 'Browser#add_checker is deprecated. Use Browser#checkers#add instead.'
+      @checkers.add(checker, &block)
     end
 
     #
-    # Deletes checker.
-    #
-    # @example
-    #   checker = lambda do |page|
-    #     page.text.include?("Server Error") and puts "Application exception or 500 error!"
-    #   end
-    #   browser.add_checker checker
-    #   browser.goto "www.watir.com/404"
-    #   "Application exception or 500 error!"
-    #   browser.disable_checker checker
-    #   browser.refresh
+    # @deprecated Use `Watir::Checkers#disable` instead
     #
 
     def disable_checker(checker)
-      @error_checkers.delete(checker)
+      warn 'Browser#disable_checker is deprecated. Use Browser#checkers#disable instead.'
+      @checkers.disable(checker)
     end
 
     #
-    # Runs checkers.
+    # @deprecated Use `Watir::Checkers#run` instead
     #
 
     def run_checkers
-      @error_checkers.each { |e| e.call(self) } if !@error_checkers.empty? && window.present?
+      warn 'Browser#run_checkers is deprecated. Use Browser#checkers#run instead.'
+      @checkers.run
     end
 
     #
-    # Executes a block without running error checkers.
-    #
-    # @example
-    #   browser.without_checkers do
-    #     browser.element(name: "new_user_button").click
-    #   end
-    #
-    # @yieldparam [Watir::Browser]
+    # @deprecated Use `Watir::Checkers#without` instead
     #
 
-    def without_checkers
-      current_checkers = @error_checkers
-      @error_checkers = []
-      yield(self)
-    ensure
-      @error_checkers = current_checkers
+    def without_checkers(&block)
+      warn 'Browser#without_checkers is deprecated. Use Browser#checkers#without instead.'
+      @checkers.without(&block)
     end
 
     #
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -10,7 +10,7 @@ module Watir
     include Waitable
 
     attr_reader :driver
-    attr_reader :checkers
+    attr_reader :after_hooks
     alias_method :wd, :driver # ensures duck typing with Watir::Element
 
     class << self
@@ -50,7 +50,7 @@ module Watir
         raise ArgumentError, "expected Symbol or Selenium::WebDriver::Driver, got #{browser.class}"
       end
 
-      @checkers = Checkers.new(self)
+      @after_hooks = AfterHooks.new(self)
       @current_frame  = nil
       @closed = false
     end
@@ -75,7 +75,7 @@ module Watir
       uri = "http://#{uri}" unless uri =~ URI.regexp
 
       @driver.navigate.to uri
-      @checkers.run
+      @after_hooks.run
 
       url
     end
@@ -200,7 +200,7 @@ module Watir
 
     def refresh
       @driver.navigate.refresh
-      @checkers.run
+      @after_hooks.run
     end
 
     #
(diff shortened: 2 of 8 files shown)


jkotests:
I would agree to this API change.

The suggested API would make it consistent with the Cookies API.

As well, I do not think the changes will heavily impact users. I would guess that many people do not know the checkers functionality exists. For those that use it, it would just be a couple of small updates.

titusfortner:
Nice! :+1: I only got as far as thinking we'd need to pass browser into Alert. Splitting out Checkers looks great. Blackbaud uses a lot of checkers, so this definitely will make their code better. Thanks!

lib/watir-webdriver/browser.rb
```diff
@@ -10,6 +10,7 @@ class Browser
     include Waitable
 
     attr_reader :driver
+    attr_reader :checkers
```
jarmo: Since we're already changing the API itself then maybe we could use some better naming here as well instead of ambigious `checker`? What about something in the lines of `after_hooks` or something similar, because that's what they essentially are. Or?

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
...
+    #   end
+    #   browser.goto "www.watir.com/404"
+    #   "Application exception or 500 error!"
+    #
+    # @param [#call] checker Object responding to call
+    # @yield Checker block
+    # @yieldparam [Watir::Browser]
+    #
+
+    def add(checker = nil, &block)
```
jarmo: I'd rename it `#<<` to be more Ruby way. Just an opinion.

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
+module Watir
+  class Checkers
+
+    def initialize(browser)
+      @browser = browser
+      @error_checkers = []
```
jarmo: Since class name is already `Checkers` then having an instance variable called `@error_checkers` seems to be too much. Maybe `hooks` if you rename the checkers in the first place.

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
+module Watir
+  class Checkers
+
```
jarmo: I'd include `Enumerable` here and implement `#each` to make blocks iterable so we'd give more options to the users without adding much of the complexity.

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
...
+    #     page.text.include?("Server Error") and puts "Application exception or 500 error!"
+    #   end
+    #   browser.checkers.add checker
+    #   browser.goto "www.watir.com/404"
+    #   "Application exception or 500 error!"
+    #   browser.disable_checker checker
+    #   browser.refresh
+    #
+
+    def disable(checker)
```
jarmo: Rename it to `#delete` because that's what it is doing.

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
...
+      @error_checkers.delete(checker)
+    end
+
+    #
+    # Runs checkers.
+    #
+
+    def run
+      if @error_checkers.any? && @browser.window.present?
```
jarmo: Isn't there any case where browser does not need to be present? Why do we need to force it to be present here? Or what happens if it will not be present during running those checkers?

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,88 @@
...
+    #     browser.element(name: "new_user_button").click
+    #   end
+    #
+    # @yieldparam [Watir::Browser]
+    #
+
+    def without
+      current_checkers = @error_checkers
+      @error_checkers = []
+      yield(@browser)
```
jarmo: I don't know too much about inner workings of watir-webdriver, but wouldn't something like this work:
```ruby
def without
  yield(Browser.new(@browser.driver))
end
```

With this we wouldn't need to mess with arrays etc and the `#without` method would not be changing more-or-less the browser object.

It means though that you'd need to use that new browser in your block thus the example in the comments wouldn't work:

```ruby
browser.checkers.without do |browser_without_checkers|
  browser_without_checkers.element(name: "new_user_button").click
end
```

API like this would make it more explicit that only the browser inside of the block is without `checkers`.

jarmo:
Added some comments. Do not take all of them too seriously, but you asked for them :)

lib/watir-webdriver/checkers.rb
p0deje: I think alias would be great.

lib/watir-webdriver/checkers.rb
p0deje: Good idea.

lib/watir-webdriver/checkers.rb
p0deje: Agree.

lib/watir-webdriver/checkers.rb
p0deje: Agree.

lib/watir-webdriver/checkers.rb
p0deje: I haven't even thought about this. This is an interesting idea, but I'm not sure it is a better way to do that. First, because we're only modifying self state here, not browser. Second, there is a chance the new browser instance will still have checkers (e.g. with [watir-rails](https://github.com/watir/watir-rails/blob/master/lib/watir/browser.rb#L54)). I think messing with arrays is not that complicated but is more reliable.

lib/watir-webdriver/browser.rb
p0deje: I actually like "Checkers" or maybe I just got used to them. Maybe `AfterHooks` is a better name, but I don't really have a solid opinion on this.

lib/watir-webdriver/checkers.rb
p0deje: There were some exceptions (covered with [watirspec](https://github.com/watir/watirspec/blob/master/browser_spec.rb#L411-L422)).

p0deje: Maybe @titusfortner can comment here.

p0deje:
@jarmo Thanks a lot for your review, it was really great.

I have implemented some fixes and refactoring (particularly included `Enumerable` and add few collection methods). The most important question here is should we rename checkers to anything else? I'm personally fine with `Checkers` but I don't have a strong opinion about this.

lib/watir-webdriver/browser.rb
jarmo: Try to think from a user/developer perspective who is not familiar with Watir and it's concepts yet - would it be easier to understand if they'd read word `Checkers` or `AfterHooks`. We both are not target audience for this change and i might be wrong, but it seems to me that `AfterHooks` is more self-explanatory.

p0deje: What do others think? cc @jkotests @titusfortner

jkotests: If we were starting from scratch, I would go with `AfterHooks` or `Checkers`. `AfterHooks` seems more accurate. However, since we have been using the term "checkers" for a while, I would be fine with continuing to use the term.

Perhaps we should rename everything to `AfterHooks`, but also make `checkers` an alias for `after_hooks`? This would allow us to use the better terminology while allowing existing users to stay with the current terminology.

titusfortner: "error_checkers" is kind of ambiguous, and "run_checkers" even moreso, so I think doing just "checkers" is worse, especially since it allows you to do anything you want after a potential page load. I agree with @jkotests, let's rename everything to AfterHooks, and allow an alias for Browser#checkers

lib/watir-webdriver/checkers.rb
titusfortner: So... having used these a ton... what would you think about allowing for the use of names to identify them? It would make seeing what is getting run during debugging easier. It would make deleting much easier.
```
def add(checker=nil, name: nil, &block)
  name ||= size.to_s
  if block_given?
@checkers << {name => block}
...

File: lib/watir-webdriver/checkers.rb [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+      @checkers = current_checkers
+    end
+
+    #
+    # Yields each checker.
+    #
+    # @yieldparam [#call] checker Object responding to call
+    #
+
+    def each
```
titusfortner: Why do we need this? It's only getting used in the #run, can't it be done directly?

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,132 @@
...
+  #   3. Click, double-click or right-click on element.
+  #   4. Alert closing.
+  #
+
+  class Checkers
+    include Enumerable
+
+    def initialize(browser)
+      @browser = browser
+      @checkers = []
```
titusfortner: Is there any good reason not to have an attr_reader for this?

lib/watir-webdriver/checkers.rb
```diff
@@ -0,0 +1,132 @@
...
+    #   browser.checkers.add do |browser|
+    #     browser.text.include?("Server Error") and puts "Application exception or 500 error!"
+    #   end
+    #   browser.goto "www.watir.com/404"
+    #   "Application exception or 500 error!"
+    #   browser.checkers.delete browser.checkers[0]
+    #   browser.refresh
+    #
+
+    def delete(checker)
```
titusfortner: this was #disable before; another alias?

lib/watir-webdriver/checkers.rb
titusfortner: Oh. Enumerable. Hmmm.  What use cases are we expecting for this?

lib/watir-webdriver/checkers.rb
titusfortner: It's for when you have a second window open and you click a button that closes the window. The checkers kick off immediately after the click, but before the driver is gets pointed back to an active window. Without checking for window.present?, anything that makes a wire call will fail because there is nothing to make a call on.

When you're running checkers, alerts and multiple windows are a huge pain (especially since FF & Chrome behave differently), which is why I implemented #without_checkers. Neither of these checks here would be necessary if we instead had methods for #click_for_alert and #click_for_window_open and #click_for_window_close.

lib/watir-webdriver/checkers.rb
titusfortner: No, this defeats the purpose of the method. #without_checkers was implemented for things like clicks that open alerts. You want to click something to open the alert, but you can't run your checkers without it throwing an exception for trying to make a wire call when an alert is present. A better example:

```Ruby
browser.checkers.without do
  browser.element(id: "delete_object").click
  raise "Unexpected alert text: #{browser.alert.text}" unless browser.alert.text == "Are you sure?"
end
browser.alert.ok
```

lib/watir-webdriver/browser.rb
p0deje: Okay, I'll refactor to `AfterHooks` and leave old methods deprecated.

lib/watir-webdriver/checkers.rb
p0deje: Interface looks weird to me and I also think that currently you can easily do similar things since checker should just respond to `#call`:

```ruby
class MyChecker
  def call(browser)
    # ...
  end
end

browser.add_checker MyChecker.new
```

lib/watir-webdriver/checkers.rb
p0deje: Actually it should be refactored to allow `Enumerator`.

lib/watir-webdriver/browser.rb
jarmo: :+1:

lib/watir-webdriver/checkers.rb
p0deje: We just encapsulate it, why would you want to expose it?

p0deje:
I've renamed it to `AfterHooks` and entry point is `Browser#after_hooks`. I'll merge this tomorrow unless someone raises other concerns.

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+### Unreleased
+
+* Deprecated checkers. Use `Browser#after_hooks` instead.
```
jarmo: Maybe also write about previous vs new API here?

titusfortner:
Enumerable operations do not make sense for a collection of Procs. 

1) There aren't any [Proc methods](http://ruby-doc.org/core-2.2.0/Proc.html) that provide any useful ways to interact with an after hook. For instance, this kind of thing isn't helpful:
```browser.after_hooks.find {|hook| hook.to_s.include? "Proc:0x007ff459acba98"}```

2) Operations will all be identity based, not generic equality based.
```Ruby
foo_proc = Proc.new {puts "foo"}
browser.after_hooks.add {puts "foo"}
browser.after_hooks.include?(foo_proc) # ==> false
```
This is the same issue I have with the delete method. To delete an after_hook, you either have to have stored the object elsewhere (redundant), or you have to keep track of where it is stored in the array (useful only for more trivial use cases). 

This is the reason I suggested we make it an array of hashes, to make it easier to interact with them:
```Ruby
proc_404 = Proc.new {raise "Page Not Found" if browser.title.include?("404 Error") }
browser.after_hooks.add(error_404: proc_404}
browser.after_hooks.find {|h| h.key? :error_404 } # => {error_404 => Proc:0x007ff459acba98}
browser.after_hooks.find {|h| h.key? :error_500 } # => nil
```
I'm ok with our not doing that (the new company isn't using error checkers, yet), but I wanted to at least raise it as a possibility. @Blackbaud-KyleMartinez - you might want to chime in here since you probably have the biggest set of implementations for this.

p0deje:
But we still allow objects responding to `#call`, so all the problems can easily be resolved on client side:

```ruby
require 'watir-webdriver'

BasicHook = Class.new do
  attr_reader :name

  def initialize(name, &block)
    @name = name
    @block = block
  end

  def call(browser)
    @block.call(browser)
  end
end

hook1 = BasicHook.new('hook 1') { p browser.text.include?('Hook 1') }
hook2 = BasicHook.new('hook 2') { p browser.text.include?('Hook 2') }

begin
  browser = Watir::Browser.new
  browser.after_hooks.add hook1
  browser.after_hooks.add hook2
  browser.after_hooks.each { |hook| p hook.name }
ensure
  browser.quit
end


__END__

$ bunde exec ruby hooks.rb
"hook 1"
"hook 2"
```

titusfortner:
Wow. I'm so used to seeing Procs passed around as checkers that just implementing an object with a call method never occurred to me. This is wonderful.

CHANGES.md
p0deje: Added in 49a35d1ee83de31952b2c7e013f8d5c95e6bf1e3

p0deje:
Guess we're good to merge then (once CI succeeds).

jarmo:
:+1:

PR #358, opened 2015-08-20
Use latest Chrome, Ruby and PhantomJS on Travis

titusfortner:
Chrome & Chromedriver versions were both very out of date. Phantom code should either get run or deleted, I opted for deleting, any arguments for running it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,3 @@ env:
   - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs ALWAYS_LOCATE=false
   - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs PREFER_CSS=1 SELECTOR_STATS=1
   - RAKE_TASK=yard:doctest
-matrix:
-  allow_failures:
-    - env: RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs
-    - env: RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs ALWAYS_LOCATE=false
-    - env: RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs PREFER_CSS=1 SELECTOR_STATS=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 rvm:
   - 2.0
   - 2.1
+  - 2.2
 notifications:
   recipients:
     - p0deje@gmail.com
     - jarmo.p@gmail.com
+    - titusfortner@gmail.com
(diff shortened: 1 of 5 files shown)


p0deje:
I've actually been thinking that it would be great to run as many browsers as possible. Specs for Phantom fail which means it probably should be reported as Selenium bugs.

titusfortner:
The phantom tests in Selenium are passing, so if they are failing for Watir, we probably need more Selenium tests. :) I thought we were having issues with intermittent failures in phantom, so they weren't actually providing good information, which is why it was commented out. So, yes, let's add them back in and get them working.

titusfortner:
PhantomJS tests are failing because we are using 1.9; 2.0 does not have a linux binary for easy download like 1.9 did, something about issues in the way it is built. I think there is a way we can build it in Travis and use it, maybe with npm, but I don't have time to figure it out. I'm removing the code because commented out code is bad when we have revision control, and we can put it back if/when we get it working. In the meantime, we should be running Ruby 2.2 and the latest Chrome & Chromedriver binaries.

.travis.yml
```diff
@@ -16,12 +18,3 @@ env:
   - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=chrome
   - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=chrome ALWAYS_LOCATE=false
   - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=chrome PREFER_CSS=1 SELECTOR_STATS=1
-  # - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs
-  # - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs ALWAYS_LOCATE=false
-  # - RAKE_TASK=spec WATIR_WEBDRIVER_BROWSER=phantomjs PREFER_CSS=1 SELECTOR_STATS=1
-  - RAKE_TASK=yard:doctest
```
p0deje: Why?

titusfortner: Because it is bad practice to comment out code? :)
We can add an issue to get Phantomjs working.

p0deje: You remove doctests :)

titusfortner: Oh, that was a mistake. Delete happy mood I guess. :)

p0deje: I'll fix.

titusfortner:
Oh, I haven't merged it yet, I'll rebase and add it in, though.

p0deje:
I'm currently trying to install PhantomJS 2 on Travis, I'll push the changes here if I succeed. Otherwise, let's merge it later today.

titusfortner:
Ah, sounds good. Thanks.

PR #362, opened 2015-09-02
Wait#until returns result of block

chrismikehogan:
Addresses https://github.com/watir/watir-webdriver/issues/360

### DIFF
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -37,7 +37,10 @@ module Watir
 
       def until(timeout = nil, message = nil)
         timeout ||= Watir.default_timeout
-        run_with_timer(timeout) { return true if yield(self) }
+        run_with_timer(timeout) do
+          result = yield(self)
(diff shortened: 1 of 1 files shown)


p0deje:
Thanks!

titusfortner:
If we are going to make this change (see comment in issue), I think Bret's suggestion is better.

p0deje:
It won't work, because it doesn't care if `yield(self)` returns true or false. It just returns after the first execution.

titusfortner:
Ah yeah, returns in blocks, that makes sense. :)

bret:
Oh I see. Thanks.

chrismikehogan:
Gracias, all.

When do patch cuts typically get made?

titusfortner:
Can you reference github instead of the gem in your gemfile for the time being?

chrismikehogan:
@titusfortner yep I'm all set locally, just curious

chrismikehogan:
curious

PR #363, opened 2015-09-09
Added methods to support when_enabled

Rodney-QAGeek:
Corresponding tests located here https://github.com/watir/watirspec/pull/69

### DIFF
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -404,6 +404,17 @@ module Watir
     end
 
     #
+    # Returns true if this element is exists and is enabled on the page.
+    #
+    # @return [Boolean]
+    #
+
+    def enabled?
+      assert_exists
+      assert_enabled
+    end
+
+    #
     # Returns true if the element exists and is visible on the page.
     #
     # @return [Boolean]
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -404,14 +404,13 @@ module Watir
     end
 
     #
-    # Returns true if this element is exists and is enabled on the page.
+    # Returns true if this element is present and enabled on the page.
     #
     # @return [Boolean]
     #
 
     def enabled?
-      assert_exists
-      assert_enabled
+      present? && element_call { @element.enabled? }
     end
 
     #
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -94,15 +94,8 @@ module Watir
     end
   end
 
-  #
-  # Wraps an Element so that any subsequent method calls are
-  # put on hold until the element is present (exists and is visible) on the page.
-  #
-
-  class WhenPresentDecorator
+  class BaseDecorator
     extend Forwardable
-
-    def_delegator :@element, :present?
 
     def initialize(element, timeout, message = nil)
       @element = element
@@ -114,14 +107,27 @@ module Watir
       @element.respond_to?(*args)
     end
 
-    def method_missing(m, *args, &block)
+    def method_missing(decorator, m, *args, &block)
       unless @element.respond_to?(m)
         raise NoMethodError, "undefined method `#{m}' for #{@element.inspect}:#{@element.class}"
       end
 
-      Watir::Wait.until(@timeout, @message) { @element.present? }
+      Watir::Wait.until(@timeout, @message) { @element.send(decorator) }
 
       @element.__send__(m, *args, &block)
+    end
+  end
+
+  #
+  # Wraps an Element so that any subsequent method calls are
+  # put on hold until the element is present (exists and is visible) on the page.
+  #
+
+  class WhenPresentDecorator < BaseDecorator
+    def_delegator :@element, :present?
+
+    def method_missing(m, *args, &block)
+      super(:present?, m, *args, &block)
     end
   end # WhenPresentDecorator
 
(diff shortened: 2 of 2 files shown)


lib/watir-webdriver/elements/element.rb
```diff
@@ -404,6 +404,17 @@ def visible?
     end
 
     #
+    # Returns true if this element is exists and is enabled on the page.
+    #
+    # @return [Boolean]
+    #
+
+    def enabled?
+      assert_exists
```
titusfortner: We want this to always return true/false (or at least truthy/false), and #assert_exists and #assert_enabled will both raise an exception if they fail.

Also, is there ever a case where we would want to know if an element is enabled if it isn't visible?

What if we did ```present? && element_call { @element.enabled? }```

titusfortner:
@p0deje - Do you know why the yardoc tests failed? I've not yet looked at that gem to figure out what it's actually testing. :)

The WhenEnabledDecorator looks good to me (and will be useful for my tests as well). There might be a way to DRY up the code since this is much of it is the same as WhenPresentDecorator, but I'm not sure it's worth doing right now.

lib/watir-webdriver/elements/element.rb
p0deje: It definitely should return true or false (not truthy or falsy).

p0deje:
Doctests are failing because a proper page, where element in `@example` can be located, has to be used. See [doctest_helper](https://github.com/watir/watir-webdriver/blob/master/support/doctest_helper.rb) for examples.

I'm curious why is it needed? I personally never faced a case when I need to wait for element to become "enabled", so having a decorator looks like an overkill to me.

Also, do you think it makes sense to have `enabled?` defined on `Element`? I know it's actually used there (e.g. in `#click`), but as far as I understand, it is only applicable to a set of elements (e.g. inputs). Another problem is to have it overwritten by `Button/Input/Select` classes.

titusfortner:
The reason for this is javascript libraries. Often it is the case that a click event or text input will change the state of a button from disabled to enabled. Then you get race conditions where Watir takes the action that will change the status, but interacts with the element before the status has actually changed. It clicks the button, but nothing happens because the element is still disabled. My code currently includes several instances of ```Watir::Wait.until { foo_element.enabled? }; foo_element.click```

As far as implementation goes, I think a better question is when ```!disabled``` based on the attribute is going to be different from the ```@element.enabled?``` call. If it's never different, then we can just do ```respond_to :disabled && !disabled``` in the Element class and remove it from the subclasses, since there are 9 elements in html_elements.rb that have a disabled attribute, but we only implement it for Button/Input/Select. If disabled attribute and @element.disabled return different values, then perhaps ```(respond_to :disabled && !disabled) || (present? && element_call { @element.enabled? }```

lib/watir-webdriver/elements/element.rb
Rodney-QAGeek: Totally agree, That makes a lot more sense!

Rodney-QAGeek:
My use case is exactly what Titus mentioned. On input into a text field my submit button is disabled, the site makes an ajax request and re-enables the button depending on the returned response. i.e. check if an email address is available for use.

Also the "WhenEnabledDecorator" is very much copied code, I don't mind taking a stab at DRY'ing this up a bit. I didn't modify anything in the initial PR to keep it simple as as possible and to the point.

Thanks for responding so quickly and for the amazing feedback!

titusfortner:
@Rodney-QAGeek - You should rebase your code to master, make the changes suggested here, and force push to your when_enabled branch so we can pull this in.

Rodney-QAGeek:
Yep, sorry about the delay, I'll work on this tomorrow!

Rodney-QAGeek:
Hey @p0deje I've tried addressing the CI issues, but I need a helping hand. It seems like it's failing now because it's attempting to pull the master branch from the watirspec submodule, but my changes are on a different branch found here. https://github.com/watir/watirspec/pull/69

I've also not found any specific updates I needed to make from your doctest_helper examples.

Any help would be greatly appreciated.

p0deje:
Hey @Rodney-QAGeek ,

Usually, we avoid commiting watirspec submodule changes until it's merged to master, so if your specs are passing locally, just rebase this pull request to exclude `.gitmodules` and I'll bump watirspec submodule.

Regarding to doctests, the examples you document for methods are executed as it is, so since you are using `button(name: "submit")`, it has to be present on the page that is currently opened in browser. By default, all doctests are [opened](https://github.com/watir/watir-webdriver/blob/master/support/doctest_helper.rb#L12) with [`forms_with_input_elements`](https://github.com/watir/watirspec/blob/master/html/forms_with_input_elements.html) page, so you can just change your example to use elements from this page:

```ruby
# @example
#   browser.button(name: "delete_user_submit").when_enabled.click
```

lib/watir-webdriver/wait.rb
```diff
@@ -114,18 +107,44 @@ def respond_to?(*args)
       @element.respond_to?(*args)
     end
 
-    def method_missing(m, *args, &block)
+    def method_missing(decorator, m, *args, &block)
```
p0deje: I'm not sure it's a very good idea to have custom API for `method_missing` even though it's not designed to be used directly.

Maybe it's better to duck type the method that should be waited for? For example:

```ruby
class BaseDecorator
  def method_missing(m, *args, &block)
    Watir::Wait.until(@timeout, @message) { wait_until }
  end
end

class WhenPresentDecorator < BaseDecorator
  private
  def wait_until
    @element.present?
  end
end

class WhenEnabledDecorator < BaseDecorator
  private
  def wait_until
    @element.enabled?
  end
end
```

@titusfortner Any idea?

Rodney-QAGeek: I totally agree with you here on this. I made the change. If you or Titus wants me to change it back, it's easy enough to do. I've made all of your suggestions and because rebasing was proving to be convoluted I've opened fresh PR's for both watir-webdriver (https://github.com/watir/watir-webdriver/pull/370) and watirspec (https://github.com/watir/watirspec/pull/72).

lib/watir-webdriver/wait.rb
```diff
@@ -119,13 +112,37 @@ def method_missing(m, *args, &block)
...
     end
+  end
+
+  #
+  # Wraps an Element so that any subsequent method calls are
+  # put on hold until the element is present (exists and is visible) on the page.
+  #
+
+  class WhenPresentDecorator < BaseDecorator
+    private
```
p0deje: Add empty line before and after `private`.

lib/watir-webdriver/wait.rb
```diff
@@ -94,16 +94,9 @@ def wait_while(*args, &blk)
...
-  #
-  # Wraps an Element so that any subsequent method calls are
-  # put on hold until the element is present (exists and is visible) on the page.
-  #
-
-  class WhenPresentDecorator
+  class BaseDecorator
     extend Forwardable
 
-    def_delegator :@element, :present?
```
p0deje: This should be part of custom delegator, isn't it?

p0deje:
@Rodney-QAGeek Would you please rebase and merge into one commit this (and also resolve conflicts)?

Rodney-QAGeek:
I believe we can blow this away. If we merge https://github.com/watir/watir-webdriver/pull/370 instead.

Rodney-QAGeek:
Sorry about the noise, I'll delete this PR.

PR #366, opened 2015-09-24
get browser text like text of any other element

titusfortner:
This gives us:
Stale Element Protection. I hit this race condition working with watirspecs on my mac yesterday.
and
Context Protection. Current code will give different responses depending on which frame has recently been pointed to. Browser#text should always return top level, and IFrame#text needs to be used if text from a nested context is desired.

### DIFF
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -170,7 +170,7 @@ module Watir
     #
 
     def text
-      @driver.find_element(:tag_name, "body").text
+      body.text
     end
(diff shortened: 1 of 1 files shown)


p0deje:
:+1: 

A watirspec for it would be great too (showing browser vs iframe).

titusfortner:
Good call on the spec. This fails before this code change and passes after: https://github.com/watir/watirspec/commit/d1277e749b8ee41904d5cd0b07a793a553b2d7ef

PR #367, opened 2015-09-24
performance improvement for large drop downs

titusfortner:
Our site has some drop downs with a ridiculous number of options in them (over 1,000, don't ask).
The current implementation requires a complete iteration of all options more than once to return a response.
This implementation will return a true response as soon as it finds it, and doesn't add much overhead for iterating through everything to give a false response if it isn't found.

### DIFF
--- a/lib/watir-webdriver/elements/select.rb
+++ b/lib/watir-webdriver/elements/select.rb
@@ -86,17 +86,22 @@ module Watir
 
     def selected?(str_or_rx)
       assert_exists
-      matches = element_call do
-        @element.find_elements(:tag_name, 'option').select do |e|
-          str_or_rx === e.text || str_or_rx === e.attribute(:label)
+
+      match_found = false
+
+      element_call do
+        @element.find_elements(:tag_name, 'option').each do |e|
+          matched = str_or_rx === e.text || str_or_rx === e.attribute(:label)
+          if matched
+            return true if e.selected?
+            match_found = true
+          end
         end
       end
 
-      if matches.empty?
-        raise UnknownObjectException, "Unable to locate option matching #{str_or_rx.inspect}"
-      end
+      raise(UnknownObjectException, "Unable to locate option matching #{str_or_rx.inspect}") unless match_found
 
-      matches.any? { |e| e.selected? }
+      false
     end
 
     #


lib/watir-webdriver/elements/select.rb
```diff
@@ -86,17 +86,20 @@ def select_value(str_or_rx)
...
       assert_exists
-      matches = element_call do
-        @element.find_elements(:tag_name, 'option').select do |e|
-          str_or_rx === e.text || str_or_rx === e.attribute(:label)
+
+      matches = false
+
+      element_call do
+        @element.find_elements(:tag_name, 'option').each do |e|
+        matched = str_or_rx === e.text || str_or_rx === e.attribute(:label)
```
p0deje: please indent

lib/watir-webdriver/elements/select.rb
```diff
@@ -86,17 +86,20 @@ def select_value(str_or_rx)
...
+        return true if matched && e.selected?
+        matches = true if matched
```
p0deje: this can be simplified

```ruby
if matched
  matches = true
  return true if e.selected?
end
```

lib/watir-webdriver/elements/select.rb
```diff
@@ -86,17 +86,20 @@ def select_value(str_or_rx)
...
         end
       end
 
-      if matches.empty?
-        raise UnknownObjectException, "Unable to locate option matching #{str_or_rx.inspect}"
-      end
+      return false if matches
```
p0deje: That breaks backward compatibility, isn't it?

titusfortner: Essentially 
if there is a match and it is selected  --> true
if nothing that matches is selected --> false
If nothing matches --> raise error

What backward compatibility does it break?

p0deje: Sorry, I've misread the code.

p0deje:
Seems fine to me, though I'm a little confused with `matched` vs `matches` variable names, though I didn't come up with better naming.

titusfortner:
Yeah I'm going to change matches to found_match I think. Thanks.

PR #368, opened 2015-09-28
rake task to run specs with all browsers

titusfortner:
```rake all_browsers``` works to run all applicable tests for all browsers applicable for the operating system. I'm not a rakefile expert, so this probably can be improved.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -121,3 +121,23 @@ end
 load "spec/watirspec/watirspec.rake" if File.exist?("spec/watirspec/watirspec.rake")
 
 task default: [:spec, 'yard:doctest']
+
+
+namespace :spec do
+  require 'selenium-webdriver'
+  desc 'Run specs in all browsers'
+  task all_browsers: [:firefox,
+                      :firefox_nightly,
+                      :chrome,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Rakefile [hunk 1]
```diff
@@ -121,3 +121,44 @@ end
...
+
+def windows?
+  require 'selenium-webdriver'
+  Selenium::WebDriver::Platform.os == :windows
+end
+
+desc 'Run specs in all browsers'
+task all_browsers: [:firefox,
+                    :chrome,
+                    (:safari if mac?),
```
p0deje: You can avoid defining custom methods:

```ruby
(:safari if Selenium::WebDriver::Platform..mac?)
```

Rakefile
```diff
@@ -121,3 +121,44 @@ end
...
+                    (:ie if windows?),
```
p0deje: You haven't defined `:ie` task

p0deje:
What do you think about defining `:spec` as namespace so we could use it like:

```bash
$ rake spec:firefox # default
$ rake spec:chrome
# so on
$ rake spec:all_browsers
```

titusfortner:
Thanks for the help, it is much cleaner now.

PR #369, opened 2015-09-29
saves 4 wire calls for each navigation

titusfortner:
I was going through the Selenium code with @abaird and noticed that there was a lot more chatter than I expected for the navigation call. I don't think we need to go through the extra overhead of the url method here when we already have the uri that gets returned anyway. 

We just called the /url endpoint, and there is no scenario where we aren't in the top browsing context with an active browser after this. 

The current code Does a /window call then a /windows call then a /frame call then finally another /url call.

### DIFF
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -77,7 +77,7 @@ module Watir
       @driver.navigate.to uri
       @after_hooks.run
 
-      url
+      uri
     end
 
     #


titusfortner:
Actually, current code will return the url value after a redirect and this code won't, so this could cause some backward compatibility issues.

titusfortner:
Though backward compatibility with watir-classic is to return the time it took, not the url or redirected ur. https://github.com/watir/watir-classic/blob/master/lib/watir-classic/browser.rb#L346

p0deje:
I believe that was done unintentionally in https://github.com/watir/watir-webdriver/commit/bc2d41bee143255ad7ee0044d3e973b7670e308b because it's pretty easy to misspell `uri` vs `url`.

:+1:

PR #370, opened 2015-10-01
Added code to support when_enabled

Rodney-QAGeek:
https://github.com/watir/watirspec/pull/72

### DIFF
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -94,15 +94,7 @@ module Watir
     end
   end
 
-  #
-  # Wraps an Element so that any subsequent method calls are
-  # put on hold until the element is present (exists and is visible) on the page.
-  #
-
-  class WhenPresentDecorator
-    extend Forwardable
-
-    def_delegator :@element, :present?
+  class BaseDecorator
 
     def initialize(element, timeout, message = nil)
       @element = element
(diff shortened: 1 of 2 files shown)


lib/watir-webdriver/wait.rb
```diff
@@ -119,13 +112,37 @@ def method_missing(m, *args, &block)
         raise NoMethodError, "undefined method `#{m}' for #{@element.inspect}:#{@element.class}"
       end
 
-      Watir::Wait.until(@timeout, @message) { @element.present? }
-
+      Watir::Wait.until(@timeout, @message) { wait_until }
+
```
p0deje: Please, don't have empty lines with spaces.

lib/watir-webdriver/wait.rb
```diff
@@ -119,13 +112,37 @@ def method_missing(m, *args, &block)
...
     end
+  end
+
+  #
+  # Wraps an Element so that any subsequent method calls are
+  # put on hold until the element is present (exists and is visible) on the page.
+  #
+
+  class WhenPresentDecorator < BaseDecorator
+    private
```
p0deje: Please, add empty lines before and after `private`.

lib/watir-webdriver/wait.rb
```diff
@@ -94,16 +94,9 @@ def wait_while(*args, &blk)
...
-  #
-  # Wraps an Element so that any subsequent method calls are
-  # put on hold until the element is present (exists and is visible) on the page.
-  #
-
-  class WhenPresentDecorator
+  class BaseDecorator
     extend Forwardable
 
-    def_delegator :@element, :present?
```
p0deje: This should be a part of custom decorator, isn't it?

Rodney-QAGeek: Good catch, If I understand this correctly now, they should be added to each the Present and Enabled Decorators, right?

p0deje: Take a look at #216.

I don't think it makes sense to add delegator for `#enabled?`, but `#present?` is definitely needed.

Rodney-QAGeek: :pray: Thank you for this history. It makes much more sense now. I'll update and rebase.

p0deje:
:+1:

I'll merge this when watir/watirspec#72 is finished.

lib/watir-webdriver/wait.rb
```diff
@@ -94,16 +94,9 @@ def wait_while(*args, &blk)
...
   end
 
-  #
-  # Wraps an Element so that any subsequent method calls are
-  # put on hold until the element is present (exists and is visible) on the page.
-  #
-
-  class WhenPresentDecorator
+  class BaseDecorator
     extend Forwardable
```
p0deje: Just caught my eyes.

Since delegator is not used in base decorator, let's move this `extend` to `WhenPresentDecorator`.

Rodney-QAGeek: Derp! Good catch!

Rodney-QAGeek:
Fixed and tests are passing :)

p0deje:
Thanks for pull request and for a bunch of changes that I've asked you to apply!

PR #372, opened 2015-10-01
remove unnecessary :needs_server tags with html in watirspecs

titusfortner:
Having html files in watir-webdriver project requires the server to display based on the file implementation. I don't see a problem with putting them in the watirspec project so we can run these specs without the server. This PR is in conjunction with https://github.com/watir/watirspec/pull/73


p0deje:
See my comments in watir/watirspec#73.

titusfortner:
To be clear, you are ok with moving the 4 html files into watirspec?

p0deje:
:+1:

PR #373, opened 2015-10-01
support for MS Edge browser

titusfortner:
The biggest issue for Edge support is that the w3c protocol has unknown element values throwing an exception instead of returning nil. I'm not sure if we want to specify this as and edge-only implementation, or if there is anything wrong with always rescuing this call and returning nil.

The location_once_scrolled_into_view is admittedly a workaround for an Edge bug, but it affected many of the specs. Same with retrying when Edge lost contact of the top level browsing context. This latter might not be a bug depending on how you read the w3c spec, but it is hard to say.

Anyway, I want to be able to release this next version with "edge support," but I'm not sure how accommodating we want to be to their current bugs. Maybe we have a separate class that they could be imported only if they are using Edge that overwrites existing classes with Edge specific adjustments?

I've filed 13 bugs with the Edge project so far, so hopefully these things get fixed/improved soon.

### DIFF
--- a/lib/watir-webdriver/attribute_helper.rb
+++ b/lib/watir-webdriver/attribute_helper.rb
@@ -74,7 +74,7 @@ module Watir
 
     def define_boolean_attribute(mname, aname)
       define_method mname do
-        attribute_value(aname) == "true"
+        attribute_value(aname) =~ /(t|T)rue/
       end
     end
 
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -346,6 +346,10 @@ module Watir
         driver.switch_to.default_content
         true
       end
+    rescue Selenium::WebDriver::Error::NoSuchWindowError
+      raise unless driver.capabilities.browser_name == 'MicrosoftEdge'
+      driver.switch_to.window(driver.window_handle)
+      retry
     end
     alias_method :ensure_not_stale, :assert_exists
 
--- a/lib/watir-webdriver/elements/element.rb
+++ b/lib/watir-webdriver/elements/element.rb
@@ -125,6 +125,7 @@ module Watir
 
           action.perform
         else
+          @element.location_once_scrolled_into_view if driver.capabilities.browser_name == 'MicrosoftEdge'
           @element.click
         end
       end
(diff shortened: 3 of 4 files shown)


p0deje:
Not sure how is this different from #374, so I've posted all my comments there :)

titusfortner:
Sorry this got confusing #374 includes this one and #372.

The big thing right now is that w3c spec says that if an attribute isn't present that an exception should be raised, not nil returned. That isn't a workaround, that is a new requirement. I put the edge conditionals in there because it is the only one currently using w3c, though Firefox will be using it soon as well, so maybe we need to implement a #w3c? method to make this more clear.

I'm ok with these other items not getting added, though I'd like to find a better solution to browser specific bugs that we know about and have workarounds for, because it seems silly that everyone who uses this software needs to either put up with the bug or write their own workarounds.

p0deje:
I'm confused why watir-webdriver has to handle compliance to w3c spec rather than selenium-webdriver. I believe that kind of attribute handling should be done in selenium-webdriver.

titusfortner:
Selenium is handling the compliance. Selenium will correctly raise an exception when looking up attributes that aren't there. So when Watir makes a call for ```label.attribute(:for)``` and the label element doesn't have a 'for' attribute defined, Selenium, in accordance with the spec, is going to raise an exception. Since we check that every time just in case, it is our code that needs to rescue the exception to maintain current expectations. I'm not sure how we go about suggesting the spec not change in this way, but as it stands this is how Edge is doing things, and likely how Firefox will be doing things.

p0deje:
Oh, I see now.

I guess the better way then is to let selenium raise custom exception (e.g. `NoAttributeError` instead of `UnknownError`) and always rescue it without checking if the browser is Edge or not.

titusfortner:
Yes, many of Edge's errors are still Unknown. :)

And with all of this being said, now that I'm looking at the w3c spec again, it doesn't say at all what I remember it saying. So, I probably have no idea what I'm talking about and need to go file another bug with Edge.

p0deje:
Ok, waiting for your comments tomorrow :+1:

titusfortner:
I'll table these for now.

PR #374, opened 2015-10-01
Edge support specs

titusfortner:
This requires 2 other pull requests to be accepted, but this has all specs passing in Microsoft Edge


lib/watir-webdriver/attribute_helper.rb
```diff
@@ -74,7 +74,7 @@ def define_string_attribute(mname, aname)
 
     def define_boolean_attribute(mname, aname)
       define_method mname do
-        attribute_value(aname) == "true"
+        attribute_value(aname) =~ /(t|T)rue/
```
p0deje: The method should return boolean, while you changed it to return nil or fixnum. It's better to use one of:

```ruby
!!(attribute_value(aname) =~ /(t|T)rue/)
# or 
/(t|T)rue/ === attribute_value(aname)
```

lib/watir-webdriver/elements/element.rb
```diff
@@ -125,6 +125,7 @@ def click(*modifiers)
 
           action.perform
         else
+          @element.location_once_scrolled_into_view if driver.capabilities.browser_name == 'MicrosoftEdge'
```
p0deje: Guess that's a bug in EdgeDriver, shouldn't element be automatically scrolled into view?

lib/watir-webdriver/browser.rb
```diff
@@ -346,6 +346,10 @@ def assert_exists
         driver.switch_to.default_content
         true
       end
+    rescue Selenium::WebDriver::Error::NoSuchWindowError
```
p0deje: Why do we need to rescue this? Isn't window handling a part of user responsiblity?

p0deje:
Generally, I'm :-1: on this because we're pushing so many edge-specific hacks to watir-webdriver, while is a pain to support and (in some cases like returning empty string for attribute) is meaningless.

I believe the better way would be to wait until EdgeDriver is stable, update pull request to remove all these "rescue" hacks and release as usually. watir-webdriver has never really tried to handle all edge cases in webdriver implementations that should be fixed in upstream.

lib/watir-webdriver/elements/element.rb
titusfortner: Definitely a bug, I filed it with them.

lib/watir-webdriver/browser.rb
titusfortner: The problem is that sometimes Edge just loses the default context and throws this error, even when the user is doing normal things. This happens several times running our specs if this isn't rescued.

titusfortner:
Tabling for now.

PR #375, opened 2015-10-02
optimize Select #include? similar to #selected?

Conky5:
Optimize `#include?` similar to `#selected?` for `Select`, as desired by the TODO comment.

### DIFF
--- a/lib/watir-webdriver/elements/select.rb
+++ b/lib/watir-webdriver/elements/select.rb
@@ -49,7 +49,7 @@ module Watir
 
       element_call do
         @element.find_elements(:tag_name, 'option').any? do |e|
-          str_or_rx === e.text
+          str_or_rx === e.text || str_or_rx === e.attribute(:label)
         end
       end
     end
--- a/lib/watir-webdriver/elements/select.rb
+++ b/lib/watir-webdriver/elements/select.rb
@@ -45,8 +45,13 @@ module Watir
     #
 
     def include?(str_or_rx)
-      # TODO: optimize similar to selected?
-      options.any? { |e| str_or_rx === e.text }
+      assert_exists
+
+      element_call do
+        @element.find_elements(:tag_name, 'option').any? do |e|
+          str_or_rx === e.text
+        end
+      end
     end
 
     #


p0deje:
Thanks for a pull request!

I think the purpose of this TODO was to provide with similar to `#selected?` logic:

1. raise `UnknownObjectException` if no options present
2. check both option text and label attribute

Thus, I think the same changes should be applied.

Also, can you write a specs for all these cases. An example of such specs for `#selected?` are in [watirspec](https://github.com/watir/watirspec/blob/master/select_list_spec.rb#L217-230). Let me know if you need any help with this.

Conky5:
Howdy @p0deje! Thanks for your response.
1. I'm not sure raising an `UnknownObjectException` makes sense for `#include?` since it is only checking if the option exists.  The existing behavior is to just return `False` if the option does not exist.  A select list without any options would just always return `False`.
2. I'll add the label attribute check and tests, the `#include?` description actually says "text or label" already https://github.com/watir/watir-webdriver/blob/master/lib/watir-webdriver/elements/select.rb#L41

titusfortner:
The current implementation checks several times if each element is enabled, which this code doesn't do. Are select list options often disabled? And do we want to change our true/false based on that? Do we need a spec for this to verify this solution works?

Without the enabled checks, this solution makes significantly fewer wire calls and is more efficient at finding the first matching element.

I agree with @Conky5, it makes sense to raise an error if we are passing in a parameter and asking for information about it when it doesn't exist. Asking if it is included in the list should return true/false. I would see a potential test being, is it included, then check to see if it is selected so it doesn't raise the exception in the latter call.

p0deje:
> The current implementation checks several times if each element is enabled, which this code doesn't do. Are select list options often disabled? And do we want to change our true/false based on that? Do we need a spec for this to verify this solution works?

I think it doesn't make sense to check for enabled/disabled in `#include?` because can question method is responding to is if it's included, not can be selected.

> I agree with @Conky5, it makes sense to raise an error if we are passing in a parameter and asking for information about it when it doesn't exist. Asking if it is included in the list should return true/false. I would see a potential test being, is it included, then check to see if it is selected so it doesn't raise the exception in the latter call.

Agreed, let's stick to this behavior.

titusfortner:
Agreed. Definitely more performance boost than I expected. @Conky5 - Nice work.

Conky5:
Thanks @titusfortner :+1:

PR #378, opened 2015-10-08
Trying to select an option that is disabled should raise an error.

Blackbaud-KyleMartinez:
If select list has an option that is disabled web-driver should through an error if you try to select it.

### DIFF
--- a/lib/watir-webdriver/elements/select.rb
+++ b/lib/watir-webdriver/elements/select.rb
@@ -167,6 +167,7 @@ module Watir
           no_value_found(string)
         end
 
+        raise Selenium::WebDriver::Error::ElementNotSelectableError, "Element with string #{string} is disabled" unless e.enabled?
         e.click unless e.selected?
 
         safe_text(e)
@@ -193,6 +194,7 @@ module Watir
         element = elements.find { |e| matches_regexp?(how, e, exp) }
         no_value_found(exp) unless element
 
+        raise Selenium::WebDriver::Error::ElementNotSelectableError, "Element with regex #{exp} is disabled" unless element.enabled?
         element.click unless element.selected?
 
         safe_text(element)


Blackbaud-KyleMartinez:
@titusfortner The seems to be the easiest fix without knowing too much about the webdriver framework. Let me know if you would like me to change anything.

titusfortner:
Kyle! Thanks for the PR. Can you write a spec that demonstrates the undesired behavior without this PR? Click already asserts enabled before attempting to click. Is this just looking to add a more specific error? Or do you actually want to try to find a matching option that is enabled?

Blackbaud-KyleMartinez:
The scenario is that I have a select list and in that select list is an option that I can see but cannot actually select. I expect an error to be thrown because that select option is disabled. Currently when I do Page.browser.select_list().set "disabled option" no error is given by webdriver and it acts as if the action completed but nothing actually happened. I'll try to write a spec up that demonstrates the behavior.

titusfortner:
Oh interesting, you should be getting an ArgumentError with that code.
#set is not a valid method on the Select class. You want #select.
In this case #set should be looking for a svg element with tag_name: 'set' and raising an exception that you aren't passing in a properly formatted selector to it.

@p0deje - since #set is an SVG element, that means we can't nest it under any Element that takes #set since we are overriding that method? I don't think this is a likely scenario, but it is interesting.

Blackbaud-KyleMartinez:
I'm sorry Titus. We are using the Watir::Select select method. I thought we were using set for some reason.

p0deje:
@titusfortner Actually, `#set` that is defined as SVG element is being overriden by custom `#set` implementations because `Container` is included __before__ everything else.

p0deje:
@titusfortner That is happening because `#click` is sent to WebDriver element, not the Watir one.

I am not sure why we don't just use `#options` to select proper value. I guess because of performance, but still maybe it'd be more reliable to work with Watir elements rather than WebDriver ones.

PR #379, opened 2015-10-09
Element#enabled? should raise when element does not exist

p0deje:
Since #363, we have an inconsistent behavior of `#enabled?` across different elements:

1. `Element#enabled?` returns `false` if element is not present.
2. `Button#enabled?` raises `UnknownObjectException` if element is not present (same for other `Element` children implementations).

To me, it makes perfect sense to raise `UnknownObjectException` if element is not present, so I've decided to that it would be better to define the correct behavior on `Element` class and remove all children implementations (though watirspecs are still subject to be removed).

I don't want to push this directly to master without notifying @Rodney-QAGeek and @titusfortner. Since this method is used by `#when_enabled` decorator, it might break somethings for you guys. However, I believe the correct solution is to chain decorators, e.g. `element.when_present.when_enabled.click`.

### DIFF
--- a/lib/watir-webdriver/elements/button.rb
+++ b/lib/watir-webdriver/elements/button.rb
@@ -34,16 +34,6 @@ module Watir
       end
     end
 
-    #
-    # Returns true if this element is enabled.
-    #
-    # @return [Boolean]
-    #
-
-    def enabled?
-      !disabled?
-    end
-
     private
 
     def locator_class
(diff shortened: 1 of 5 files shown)


Rodney-QAGeek:
I have no problems updating my tests with this new syntax! Previously my thinking was there wouldn't be a scenario where I would call when_enabled without the precondition of it being present.

titusfortner:
Sorry I didn't have a chance to look at this last week. The subclasses were doing an attribute check on enabled, which is a different from the wire call to enabled. I'm not sure if one could be true while the other is false in a way that could break backward compatibility. In general I like the consistency approach, though.

PR #380, opened 2015-10-10
Fix random failures on Travis

p0deje:
This pull requests changes watir-webdriver documentation to use Github instead of Watir.com because the latter might be slow/inaccessible which causes doctests failures on Travis.

I'm also trying to debug random failure of window switching spec for Chrome that happen on Travis, but no luck so far. @titusfortner If you could take a look, that'd be great.

### DIFF
--- a/lib/watir-webdriver/browser.rb
+++ b/lib/watir-webdriver/browser.rb
@@ -65,7 +65,7 @@ module Watir
     # Goes to the given URL.
     #
     # @example
-    #   browser.goto "www.watir.com"
+    #   browser.goto "github.com/watir/watir"
     #
     # @param [String] uri The url.
     # @return [String] The url you end up at.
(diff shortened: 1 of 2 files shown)


titusfortner:
Sorry, I don't remember seeing this in October. This looks good to me.

p0deje:
Chrome window specs are still failing randomly. Do you have any ideas because I cannot reproduce them locally. 



Cheers,Alex

titusfortner:
@p0deje - Yes, I know exactly what is going on there. I'm overhauling the specs on Selenium to document non-compliance reasons and to work on all browsers/platforms local & remote, and I'm updating/verifying that the watir specs work with everything as well in the process. I've been at it for a couple months now, but I think I'm close to finishing. :-)

titusfortner:
These are fixed now

PR #386, opened 2015-11-24
Added methods to wait for page load

westlakem:
Created methods to help wait for page loads on dynamic websites and SPAs. These methods will help usability when testing sites with ajax or angular included in them.


westlakem:
Forgot to pass browser

Issue #260, opened 2014-05-28
Request support for AngularJS specific attributes to watir-webdriver


anilreddy:
Presently we has lot of angular js applications to test and we are using protractor for automating it.

Protractor is not a matured framework and we need to have a javascript knowledge for it.

So, can u please extend watir-webdriver support by adding angular attributes like:

  by.binding
  by.model
  by.repeater
        .row
        .col

PR #387, opened 2015-11-24
Added methods to wait for page load

westlakem:
Added methods to be used in conjunction with after_hooks to wait for browser rendering to complete.

They are:
wait_for_page_load
wait_for_angular_completion
wait_for_ajax_completion

and the associated code to go along with them.

### DIFF
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -148,7 +148,7 @@ module Watir
         #angular not used in the application, continue as normal
       end
     end
-    
+
 
     #
     # Make calls to the browser waiting for ajax to complete
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -4,7 +4,8 @@ require 'watir-webdriver/wait/timer'
 module Watir
   module Wait
 
-    class TimeoutError < StandardError ; end
+    class TimeoutError < StandardError;
+    end
 
     INTERVAL = 0.1
 
@@ -83,6 +84,61 @@ module Watir
 
     end # self
   end # Wait
+
+  module PageLoad
+
+    #
+    # Waits for all page requests to finish before acting on the browser
+    # To be used in conjuction with after_hooks
+    #
+    # @example
+    #   browser.after_hooks.add {Watir::Wait.wait_for_page_load}
+    #
+
+    def wait_for_page_load(timeout)
+      wait_for_angular(timeout)
+      wait_for_ajax(timeout)
+    end
+
+    #
+    # Wait for angular requests to complete before continuing
+    # To be used in conjunction with after_hooks
+    #
+    # @example
+    #   browser.after_hooks.add {Watir::Wait.wait_for_angular}
+    #
+
+    def wait_for_angular(timeout = 30)
+      angular_element = "document.querySelectorAll('[ng-app]')[0]"
+      begin
+        @browser.execute_script("angular.element(#{angular_element}).scope().pageFinishedRendering = false")
+        @browser.execute_script("angular.getTestability(#{angular_element}).whenStable(function(){angular.element(#{angular_element}).scope().pageFinishedRendering = true})")
+      end
+      Watir::Wait.until(timeout, "waiting for angular to render") {
+        @browser.execute_script("return angular.element(#{angular_element}).scope().pageFinishedRendering")
+      }
+    rescue Selenium::WebDriver::Error::InvalidElementStateError
+      #angular not used in the application, continue as normal
+    end
+  end
+
+  #
+  # Wait for all ajax request to complete before continuing
+  #
+  # @example
+  #   Watir::Wait.wait_for_ajax
+  #
+  # currently only jQuery Ajax is supported
+  
+  def wait_for_ajax(timeout=30)
+    if @browser.execute_script('return (typeof jQuery != "undefined")')
+      if @browser.execute_script('return jQuery.active') > 0
+        Watir::Wait.until(timeout, "waiting for ajax") {
+          @browser.execute_script('return jQuery.active') == 0;
+        }
+      end
+    end
+  end
 
   module Waitable
     def wait_until(*args, &blk)
--- a/lib/watir-webdriver/wait.rb
+++ b/lib/watir-webdriver/wait.rb
@@ -92,53 +92,78 @@ module Watir
     # To be used in conjuction with after_hooks
     #
     # @example
-    #   browser.after_hooks.add {Watir::Wait.wait_for_page_load}
+    #   browser.after_hooks.add {Watir::PageLoad.wait_for_page_load}
     #
 
-    def wait_for_page_load(timeout)
-      wait_for_angular(timeout)
-      wait_for_ajax(timeout)
+    def self.wait_for_page_load(timeout = 30)
+      lambda do |browser|
+        self.wait_for_angular(timeout, browser)
+        self.wait_for_ajax(timeout, browser)
+      end
     end
 
     #
-    # Wait for angular requests to complete before continuing
-    # To be used in conjunction with after_hooks
+    # Waits for all angular actions to finish before acting on the browser
+    # To be used in conjuction with after_hooks
     #
     # @example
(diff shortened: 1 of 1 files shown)


westlakem:
Closed pull request due to mistake in begin/rescue placement

westlakem:
Reopening pull request as begin/rescue placement has been fixed.

titusfortner:
@westlakem - Thanks for writing this up.

I'm a big fan of having this functionality available, and this might be the right place for it. I know it's been discussed the option of having an angular specific gem as an add-on to Watir. I also know the Page Object gem handles this kind of thing by allowing users to designate a framework for the pages they are on. In general I think Watir over-delegates to Page Object gems, but that is a much longer conversation. @p0deje @bret @rubytester - we should discuss how we want to approach these kinds of things.

bret:
I'm reluctant to pull this into Watir-webdriver, and would rather suggest that this be published as a separate library. As far as I can tell, there is no technical reason why this needs to be part of Watir, since it only amounts to new, independent methods that are called via Watir's existing after_hooks method.

Also there are no unit tests for this, so Watir team would be hard pressed to take over responsibility for maintaining this.

westlakem:
@bret I appreciate the feedback, and this could be made into a separate library if that is the route you wish to go with.  The reason I included it into Watir was because the existing Wait methods were present there as well (EX: wait_until_present, when_present, wait_until, etc), and I thought the functionality was around the same line as those methods.

The problem with writing unit tests for these methods is that they all rely on a JavaScript callback from the browser or are just utility methods to call those functions.  I understand the want for unit tests, but unless I spin up a browser and inject a dummy html page that includes an ajax call and some angular directives, i'm not sure how this could be tested.

titusfortner:
Our [current specs](https://github.com/watir/watirspec) spin up a Sinatra server for testing. I'm pretty sure the overhead for jquery is low, but I'm guessing Angular is a bit higher.

westlakem:
@titusfortner I looked at watirspec and honestly could not tell you how that worked.  I also did a google search for examples and directions / api guide, but nothing popped up.  I cannot write tests without documentation on how to use the tool.

titusfortner:
Yeah, that implementation is not the most straight forward. The Sinatra server gets started through an overcomplicated series of commands and then it loads the files from [here](https://github.com/bootstraponline/angular_automation). Looks like there are already a couple of pages that import jquery, so testing that shouldn't be too complicated? The angular stuff will be much harder, which is why it might warrant a separate repo. I know [a lot of stuff was done in Ruby](https://github.com/bootstraponline/angular_automation), (but would need a significant overhaul to be useful), and I've been meaning to look at [what Ed Manlove did for Python](https://www.youtube.com/watch?v=qbMtbA5spSY) 
I'd love to discuss these things with the other devs at the [Test Automation Bazaar](https://www.facebook.com/events/924324614311200/)

p0deje:
I agree with @bret that this should be made a separate library:

1. Why should Watir know anything about Angular?
2. Why should Watir know anything about jQuery?

There are already [several libraries](https://github.com/watir/watir#other) providing extensions to different Watir APIs, this can and should be done separately.

I would actually consider merging this if it relied on some standard JavaScript API, but not on particular JavaScript framework.

titusfortner:
How do we accomplish this without requiring another project to Monkey Patch Watir?

p0deje:
No need to monkey patch, it can be new methods on browser instance, so it's just Watir::Browser class extended.

titusfortner:
Technically, adding new methods to an existing class is monkey patching, which is one of the things I'd like to see us avoid people needing to do outside of the project.

I'd rather see these methods put into separate files with an optional include for projects that use either jquery or angular. Or separate folders, because it would also be nice to add non-html angular selectors into the project a la #260.

Again, I'd like to minimize common usages for CSS and completely remove the need for any XPath in the project. Angular is popular enough that I'm a fan of expanding Watir to support it, even if it has to be included separately.

p0deje:
We're stepping on a very fragile path if we start pulling Angluar/jQuery/whatever.js code into Watir. Regardless of having this as a separate folder or not.

We can re-iterate on that, but I don't see a reason why this should be in Watir itself and not `watir-angular`/`watir-jquery` gem. IMO provide mechanisms for extending Watir is the approach we settled on, didn't we?

westlakem:
@p0deje 
I'm with @titusfortner for one simple reason.  What happens when you need multiple? What if (I've seen it done, not in my current project thankfully) you would need watir-angular, watir-jquery and watir-react? I have no problem making these separate libraries, but watir needs to sit as the common interface between them, not have them extend watir.  Maybe some kind of optional include if the project includes the other libraries?

titusfortner:
We can create a Watir::EventListener class that lets you do something like:

``` ruby
EventListener.new.listener.add(:after_wait) do |browser, timeout| 
    Watir::Angular.wait(timeout, browser)
    Watir::JQuery.wait(timeout, browser)
end
```

My only consideration here is whether we should expect users to find this gem somewhere out in the world and decide they need to use it, or host it as a separate gem in the Watir project, or include it in a separate folder in the Watir project directly. 

My thoughts on "extensibility" right now is that it allows for separation of concerns, but I don't think that means you have to relegate the code to other gems. It does mean that if someone wants a Watir::React or Watir::Ember they can be added as well in a non-required but available way, without needing separate gems. If we need to at some point deprecate support (like we've been doing with things in the Selenium project), and it shouldn't affect the rest of the project at all. I don't see value in keeping the Watir project "small" for the sake of it, so long as we don't make it "convoluted."

Another consideration is whether this should live in PageObject code instead, in which will be another discussion as to inclusion or non-inclusion in the Watir repository. :)

westlakem:
@titusfortner 

I don't think this should be in PageObject at all... PageObject is a design pattern around how to define elements.

titusfortner:
PageObject is a design pattern to abstract details of the UI details of a given page. That includes elements, actions, and state-based attributes (i.e. this page includes angular, this page includes jquery, this page includes both, this page includes neither). If you notice Cheezy's Page Object gem implements support for [frameworks](https://github.com/cheezy/page-object/blob/master/lib/page-object/javascript_framework_facade.rb). It might make sense to follow his lead on this rather than adding it directly to the Watir code. Or it might not.

bendizen:
I really like the ability to add these as per Titus's EventListener.  It would be great if you could add whatever you wanted to an event hook.  Some older proprietary tools like TestComplete allowed you to add any arbitrary code to to various event hooks.  I think it would be awesome if the Angular and JQuery waits were just part of Watir project if the maintenance isn't too much.  But the baseline ability of being able to add any custom code to fire in an event listener class would be a fantastic addition.

lib/watir-webdriver/wait.rb
```diff
@@ -84,6 +85,87 @@ def run_with_timer(timeout, &block)
...
+        #angular not used in the application, continue as normal
+      end
+    end
+
+
+    #
+    # Make calls to the browser waiting for ajax to complete
+    #
+
+    def self.wait_for_ajax(timeout=30, browser)
```
akostadinov, line 157: sounds like better name would be `wait_for_jquery`, just a nitpick and unrelated to the current discussion

titusfortner:
@westlakem I've finally put something together for Angular:
https://github.com/titusfortner/watir_angular

Lets move further discussion on this issue there. It isn't exactly polished, so by all means raise issues and give feedback.

I'm not sure creating an entirely separate gem makes sense for jQuery, since it is just the timeouts, not the locators as well, but maybe?

The equivalent code can be added to a project with these two methods:

```ruby
def wait_for_jquery(browser, timeout: Watir.default_timeout)
  return unless browser.execute_script('return (typeof jQuery != "undefined")')
  return unless browser.execute_script('return jQuery.active') > 0
  msg = "waiting for jQuery queue to empty"
  browser.wait_until(timeout: timeout, message: msg) { |b| b.wd.execute_script('return jQuery.active') == 0 }
end

browser.after_hooks.add ->(browser) { wait_for_angular(browser) }
```

I still kind of think that this behavior should be addressed at the page level and not the global test level, but maybe once we get #587 working in Watir 7.0, we'll find we have more flexibility to handle more advanced operations in Page Objects.

PR #392, opened 2016-01-11
Refactoring of locators

p0deje:
The following pull request refactors huge `ElementLocator` and all its children to a more atomic separate classes. The end result is far for being perfect, but even in this shape it is more clear and readable:

1. Split element locator to locator, which is injected with selector builder and element validator classes.
2. Separate XPath and CSS selector building.
3. Namespace new classes under element class.
4. Dynamically lookup new classes so that any element can override default behavior (As a side effect, I had to introduce `Cell` and `Row` classes).

The changes don't modify locator logic, only reorganizes it so that responsibility are more clear and follow-up refactoring can be done.

I also made `Element#element_class` purely dynamic since in all cases it can be calculated from element collection class name.

This pull request also removes the long deprecated case when `textarea` can be located with `text_field`. Related watirspec changes will be pushed once it's merged.

@titusfortner Please take a look. If you are fine with the changes, I'm going to merge them.

P.S. That's going to be a subject of my talk at Bazaar 2016 - the way watir-webdriver locates elements.


titusfortner:
I definitely love PRs that remove more code than they add. :)

1. I'm not going to have enough time to parse this before the weekend
2. I think any major code changes like this should wait until after the other PRs I just suggested
3. I'd actually rather see us just ditch the prefer_css code. I think Dave did a good job [showing that there is no performance gain](http://elementalselenium.com/tips/32-xpath-vs-css), and I'd like to remove configurations rather than try to figure out if someone is using them when trying to help them.
4. I don't actually think this is the right way to deal with text_fields

But I'm super excited to get to talk to you about all of this in person this weekend!

lib/watir-webdriver/locators/element/locator.rb
```diff
@@ -0,0 +1,247 @@
...
+          element = find_first_by_one
+        else
+          element = find_first_by_multiple
+        end
+
+        # This actually only applies when finding by xpath/css - browser.text_field(:xpath, "//input[@type='radio']")
+        # We don't need to validate the element if we built the xpath ourselves.
+        # It is also used to alter behavior of methods locating more than one type of element
+        # (e.g. text_field locates both input and textarea)
+        element_validator.validate(element, @selector) if element
```
titusfortner: I think we can restrict this more to just the elements that need it. Should get a small performance improvement on every location call.

lib/watir-webdriver/locators/element/selector_builder/css.rb
```diff
@@ -0,0 +1,63 @@
+module Watir
+  class Element
+    class SelectorBuilder
+      class CSS
```
titusfortner: I think we should get rid of the prefer_css toggle entirely. There is no performance improvement and even with this refactor I think it unnecessarily complicates things.

lib/watir-webdriver/element_collection.rb
```diff
@@ -89,16 +89,33 @@ def to_a
...
+      selector_builder = selector_builder_class.new(@parent, @selector, element_class.attribute_list)
+      locator = locator_class.new(@parent, @selector, selector_builder, element_validator)
+
+      @elements ||= locator.locate_all
     end
 
-    # overridable by subclasses
     def locator_class
-      ElementLocator
+      Kernel.const_get("#{element_class}::Locator")
```
titusfortner: We should use #const_get on the immediate parent of the element class rather than using Kernel

titusfortner:
Oh and before I get too far nit-picking this... Thanks for this PR. This reorganization was sorely needed. I think this was the most confusing part of the Watir codebase, and I'm excited that we can get it more accessible.

lib/watir-webdriver/elements/checkbox.rb
```diff
@@ -50,8 +50,5 @@ def checkboxes(*args)
   end # Container
 
   class CheckBoxCollection < InputCollection
```
titusfortner: We should change this to one of the "good" approaches, no preference from me on which - https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#single-line-classes

p0deje:
The last commit pretty much summarizes the final version of this refactoring. There is still a lot of things to improve, but it's mergeable in this state and custom locators engines can be implemented. Once it's merged and we can make a release, I'll be upload to release watizzle as well, but so far you can take a look at [github](https://github.com/p0deje/watirsome).

The latest commit namespaces all locators under `Watir::Locators` and makes it overridable making locators lookup purely dynamic.

```ruby
Watir.locator_namespace = Watizzle::Locators
```

It's not neither real dependency injection nor a good example of composition, but at least allows to create custom locator engines without monkey patching Watir itself.

@titusfortner It would be great if you could have a look at this PR and also at the way Watizzle works (as an example of locator-engine) and share your thoughts. I believe we can merge it like this since the PR is getting way to huge.

p0deje:
@titusfortner I guess you're pretty busy, so I'm going to merge this PR in a couple of days unless I hear a response from you :)

lib/watir-webdriver/element_collection.rb
```diff
@@ -89,16 +89,37 @@ def to_a
...
+      selector_builder = selector_builder_class.new(@parent, @selector, element_class.attribute_list)
+      locator = locator_class.new(@parent, @selector, selector_builder, element_validator)
+
+      @elements ||= locator.locate_all
     end
 
-    # overridable by subclasses
     def locator_class
-      ElementLocator
+      Kernel.const_get("#{Watir.locator_namespace}::#{element_class_name}::Locator")
```
titusfortner: Shouldn't we use `Watir.const_get` to maintain the namespace?

lib/watir-webdriver/locators/element/locator.rb
```diff
@@ -0,0 +1,249 @@
+module Watir
+  module Locators
+    class Element
+      class Locator
+        attr_reader :selector_builder
+        attr_reader :element_validator
+
+        WD_FINDERS = [
```
titusfortner: What do you think about using the [constant defined in Selenium](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/search_context.rb#L25)?

lib/watir-webdriver.rb
```diff
@@ -62,6 +56,19 @@ def prefer_css=(bool)
...
+      @locator_namespace ||= Watir::Locators
+    end
+
+    #
+    # Whether the locators should be used from a different namespace.
+    # Defaults to Watir::Locators.
+    #
+
+    def locator_namespace=(mod)
+      @locator_namespace = mod
```
titusfortner: Do we want to be able to use a class with a different namespace than `Watir`?

lib/watir-webdriver/element_collection.rb
```diff
@@ -89,16 +89,37 @@ def to_a
...
+    rescue NameError
```
titusfortner: Wouldn't it be better to use `defined?`instead of rescuing error?

lib/watir-webdriver/element_collection.rb
p0deje: We can't because `Watir.locator_namespace` is not necessarily defined under `Watir` namespace. For example, watizzle uses `Watizzle::Locators`.

lib/watir-webdriver/locators/element/locator.rb
p0deje: I think we can use it, why not. I'll update the PR.

lib/watir-webdriver.rb
p0deje: Exactly.

```ruby
Watir.locator_namespace = Watizzle::Locators
```

lib/watir-webdriver/element_collection.rb
p0deje: To use defined, we should first cast string to constant using `const_get` which raises `NameError`.

Another approach would be something like this, but I'm not sure if it's better than existing one since we make more than one call:

```ruby
def locator_class
  if Kernel.const_defined?("#{Watir.locator_namespace}::#{element_class_name}::Locator")
    Kernel.const_get("#{Watir.locator_namespace}::#{element_class_name}::Locator")
  else
    Kernel.const_get("#{Watir.locator_namespace}::Element::Locator")
  end
end
```

lib/watir-webdriver/locators/element/locator.rb
p0deje: On the other hand, it's Selenium private API so I'm not sure Watir should depend on it.

titusfortner: Out of curiosity, is there a reason it is private? I guess `SearchContext` is a weird module to reference things from.

p0deje: Don't know. We can make it public, but let's keep that for followup refactoring.

titusfortner: Oh, completely agree. Just musing.

PR #394, opened 2016-01-13
final release as watir-webdriver

titusfortner:
My Proposed Plan:

1. Merge this
2. Deploy the final watir-webdriver gem on rubygems
3. Rename existing watir github repo to watir_meta & update the Readme about it being deprecated; it does not need a version change or deploy to rubygems, just a commit for the Readme
4. Rename watir-webdriver github repo to watir
5. Merge #395 
6. Add Note to Readme and gemspec of watir-classic that it is no longer actively maintained with a recommendation to `require watir`and deploy to rubygems
7. Steaks for everyone!

PR #395, opened 2016-01-13
Watir transition

titusfortner:
This should not be merged until after we follow the steps I outlined in the previous PR, and have renamed this github repo to watir

Most of the noise in this PR is from getting rid of the git submodule and pulling all of the watirspec files directly into the project. There aren't any actively maintained projects that use watirspec, so having them in a separate repository is unnecessarily complicated.

I've followed [this guide](http://x3ro.de/2013/09/01/Integrating-a-submodule-into-the-parent-repository.html) to maintain commit history (if not the commit hash) on all of the watirspec code.

The final 5 commits here are mostly housekeeping now that everything is in one repo and renamed.

This isn't everything that I think we need for the big Watir 6.0 release, unfortunately. I've done a lot more overhauling of the specs to support running Edge and Marionette and better comments around why certain specs can't be run on various browsers... but that code isn't ready yet. 

I'm doing this PR now so we can discuss this week (before/during the conference) about how we want this to proceed.


titusfortner:
@p0deje - I updated this PR to not include the pulling in of the watir-spec code until we evaluate how things should be organized to support testing the sizzle code. (also, where does that code live, I can't find it).

Please double check I didn't miss anything. I renamed watir-webdriver to watir, updated the copyrights, and changed the naming of WebDriver to Selenium in most instances. As far as naming things goes, WebDriver should be used for the W3C specifications, and Selenium should be used for the code that implements it, which is what Watir is actually using. For instance, the WebDriver spec [only supports 4 locators](https://w3c.github.io/webdriver/webdriver-spec.html#h-element-retrieval), so it is more accurate to call them selenium locators, etc.

But let me know if you think anything here should get changed. Thanks!

lib/watir/extensions/alerts.rb
```diff
@@ -5,7 +5,7 @@ module Watir
   #
   # Module provided by optional require:
   #
-  #   require "watir-webdriver/extensions/alerts"
+  #   require "watir/extensions/alerts"
```
p0deje: Maybe we can drop it with new release? It has been deprecated for ages.

p0deje:
Seems okay to me, though we're renaming main gems, so I'd like to get @jarmo and @jarib as the gem owners are now different for their opinion.

jarib:
Fine by me. Do whatever you think is best.

watir-webdriver.gemspec
```diff
@@ -1,36 +1,37 @@
...
+  s.name        = 'watir'
+  s.version     = '6.0.0'
   s.platform    = Gem::Platform::RUBY
-  s.authors     = ["Jari Bakken"]
-  s.email       = ["jari.bakken@gmail.com"]
-  s.homepage    = "http://github.com/watir/watir-webdriver"
-  s.summary     = %q{Watir on WebDriver}
-  s.description = %q{WebDriver-backed Watir}
-  s.license     = 'MIT'
+  s.authors     = ['Alex Rodionov', 'Titus Fortner']
```
jarmo: If I would be @jarib I would still like to see my name here, because after all he is the original author. No? :)

jarmo:
Ok, as I understand then the main idea is to rename `watir-webdriver` to `watir`.

1) Is there an upgrading guide somewhere?
2) What is the plan with current `watir` Github project? Rename it to something else?
3) What is the plan with current `watir` and `watir-classic` gems? Shouldn't we release at least one more `watir` with a deprecation warning or something? I don't think that `watir-classic` is impacted by this change directly (except when it is used as a driver through current `watir` gem).
4) Should we add methods from current `watir` to new `watir` to raise a better `Exception` than `NoMethodError`? I mean public API methods in [here](https://github.com/watir/watir/blob/master/watir/lib/watir/loader.rb).
5) I think that all existing watir-related libraries will work, but would be better to make sure ([watir-rails](https://github.com/watir/watir-rails), [watir-rspec](https://github.com/watir/watir-rspec)).

No other comments right now :)

watir-webdriver.gemspec
titusfortner: I'll definitely put it back if @jarib wants. I considered that one more "contact info," than "copyright info" (which I obviously didn't change).  His call.

jarib: No worries, I've also considered it a list of active maintainers.

lib/watir/extensions/alerts.rb
titusfortner: Well, this is the conversion not necessarily the new release. I'd actually like to incorporate your locator changes as part of Watir 6.0 release to differentiate it from the final watir-webdriver release (and especially if we make non-backward compatible changes). I agree this would be a good time to remove old alert implementation.

titusfortner:
@jarmo - I spelled out the process here: https://github.com/watir/watir-webdriver/pull/394
 Let me know if that isn't specific enough. 

As for a deprecation notice on watir before the rename... I guess it would affect people who are versioning their gemfiles with ~>5 instead of leaving it open, telling them to upgrade. We can do that. Mostly though, we should probably be ready with a tutorial or explanation for people who start posting "why my tests are breaking with Watir 6.0?" 

It would be great if we could get the website updated first... @zeljkofilipin said he could convert our Wordpress sites into a jekyll site so we can all do updates/PRs much more easily. But he also said he's especially busy right now.

akostadinov:
Now that old firefox driver does not work with firefox 47 I think it is ultimate to have at least a beta version gem that can work with the `marionette` driver. Any existing workarounds?

I have tried the following to no avail:
```
        # @browser = Watir::Browser.new :firefox, :profile => firefox_profile, :http_client=>client, marionette: true
        # @browser = Watir::Browser.new :firefox, marionette: true
        # @browser = Watir::Browser.new :marionette
```

titusfortner:
Firstly, Selenium still supports Firefox 47, there is a bug in Firefox 47.0, so we're waiting for 47.01patch that fixes it.
And this should work to start a browser:
`b = Watir::Browser.new :firefox, marionette: true`
Marionette support is still only 75-80% of Legacy Driver.

akostadinov:
@titusfortner , a major issue is that marionette does not support untrusted certs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1103196
And I don't see that actively being addressed unfortunately. Also I don't know how to add certs on the fly, I hope there is a way, otherwise it's almost a show stopper unless one has very static test harness and test executing machines.

titusfortner:
Like I said, Marionette is only 75-80% functional right now. Certs aren't top priority for them. You can use Firefox 45 (the current ESR) with legacy driver, or try chromedriver.

akostadinov:
Misread last comment, sorry. Now I understand 47.0.1 will be supported by legacy driver. I hope legacy will be supported until marionette is feature complete. Thank you.

titusfortner:
This PR has been updated and rebased

1. Includes #394 
2. Passes with these specs - https://github.com/watir/watirspec/pull/84 on OSX
3. Requires Selenium 3.0 

The plan is to release this as 6.0.0.beta1 and do a production release when Selenium does the production release of 3.0. We could wait until 3.0 is released to production, but I think we've waited long enough for the selenium team, and I'd rather see a beta with this code than nothing. :)

@p0deje - I believe this code is ready to go unless you have any issues with it.

p0deje:
👍 

I'm still not sure how we're going to address watir/watir Github project, rename it something else or what?

p0deje:
Nevermind, see plan in #394

PR #409, opened 2016-03-10
native events not needed

titusfortner:
Selenium no longer supports any Firefox browsers with Native Events. The one spec that was blocked for synthetic events works just fine on Firefox 45. IE supports the toggle, but we don't have any tests for which it applies.


p0deje:
:+1:

titusfortner:
watirspecs were failing. Looks to be expecting firefox_args to always return driver & args, even when args are empty, fixing & rerunning.

PR #410, opened 2016-03-10
support testing with remote standalone server

titusfortner:
Many people use Watir on AWS or Saucelabs, so we should test our code through the remote server.

### DIFF
--- a/spec/implementation.rb
+++ b/spec/implementation.rb
@@ -7,6 +7,7 @@ class ImplementationConfig
 
   def configure
     set_webdriver
+    start_remote_server if remote? && !ENV["REMOTE_SERVER_URL"]
     set_browser_args
     set_guard_proc
     add_html_routes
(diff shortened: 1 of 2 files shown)


spec/implementation.rb
```diff
@@ -16,6 +17,33 @@ def configure
...
+    at_exit { @server.stop }
+  end
+
+  def remote_server_jar
+    require 'open-uri'
+    file_name = "selenium-server-standalone.jar"
+    return file_name if File.exist? file_name
+
+    open(file_name, 'wb') do |file|
+      file << open('http://goo.gl/qTy1IB').read
```
p0deje: Let's make this an environment variable so that it can be overridden (and more explicit).

spec/implementation.rb
```diff
@@ -16,6 +17,33 @@ def configure
...
+                                     :port       => Selenium::WebDriver::PortProber.above(4444),
+                                     :log        => !!$DEBUG,
+                                     :background => true,
+                                     :timeout    => 60)
+
+    @server.start
+    at_exit { @server.stop }
+  end
+
+  def remote_server_jar
```
p0deje: Can't [Selenium download](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/server.rb#L39-L42) it?

titusfortner:
Ah, yet another cool Selenium feature. I added the ENV variable, but some of these ENVs seem overkill since as far as I know only you and I are running these things, and we aren't using most of them. :) One watirspec needs updating for getting browser name "chrome" instead of "remote," otherwise everything passes the remote tests.

titusfortner:
merged

PR #414, opened 2016-03-20
fix issues with stale elements

titusfortner:
Looking at how these tests were stubbed, I noticed some issues.

1) Result of element going stale between assert exists and action should be the same as other times an element goes stale - (UnknownObjectException not StaleElementReferenceError)

2) Removing an element from the DOM isn't accurately modeling an element going stale, so I don't even know what that is attempting to test.

3) Due to performance issues we no longer check for staleness from the top. We check the bottom element and re-look everything up if that one is stale.

### DIFF
--- a/spec/element_spec.rb
+++ b/spec/element_spec.rb
@@ -66,46 +66,15 @@ describe Watir::Element do
       browser.goto WatirSpec.url_for('removed_element.html')
     end
 
-    it "does not propagate StaleElementReferenceErrors" do
-      button = browser.button(id: "remove-button")
-      element = browser.div(id: "text")
-
-      expect(element).to exist
-      button.click
-      expect(element).to_not exist
-    end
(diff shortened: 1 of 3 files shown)


spec/element_spec.rb
```diff
@@ -66,51 +66,21 @@
...
-
-      browser.refresh
-      expect(stale_element.present?).to be Watir.always_locate?
-    end
   end
 
   describe "#element_call" do
 
     it 'handles exceptions when taking an action on an element that goes stale during execution' do
+      Watir.always_locate = true
```
p0deje: You need to reset this after test or it is going to affect all other tests.

titusfortner:
Whoops. I put that in there to make sure it worked for both, it wasn't supposed to stay in. I'll fix it. Thank you.

PR #416, opened 2016-03-20
FramedDriver methods match browser methods

titusfortner:
There is no reason to use an atom here, when when we can make the same selenium call. Also, no reason not to implement text here as well.


p0deje:
:+1: to merge this though we need some specs for it.

PR #433, opened 2016-07-21
implemented after hooks with selenium event listeners

titusfortner:
DO NOT MERGE - 
This code monkey patches Selenium for it to work, but I wanted to show off this proof-of-concept for how we can use Selenium's Event Listeners to do our after hooks. 

I think we'll actually want to create some kind of WatirEventListener sub-class to allow more extensible options for adding and removing before & after hooks, along with the current defaults.


titusfortner:
Was proof of concept code only.

Issue #447, opened 2016-08-22
Regex Conversion Engine


titusfortner:
We discussed in [this PR](https://github.com/watir/watir/pull/331) making additional improvements to our current regular expression conversions.

Specifically, by relying more extensively on XPath's "contains" functionality, we could drastically decrease the number of element text calls we make when doing text searches by regular expression. This could make text regex calls significantly more performant.

titusfortner:
I think I've mostly done everything I want to with this one after #822 is merged.

We won't be able to allow text locator to leverage `SelectorBuilder::XPath#predicate_conversion` until we do the deprecations in Watir 7. Also, we probably should figure out a solution to #787 before closing this.

titusfortner:
Everything is good except for 787, no need to track it in both places

Issue #343, opened 2015-05-26
Unable to select nested text element with text selector


titusfortner:
Since text selector returns the top most text node for text nested inside multiple text nodes, it is not possible to click the specific sub-element with the text using a text matcher.

Based on the conversation in [PR 65](https://github.com/watir/watirspec/pull/65), this might be "won't fix" but I wanted to document it as an issue.

p0deje:
Can we close it now?

Issue #342, opened 2015-05-26
Different exist/visible results for string and regexp


titusfortner:
From conversation in [PR 65](https://github.com/watir/watirspec/pull/65)

A selector of {text: String} is translated into an xpath selection and returns the first text node that contains the string, or the top-most text node if the text value is inside a nested text node. If a node that contains the text is nested inside the selected node and has the display set to "none," #visible? will return true, even when false.

A selector of {text: Regexp} gathers an array of WebDriver::Element and iterates through it to match the regular expression with the Element#text of each element. Since #text only returns what is visible on the page, it will return false to #exists? where a String match would return true.

I'm not sure which behavior should match which. I'm leaning toward the Regexp approach of returning false for #exists? if it is not visible on the page, even if it is in the DOM. I can write a spec if we agree this is desired behavior.

p0deje:
As far as I know, it's not possible to make XPath/CSS expression that would consider visibility of node. With that said, I don't think we can make the behavior with String/Regexp consistent at all.

Do you have any ideas on how to implement it?

titusfortner:
I'm not sure how sizzle would work.

Couldn't we just implement text matching the same way we do regexp matching? Something like:

```ruby
      txt_selector = selector.delete :text
      how, what = build_wd_selector(selector)
      elements = parent.find_elements(how, what)
      elements.__send__(method) { |el| el.text == txt_selector }
```

p0deje:
Should it be applied for `:text` only? What about other attributes? I feel like there are other cases like this.

It will also make text selector so much slower than it is now. As a heavy user of `:text`, I'm not a great fan of performance degradation.

titusfortner:
I... don't think this applies to other attributes? Let me consider it after I get some sleep.

Regarding performance, you definitely make a good point. Our regexp implementation is *slow, and that makes sense if you are doing a #text call for every matching element (e.g. every div on a page, etc). Maybe we can make one extra call if it is specifically a text selector just to verify it is present. 

```Ruby
        idx ||= 0
        el = @wd.find_elements(how, what)[idx]
        el == nil if selector.key?(:text) && el.text != selector[:text]
```

PR #448, opened 2016-08-22
Nested text nodes

titusfortner:
I was thinking about this guy when [discussing it on stackoverflow](http://stackoverflow.com/questions/39027914/does-watir-webdriver-have-a-wildcard-tag/39028316#39028316)

This is my proposed solution for: #343 

Instead of returning the first text node that matches the text, this code takes that element and iteratively finds the first child until no more matching child is present.
1. This much more closely follows the Principle of Least Astonishment
2. for `text: String` when the nested element is hidden (and the text is not actually displayed on the page), it will now always locate the element, but correctly return false for `#present?`
3. for `text: Regexp` when `#can_convert_regexp_to_contains?` is true, the same behavior as string 
4. for `text: Regexp` otherwise, it will still not even locate an element with text in the DOM that is not visible on the page (so this PR does not solve #342, the fix for which is in #447)
5. This PR now allows access to specific nested child elements without relying on xpath(see [this failing spec](https://github.com/titusfortner/watir-webdriver/blob/master/spec/click_spec.rb#L13-L16) that now passes). 
6. Currently `browser.element(text: /foo/)` returns the html element if the text exists on the page. After this PR it should return the element expected, though it could be significantly non-performant. I thought about adding a warning to that effect, but if someone wants something slow that's up to them to determine and at least it will work after this PR.
7. Note: this is not backward compatible, but I would not expect this to be a significant issue.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -465,17 +465,17 @@ module Watir
 
       if tag_name == "input"
         klass = case elem.attribute(:type)
-          when *Button::VALID_TYPES
-            Button
-          when 'checkbox'
-            CheckBox
-          when 'radio'
-            Radio
-          when 'file'
-            FileField
-          else
-            TextField
-          end
+                when *Button::VALID_TYPES
+                  Button
+                when 'checkbox'
+                  CheckBox
+                when 'radio'
+                  Radio
+                when 'file'
+                  FileField
+                else
+                  TextField
+                end
       else
         klass = Watir.element_class_for(tag_name)
       end
@@ -548,10 +548,18 @@ module Watir
       selector_builder = selector_builder_class.new(@parent, @selector, self.class.attribute_list)
       locator = locator_class.new(@parent, @selector, selector_builder, element_validator)
 
-      locator.locate
+      located = locator.locate
+
+      @selector.key?(:text) ? locate_nested_text_node(located) : located
     end
 
     private
+
+    def locate_nested_text_node(located_parent)
+      return nil unless located_parent
+      sub_parent = self.class.new(@parent, element: located_parent)
+      self.class.new(sub_parent, @selector).locate || located_parent
+    end
 
     def locator_class
       Kernel.const_get("#{Watir.locator_namespace}::#{element_class_name}::Locator")
--- a/spec/click_spec.rb
+++ b/spec/click_spec.rb
@@ -7,13 +7,16 @@ describe Watir::Element do
     }
 
     let(:clicker) { browser.element(id: "click-logger") }
-    let(:log)     { browser.element(id: "log").ps.map { |e| e.text } }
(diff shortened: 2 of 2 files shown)


jkotests:
Does this only change the `:text` locator?

If so, I have mixed feelings about the nested text node change. I agree that the result of `browser.element(text: /foo/)` is not likely what someone will want - ie the `html` element is rarely useful. On the other hand, I thought the recommendation was to not have things like `expect(browser.text).to include?('foo')`. Having `browser.element(text: /foo/)` should similarly be used carefully.

My biggest concern is that the `:text` locator will behave differently than the rest. For example, it sounds like for the HTML:

```
<div id="foo1" class="foo1">
  <div id="foo2" class="foo2">foo</div>
</div>
```

Will return the inner `div` when locating by text:

```
browser.element(text: /foo/).id
#=> "foo2"
```

And return the outer `div` when locating by class:

```
browser.element(class: /foo/).id
#=> "foo1"
```

I am not sure I can follow the code for when you mix a `:text` locator with another locator:

```
browser.element(text: /foo/, class: /foo/).id
#=> ?
```

It will be awkward to have to explain to people that we locate the first matching element, but if a `:text` locator, we match the deepest element.

titusfortner:
Ah, see I consider the deep-nested one to be actually the first matching. In most cases it will return the first element that has the text actually between the closing of its opening tag and the opening of its closing tag.

In your example, Most users would not expect `browser.div(text: /foo/)` or `browser.div(text: 'foo')` to match the top div at all (especially if it were much further up in the DOM where they couldn't see it).

I think the current behavior is already substantially different: to select with an attribute, the attribute has to be inside the specific html element, but to select with text, it just has to be nested somewhere beneath it in the DOM. If anything this PR actually brings the text location approach more in line with the attribute approach.

As for mixing locators... it finds the first element that matches all the parameters, then if the selector includes "text", it uses that element as the parent to locate the first child element that matches all the parameters, etc until it finds a matching element with no children that match all the parameters.

jkotests:
I view the order of elements as that returned by:

```
browser.elements
```

To me, locating an element is iterating through that collection and asking if the element meets the specifiers. I would expect `browser.element(text: 'foo')` to be equivalent to:

```
browser.elements.find { |e| e.text == 'foo' }
```

titusfortner:
@jkotests thanks for the discussion. After more thought I realize that this is not the right solution.
This code results in:

`browser.element(text: /foo/) != browser.elements(text: /foo/).first`

To solve this problem the right way, we would need to fix: `browser.elements(text: /foo/)`

Essentially the way I'm looking at this is (assuming that there is no text that adds to "foo") I would expect this to be true:

`browser.elements(text: /foo/)  == browser.elements(text: "foo")`

We could accomplish something much closer with a find_elements call using:

`browser.elements(xpath: "//*[contains(text(), 'foo')]")`

I'm not sure what other difficulties that would entail, but it would need to be part of whatever work we do in #447

Issue #424, opened 2016-05-10
can we get an element collector:  grand parent / great grand parent


royroque:
can we get an element collector:  grand parent / great grand parent

I'm working on code when there is no unique name/id of the element; so as a workaround I end up using parent multiple times.

i.e.1
instead of
browser.label(:text=>'Username').parent.parent.text_field.set 'James'
USE:
browser.label(:text=>'Username').grandparent.text_field.set 'James'
browser.label(:text=>'Username').gparent.text_field.set 'James'
browser.label(:text=>'Username').parents[1].text_field.set 'James'

i.e.2
instead of
browser.label(:text=>'Username').parent.parent.parent.text_field.set 'James'
USE:
browser.label(:text=>'Username').greatgrandparent.text_field.set 'James'
browser.label(:text=>'Username').ggparent.text_field.set 'James'
browser.label(:text=>'Username').parents[2].text_field.set 'James'

titusfortner:
Unfortunately have 2 different notions of `parent` in our code right now:
```
table = browser.table(id: 'foo')
td_element = table.td(id: 'bar')

td_element.parent.class == TableRow
td_element.instance_variable_get("@parent").class == Table
```

So, to start with, we should agree how we want to name things here.
Should `td_element.grandparent.class == Browser` or  `td_element.grandparent.class == Table` ?

I think `@parent` probably needs to get renamed to `@scope` or `@parent_scope`

The array format doesn't seem easy/desirable from an implementation standpoint, but I've been thinking about how we should switch over from the atom to use xpath for `#parent`. We could do something like this:
```
def ancestors(val)
  e = element(xpath: "./ancestor::*[#{val}]")
  e.exists? ? e.to_subtype : nil
end

def parent
  ancestors(1)
end

def grandparent
  ancestors(2)
end
```
I guess we could `#method_missing` the various `great`'s and `gg`'s as well...

p0deje:
To me using lots of `parent` when locating elements is a code smell meaning that HTML should be fixed to be more testable. However, we still allow chaining parent calls, so I don't see a reason to have a bunch of new methods for this.

👎

titusfortner:
I don't think we should assume that the tester of a website has control over the html under test.

p0deje:
Okay, but I still believe that we should not complicate API with grand parents. This can easily be extended using separate gem or locally. I also don't see a problem with chaining parent calls.

p0deje:
@titusfortner What's up with this?

[...]

PR #451, opened 2016-09-04
Parents & Siblings with XPath

titusfortner:
This PR meets all of my proposed [criteria for new features](http://watirtight.com/2016/08/25/focus-of-watir.html#evaluation).
[Associated watirspecs PR](https://github.com/watir/watirspec/pull/85)
1. Being able to easily locate elements is one of the core purposes of Watir. While this feature idea was rejected by Jari in #270, I'd love for Watir to have a mantra of "No More XPath!" It would be great if there weren't any use cases for when XPath is the best way to locate an element. I think I've previously addressed his other two concerns with this alternate implementation. (Also of note is that the new `parent` implementation addresses Jari's 3rd concern for siblings).
2. Our current javascript implementation causes issues when trying to evaluate the presence of a parent element. You can not call `#exist?` on a parent element that doesn't exist. See [this Stack Overflow question](http://stackoverflow.com/questions/38577957/rspec-and-watir-webdriver-assert-element-not-present/) for an example of when this is a problem faced by a real user.
3. This solves #424 to my satisfaction. With XPath we don't need to make it recursive, and `parent(1)` is better than needing to include `grandparent` and `grandpa` and `granddad`, etc. :)
4. Demand for this comes both from Blackbaud which made extensive usage of the feature and the people commenting in [Justin's blog](https://jkotests.wordpress.com/2013/08/21/locate-a-sibling-from-a-watir-element/)
5. Also, since these elements (including parent) are not initialized with Selenium Element objects, they can be re-looked up if the browser is refreshed or they go stale, etc.


jkotests:
What do you think about making the locator support more than just index?

For example, I have used multiple `parent` calls when I have:

```
<tr>
  <td>
    <a href="#">click me</a>
  </td>
  <td>
    <span>some known text</span>
  </td>
</tr>
```

I might need to do:

```
browser.span(text: 'some known text').parent.parent.link.click
```

The current pull request shortens the code to:

```
browser.span(text: 'some known text').parent(1).link.click
```

However, what I am really looking for is not just 2 parents up. I am specifically looking for the parent row, which happens to be 2 parents up at this point. As a result I would suggest supporting locators instead of just an index:

```
# An example using :tag_name
browser.span(text: 'some known text').parent(tag_name: 'tr').link.click

# In the event you really just want 2 parents up, :index can still be used
browser.span(text: 'some known text').parent(index: 1).link.click
```

To me, allowing all of the standard locators would allow the Watir code to be more descriptive and less brittle to changes.

titusfortner:
@jkotests - hah, I started to implement this exactly, then I thought it might be confusing / unnecessary... 

This actually made me rethink how I implemented #455... These are essentially all the same family of features.

I don't think we need to support anything more than index & tag_name unless there are any use cases I'm not thinking of...

titusfortner:
I re-implemented this and rolled in the child/children functionality as well. This now has a consistent API across sibling/parent/child interactions.

p0deje:
I still don't understand how this aligns with your goals, so maybe we should discuss that on the next call.
1. "Improve test reliability." I guess we all agree that hierarchy-based locators are very unstable.
2. "Make writing maintainable tests easier." The fact that tiny hierarchy change forces to rewrite the locators (and we need to actually see the new difference) doesn't improve maintenance.

titusfortner:
I get what you are saying, but by this logic we should remove all implementations with `index`. I've worked in places where both `text` locators and `class` locators had to be constantly updated because the app was changing so much. Instability of locators is very app-specific. Many of the sites I've needed to use this kind of thing with haven't changed in years and have very few uniquely identifying things in the DOM. Often the pattern is to find the element with some text and move to the element you need to click in relation to that text. In a lot of real-world sites, the alternative to that is a really intense XPath hierarchy anyway. I can definitely think of some additional helper methods I'd love to add for tables making use of these patterns.

I don't think this should be the first or even second option for locating elements, but it should be an option when needed, and it is obviously needed at times.

p0deje:
Let's discuss it on the next call - we have a different view on the problem :)

p0deje:
My proposal is to actually make it fully compatible with existing Watir API and support all the locators. Current solution "pretends" to support locators (e.g. tag_name, index), but then why won't it work with class, text, data-\* attributes or any other Watir locator.

For me, all the new methods should behave like `#element` and support all its locators.

I see this API like this:

``` ruby
browser.span.parent
browser.span.parent(index: 1)
browser.span.parent(class: 'foo')
browser.span.parents

browser.span.child
browser.span.child(index: 1)
browser.span.child(class: 'foo')
browser.span.children

browser.span.sibling             # next sibling
browser.span.sibling(index: 1)   # second sibling
browser.span.sibling(index: -1)  # previous sibling
browser.span.siblings            # all siblings
```

I'm not sure about `#sibling` as it's a bit confusing, need to think about it.

Also, does `#child` searches for immediate children or all of them? If all of them, than we can simply alias it to `#element`, it would be the same, won't it?

jkotests:
For `#sibling`, I think it would be less confusing if we stay with an explicit method for preceding vs next siblings. Using positive/negative `:index` values for next/previous will not be consistent with `#element`. Right now, `#element(index: -1)` returns the last matching element. Therefore, I would expect `#sibling(index: -1)` to return the last sibling (rather than the first preceding sibling).

I would also propose that we change the method names to match the XPath axis. It will make the methods more clear for those who do know XPath.
- `#preceding_sibling` instead of `#previous_sibling`
- `#following_sibling` instead of `#next_sibling`

titusfortner:
> does #child searches for immediate children or all of them?

This is the part that is slightly wonky, but it is Least Astonishment because this is the way XPath supports it (based on the symmetry in the `xpath_adjacent` method)... 

`#parent(index: 1)` walks up the DOM two levels. `#child(index: 1)` stays one level down in the DOM and returns the second element at that level.

@jkotests - I changed the names, but added an alias.

I also added a commit to be explicit about what locators are/are not supported. Because, as for supporting all of the locators... 

If the syntax is confusing, I'm more in favor of changing it than trying to expand it as if this were the same thing as any other locator.

If the syntax itself isn't confusing, but it feels incomplete to only support some of the locators... YAGNI? The feature request has been for the "index" functionality, and we recognized the value in filtering out a specific type of tag name because we can't do the "normal" Watir API approach of specifying the tag_name as the method name.

I'm not inherently opposed to supporting other locators, but I'd like to try to be more agile about this project and ship features that work even if they aren't as complete as we can make them. I'd rather provide feature that's been asked for, add the other things that make sense for us to do to the backlog for later prioritization, and then listen to users who try this code and come back with additional requests.

By which I mean, I have another dozen things I'm working on that I think are more important to this project than just ensuring symmetry in this feature, but I still think the main functionality of this feature is important. :)

jkotests:
To address the confusion between `#child` and `#element`, would it help to alias `#element` with `#descendant`? On the same line of thought, alias `#parent` with `#ancestor`?

titusfortner:
Wouldn't the confusion be in the opposite direction? Maybe child/children should be renamed to direct_child/direct_children

jkotests:
`#direct_child` would be more obvious when it comes to introducing the feature for the first time. Long term, it seems rather verbose. As much as we are pushing for not needing to use XPath and CSS-selectors, both have made the terms "child" and "descendant" more common. It seems reasonable that Watir uses the same terminology.

I was thinking that aliasing `#element` with `#descendant` would make it more explanatory in the documentation. I don't think I would personally ever use the `#descendant` over `#element`, so maybe not worth the alias.

titusfortner:
Ok, I see what you are saying, for documentation if someone looks for "what is child?" they see "ah, it's the same as direct_child, and so element is all descendants."

Yes, I'm in favor of these aliases. 

I still need to figure out the Yardoc to get this finalized.

lib/watir/elements/element.rb
```diff
@@ -1,3 +1,5 @@
+# encoding: utf-8
```
p0deje: Why add it?

spec/adjacent_spec.rb
```diff
@@ -0,0 +1,176 @@
+require File.expand_path('watirspec/spec_helper', File.dirname(__FILE__))
+
+describe "Adjacent Elements" do
```
p0deje: This should be in watirspec, isn't it?

lib/watir/elements/element.rb
titusfortner: This PR is based on code I started writing over a year ago, so likely a bad merge choice, I'll remove it.

spec/adjacent_spec.rb
titusfortner: Sure? I mean, I'm assuming everything will end up in the same place after we pull watirspecs into watir. I'm just putting new stuff in `spec/` because it is easier to run. ¯\_(ツ)_/¯

p0deje: It will still be in `spec/watirspec` directory. Knowing that you already have PR to watirspec, let's just move the file there.

titusfortner:
Next step for this is to add yardoc, then rebase

titusfortner:
Do we want `#siblings` and `#children` to return an Array of `Watir::Div` elements, or an `ElementCollection` of `Watir::Element` elements? I don't see an easy way to create the correct Collection type, but returning an Array of elements seems wrong.

lib/watir/adjacent.rb
```diff
@@ -0,0 +1,131 @@
...
+      xpath_adjacent('', opt)
+    end
+
+    private
+
+    def xpath_adjacent(direction = '', opt = {})
+      opt = opt.dup
+      index = opt.delete :index
+      tag_name = opt.delete :tag_name
+      caller = caller_locations(1,1)[0].label
```
p0deje: 1. move this into `unless opt.empty?`, this variable is not used anywhere else
2. `(1,1)` => `(1, 1)`

lib/watir/adjacent.rb
```diff
@@ -0,0 +1,131 @@
...
+      unless opt.empty?
+        raise ArgumentError, "unsupported locators: #{opt.inspect} for ##{caller} method"
+      end
+
+      xpath = "./#{direction}#{tag_name || '*'}"
+      return elements(xpath: xpath).map(&:to_subtype) unless index
```
p0deje: Maybe instead of returning array of elements we should just return `HTMLElementCollection` (i.e. we should not do `#to_subtype`). This will make methods consistent with existing Watir API.

lib/watir/adjacent.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      e = element(xpath: "#{xpath}[#{index + 1}]")
+      e.exists? ? e.to_subtype : e
```
p0deje: Same here, instead of checking for element existence, we just return element instance that is lazily located.

lib/watir/element_collection.rb
```diff
@@ -80,12 +80,38 @@ def last
...
-      # TODO: optimize - lazy element_class instance?
       @to_a ||= elements.map.with_index do |e, idx|
         element_class.new(@query_scope, @selector.merge(element: e, index: idx))
       end
     end
 
+    #
+    # Returns true if two element collections are equal.
+    #
+    # @example
```
p0deje, line 91: Let's add another example when method returns false.

lib/watir/element_collection.rb
```diff
@@ -80,12 +80,38 @@ def last
...
+    #   browser.select_list(name: "new_user_languages").options == browser.select_list(id: "new_user_languages").options
+    #   #=> true
+    #
+
+    def ==(other)
+      to_a == other.to_a
+    end
+    alias_method :eql?, :==
+
+    def +(other)
```
p0deje, line 113: I can't find where this method is used, why adding it?

p0deje (changes requested):

lib/watir/adjacent.rb
titusfortner: Hmm, so you think it would be better to have a generic collection with generic elements than an array of specific elements... What do you think of the idea of creating a new subclass called MixedElements? We can keep it lazy loaded and then do the `#to_subtype` call in the overridden `#to_a`? Or is that too complicated?

lib/watir/element_collection.rb
titusfortner: For `#siblings`. It's essentially `element.previous_siblings + element.following_siblings`, so there needed to be a plus operator.

titusfortner:
So I updated this with what you suggested. It is lazy loaded and returns a specific Collection type when a tag is specified, and a generic Collection when tag is not specified.

As an alternative, I also made this: https://github.com/watir/watir/commit/a001e5e03b81a4a7628a0df9e99adcc76825e8cf
which implements a MixedElementCollection. So you can lazy load into a generic collection, but when it creates the array of elements, each one has the correct subtype. If you want to proceed with that idea, I'll merge it to this and you can code review the whole thing.

p0deje (approved):

titusfortner:
This has already been merged into master: 904e32a2fe423b82cce60deaceecbe0729ea83ce

PR #452, opened 2016-09-05
deprecate #always_locate in favor of #stale? and #wait_until_stale?

titusfortner:
1. With performance close to equivalent, the only reason to use #always_locate is to rescue it in test code for flow control. We shouldn't be encouraging this as bad practice.
2. I'm not sure which is best in the #stale? method - `raise` or `warn` or `assert_exists`


lib/watir/elements/element.rb
```diff
@@ -493,6 +493,14 @@ def browser
       @parent.browser
     end
 
+    def stale?
+      raise Selenium::WebDriver::Error, "Can not check staleness of unused element" unless @element
```
p0deje: Why does it raise Selenium error?

lib/watir/wait.rb
```diff
@@ -246,5 +246,23 @@ def wait_while_present(timeout = nil)
...
+    # @example
+    #   element = browser.text_field(name: "abrakadbra").wait_until_stale
+    #
+    # @param [Fixnum] timeout seconds to wait before timing out
+    #
+    # @see Watir::Wait
+    # @see Watir::Element#stale?
+    #
+
+    def wait_until_stale(timeout = nil)
```
p0deje, line 259: 1. What's the use case?
2. Why don't you provide opposite method `#wait_while_stale`?

titusfortner: wait_while_stale makes no sense, because once an element goes stale it never comes back.

This is the only reason I can come up with to want to use #always_locate - when you are taking an action on an element until it goes stale and rescuing the exception as part of flow control.

lib/watir/elements/element.rb
titusfortner: Because I've been writing too much Selenium code recently... :-/

p0deje:
I don't like the fact we're adding so many `#wait_until_*` methods. Maybe that should be made a bit more Rubyish:

``` ruby
element.wait_until(&:present?)
element.wait_until(&:visible?)
element.wait_until(&:enabled?)
element.wait_until(&:stale?)
element.wait_until { text == 'foo' }
```

Same can be done for `wait_while`.

lib/watir/elements/element.rb
titusfortner: fixed this one...

titusfortner:
I'm not certain I'm following. You're suggesting we implement it so people use:

`element.wait_until(&:present?)`
instead of 
`element.wait_until_present`

If so, that is more Ruby, but it isn't nearly as obvious, so I'm not sure it solves my problem of getting people to use it more.

p0deje:
> If so, that is more Ruby, but it isn't nearly as obvious, so I'm not sure it solves my problem of getting people to use it more.

Why is it not obvious? That's one of the basic things you learn in Ruby when start to work with enumerable. It's also better because can be used for any boolean method and custom methods:

``` ruby
def visible_and_disabled?(element)
  element.visible? && element.disabled?
end

element.wait_until(method(:visible_and_disabled?))
```

We won't need to support and add so many methods and people will be able to use Watir API is a Ruby manner

titusfortner:
Your idea of basic is not mine. :) You'd be surprised how many people don't understand `&` and `#to_proc` notions. I have literally never used the `method` method, but... I definitely like the simplicity of it, so I'm on board with the change.

titusfortner:
Would we be able to do this with the decorators `when` and `while`?

p0deje:
> You'd be surprised how many people don't understand & and #to_proc notions.

I'm well aware of that, but we're developing "Web Application Testing in **Ruby**" and you want to "Help users become better programmers.". The first time new users understand this concept, Ruby becomes so much more fun to program for them.

Bloating the API sounds more like Java, not Ruby :)

> Would we be able to do this with the decorators when and while?

Why not?

titusfortner:
So if we needed a custom timeout, we'd need to do like this?

`element.wait_until(5) { |el| el.present?}`

p0deje:
No, it would be `wait_until(5, &:present?)`

titusfortner:
Oh, right, because that's the block from the `to_proc`. So, I've never used that before; very cool.

p0deje:
See how much fun it is to uncover Ruby :)

titusfortner:
If possible, I'd like to get rid of `#always_locate` toggle for 3.0.

I do think we should be able to point to polling for `stale?` with waits as we deprecate it. 

I think regardless of how we might implement something new in the future, we should be backward compatible with `#wait_until_*` syntax. If that is the case, then adding this extra one in there shouldn't be a big deal. 

Other than that concern, is there any additional reason to not merge this?

lib/watir/elements/element.rb
```diff
@@ -493,6 +493,14 @@ def browser
       @parent.browser
     end
 
+    def stale?
```
p0deje, line 504: Maybe we should add docs now that it's public API?

lib/watir/wait.rb
```diff
@@ -246,5 +246,23 @@ def wait_while_present(timeout = nil)
       # it's not present
     end
 
+    #
+    # Waits until the element is stale.
+    #
+    # @example
+    #   element = browser.text_field(name: "abrakadbra").wait_until_stale
```
p0deje: Please, remove `element =`

p0deje:
Should we now remove rescue of stale error in https://github.com/watir/watir/pull/452/files#diff-0ac8030a39db381bf5dd122f0bfdd337R245?

Let's merge in any case 👍

titusfortner:
@p0deje - it turns out what I was showing before was only half of what I'd intended it to be :-/
This is the actual deprecation of the feature, not just a deprecation warning. Let me know what issues you see. They pass all the specs in https://github.com/watir/watirspec/pull/88

lib/watir/elements/element.rb
```diff
@@ -493,6 +494,21 @@ def browser
...
 
+    #
+    # Returns true if a previously located element is no longer attached to DOM.
+    #
+    # @return [Boolean]
+    # @see Watir::Wait
+    #
+
+    def stale?
+      raise Watir::Exception::Error, "Can not check staleness of unused element" unless @element
```
p0deje, line 505: Maybe it should `assert_exists` instead? I can imagine usage of something like this:

``` ruby
div = browser.div(id: 'foo')
browser.scroll.to(:end) until div.stale?
```

Do I miss something in the implementation?

titusfortner: `assert_exists` would attempt to relocate it and so long as that locator was in the DOM it would never show up as stale. I'm not sure a way to do this with our lazy loading other than:

``` ruby
div = browser.div(id: 'foo')
div.exists?
browser.scroll.to(:end) until div.stale?
```

p0deje: This method behavior is definitely a bit different from others. I now understand why it is, so I guess we can keep it as is and see if users complain.

titusfortner:
As I was writing up a summary of what this change would mean, I realized there is one legitimate non-backward-compatible scenario with this change. I attempted to explain it here: http://watirtight.com/2016/09/11/always-locate.html

I just want to make sure we're on the same page with things before we push forward with this.

abaird:
Explanation looks good to me

titusfortner:
merged: b9b3ad865ac59a03d9eaee8ec95b84ee22bdc5b2

PR #453, opened 2016-09-05
shouldn't have to assume element is present when waiting for enabled

titusfortner:
@Rodney-QAGeek - was there a reason this wasn't included in your original PR?
I've found a few places where I needed to nest them, which I think should be default? `when_present.when_enabled`

### DIFF
--- a/lib/watir/wait.rb
+++ b/lib/watir/wait.rb
@@ -201,7 +201,7 @@ module Watir
       message = "waiting for #{selector_string} to become enabled"
 
       if block_given?
-        Watir::Wait.until(timeout, message) { enabled? }
+        Watir::Wait.until(timeout, message) { present? && enabled? }
         yield self
       else
         WhenEnabledDecorator.new(self, timeout, message)


lib/watir/wait.rb
```diff
@@ -201,7 +201,7 @@ def when_enabled(timeout = nil)
       message = "waiting for #{selector_string} to become enabled"
 
       if block_given?
-        Watir::Wait.until(timeout, message) { enabled? }
+        Watir::Wait.until(timeout, message) { present && enabled? }
```
p0deje: `present?`

titusfortner: Heh. oops. fixed.

titusfortner:
Thinking this over, I think it is actually our `enabled?` method that is wrong. This code doesn't actually verify the element is present like the comment says.

```
    # Returns true if this element is present and enabled on the page.
    #
    # @return [Boolean]

    def enabled?
      assert_exists
      element_call { @element.enabled? }
    end
```

Should we update this to:

```
    def enabled?
      assert_exists
      element_call { @element.displayed? && @element.enabled? }
    end
```

p0deje:
Why it should be visible? It can be enabled but not visible, isn't it?

titusfortner:
I guess why would we care about whether an element that isn't visible is enabled?

I mean, maybe we need some kind of `interactable?` or `usable?` method that combines these two. There should be some method that lets the user know that they are good to go to do what they want with an element without having to chain decorators.

RodneyU215:
Yeah, I'd say that having a method like `usable?` that checks both `present?` and `enabled?` is super valuable, but I'm not sure it should be baked into `enabled?`. Currently I've essentially chained the methods together by default in my PageObject gem. I'd love to remove it though if Watir implement's this.

p0deje:
In many cases it's pretty hard to understand if element is usable or not. For instance, there are tons of ways to make it unusable using JavaScript and we won't understand that until we try to interact with it.

titusfortner:
But we can do the basic checks of present && !disabled and !read-only where appropriate based on tags that have those attributes. `usable?` is probably better than `likely_usable?` or `dom_usable` :-D

There are other ways I'm contemplating dealing with this elsewhere, though, so maybe we don't need this here now.

titusfortner:
Let's revisit this later. Not a priority for right now.

PR #454, opened 2016-09-05
add #wait_until_text

titusfortner:
This is a common pattern I've seen and used. 

I'd like to see this supported:

`element.wait_until_text(5, "foo")`

instead of requiring:

`Watir::Wait.until(5) { element.text == "foo" }`

[Associated specs are here](https://github.com/watir/watirspec/pull/86)

### DIFF
--- a/lib/watir/wait.rb
+++ b/lib/watir/wait.rb
@@ -246,5 +246,24 @@ module Watir
       # it's not present
     end
 
+    #
+    # Waits until the text is present.
+    #
+    # @example
+    #   browser.text_field(name: "new_user_first_name").wait_until_text("First name")
+    #
+    # @param [String, Regexp] text value to wait for before timing out
+    # @param [Fixnum] timeout seconds to wait before timing out
+    #
+    # @see Watir::Wait
+    # @see Watir::Element#text
+    #
+
+    def wait_until_text(text, timeout = nil)
+      timeout ||= Watir.default_timeout
+      message = "waiting for #{selector_string} text to equal \"#{text}\""
+      Watir::Wait.until(timeout, message) { self.text =~ /#{text}/ }
+    end
+
   end # EventuallyPresent
 end # Watir


abaird:
I would change the order of the arguments if anything. `wait_until_text(5, 'foo')` just doesn't make sense to me. `wait_until_text('foo')` with an optional time argument is better. Still not sure if this belongs, but I suppose if we have other wait_until methods then this isn't that bad. What about `wait_while_text`?

titusfortner:
It's a common pattern that an additional convenience method would help without needing to use Watir::Wait (which many of the people I've worked with don't seem to use by default. The standard default pattern seems to be home-grown array iterations with sleeps. 

I put time first because that keeps it consistent with other methods, but you are right that does seem less intuitive to a user. So, yes, I should probably swap them.

jkotests:
Is the use case always waiting for the element to have an _exact_ text? Should we also support regular expressions to do partial matches?

titusfortner:
Changed parameter order, and added spec for getting it to work with regular expressions!

lib/watir/wait.rb
```diff
@@ -246,5 +246,24 @@ def wait_while_present(timeout = nil)
       # it's not present
     end
 
+    #
+    # Waits until the element is present.
+    #
+    # @example
+    #   browser.text_field(name: "new_user_first_name").wait_until_text
+    #
+    # @param [Fixnum] timeout seconds to wait before timing out
```
p0deje, line 256: incorrect `@param` order

lib/watir/wait.rb
```diff
@@ -246,5 +246,24 @@ def wait_while_present(timeout = nil)
...
+    # @param [String] text value to wait for before timing out
+    #
+    # @see Watir::Wait
+    # @see Watir::Element#text
+    #
+
+    def wait_until_text(text, timeout = nil)
+      timeout ||= Watir.default_timeout
+      message = "waiting for #{selector_string} text to equal \"#{text}\""
+      Watir::Wait.until(timeout, message) { self.text =~ /#{text}/ }
```
p0deje, line 265: why is it regular expression?

lib/watir/wait.rb
```diff
@@ -246,5 +246,24 @@ def wait_while_present(timeout = nil)
       # it's not present
     end
 
+    #
+    # Waits until the element is present.
```
p0deje: Incorrect docs :)

lib/watir/wait.rb
titusfortner: because you can put a string inside there and it will work; not tied to that implementation if there is something better.

titusfortner:
@p0deje - fixed the documentation issues. Thanks

p0deje:
What about `#wait_while_text`?

lib/watir/wait.rb
p0deje: Why not use exact string?

titusfortner: We want to be able to support both Regexp match and String match. This seemed like an easy way to do that.

titusfortner:
I've needed a `#wait_until_text` on many occasions, but never a `#wait_while_text`
Maybe we should look into metaprogramming these to automatically get both.

p0deje:
Yeah, let's just have `#wait_until(&blk)` :)

lib/watir/wait.rb
p0deje: Yeah, but it doesn't support exact strings.

titusfortner: Oh, I see what you are saying. Sigh. Right.

titusfortner:
Closing to implement with #458

Issue #329, opened 2015-04-01
Direct Children


titusfortner:
@alexhanh recently suggested that [this behavior](http://www.natontesting.com/2009/08/19/how-to-get-only-direct-child-objects-in-watir/) makes more sense for a testing framework.

Currently it can be achieved by doing: 
browser.ul(:id, 'main').lis(xpath: './*')

Does it make more sense for a testing framework to make this the default behavior?

Also, one of the strengths of Watir is the lack of need to use xpath or css to access elements. As such, is there a more user-friendly / obvious way we can let users toggle between these two behaviors?

Watir.all_descendants = true
Watir.only_children = true
browser.ul(:id, 'main').lis.children
browser.ul(:id, 'main').lis.descendants
browser.ul(:id, 'main').lis(children: true)
browser.ul(:id, 'main').lis(descendants: true)
browser.ul(:id, 'main').child_lis
browser.ul(:id, 'main').descendant_lis

p0deje:
My vote is :-1: for that because if we follow the children approach, the XPath is smth like `//ul[@id="main"/lis`, while currently it's `//ul[@id="main"//lis`. This means that if you want to skip few children, you can easily do it now, but if we change the approach, you will have to explicitly specify each children. This makes Watir API calls chain similar to writing XPath manually and makes tests brittle.

alexhanh:
I also don't know if the behaviour should be default. In some cases, it seems direct children is what people expect it to be, in other cases not. Also according to @p0deje it seems like the current way allows more flexiblity.

In my personal experiences, most of the times when I'm quering for 'lis', 'trs' and 'tds', I expect and need only children. But a crawler, for example, could very well want to query for all the 'as' as descendants.

But what I'm pretty sure to suggest is to have a lean Watir way to query just direct children without having to use XPath or CSS selectors. In my experience of using Watir, the need to query direct children comes often enough to justify a shortcut. Hopefully we'll get more opinions about this. So how the syntax could look like?

`lis.children` does this get all the `lis` and return all of their children (regardless of the tag type)? Seems maybe too ambiguous. Same argument applies to `lis.descendants`.

`lis(children: true)` better since the additional children restriction ties to the `lis` more clearly.

`child_lis` the most tidy and explicit one, I think. I guess the bracket style allows more parameters to be added in the future (can't come up with ideas what right now). But I guess you could do the same with `child_lis(something: false)` too.

PS. I've personally used watir-webdriver far more for automation purposes than testing.

[...]

PR #455, opened 2016-09-05
implement :children locator to restrict search for child element(s)

titusfortner:
This implements #329 
Removes the need for XPath to find direct child elements, and provides functionality that is the opposite of `#parent` method.

Demand for this feature comes from both [natontesting.com article](https://web.archive.org/web/20120823010857/http://www.natontesting.com/2009/08/19/how-to-get-only-direct-child-objects-in-watir) and @alexhanh

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -128,6 +128,10 @@ module Watir
 
           if selector.key? :index
             raise ArgumentError, "can't locate all elements by :index"
+          elsif selector.key?(:children) && (selector.key?(:xpath) || selector.has_key?(:css))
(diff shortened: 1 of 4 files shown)


abaird:
Maybe I'm missing something - does this actually implement the behavior `browser.div(id:'foo').divs(:children)`?

titusfortner:
The code implements

`browser.div(id:'foo').divs(children: true)` # ==> all immediate children

and

`browser.div(id:'foo').div(children: true)` # ==> does not match unless it is an immediate child

titusfortner:
Closing this in favor of what I've updated in #451

PR #456, opened 2016-09-06
remove getAttribute atom in favor of Selenium call

titusfortner:
The getAttribute atom has been maintained and updated in the selenium project, but stayed static in Watir for several years. The existing browser drivers have been using the selenium atom to return values for `@element.attribute` so technically Watir isn't returning the latest and greatest implementation for what attribute is. All specs pass with this implementation.

@p0deje - do you have any idea why it was implemented this way?

Also we're going to need to figure out how to support the w3c property vs attribute distinctions, but that will be later.


p0deje:
I have no idea why it was like that in the first place, but if specs are passing I'm happy to merge it.

PR #457, opened 2016-09-06
better selector_string

titusfortner:
# selector_string is used in error messages, so it is useful to have the full nested value returned, instead of just the single most nested selector.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -551,6 +551,14 @@ module Watir
       locator.locate
     end
 
+    def selector_string
+      if @parent.is_a?(Browser)
+        @selector.inspect
+      else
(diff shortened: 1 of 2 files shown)


lib/watir/elements/element.rb
```diff
@@ -576,7 +576,8 @@ def element_class_name
     end
 
     def selector_string
-      @selector.inspect
+      return @selector.inspect if @parent.is_a? Browser
+       "#{@parent.send :selector_string} --> #{@selector.inspect}"
```
p0deje: indentation is wrong

p0deje: make selector_string protected, not private and just call `@parent.selector_string`

titusfortner:
Updated

titusfortner:
This needs more tlc, I'll re-open after writing specs

PR #460, opened 2016-09-09
change default from assert_exists to when_present for actions

titusfortner:
I've written and re-written this code a dozen times in the past year and a half, and I've held off on a PR it because I wasn't sure about how to have the discussion on moving things in this direction. But, I think this is the single most important feature for Watir to add at this point.

@Rodney-QAGeek - this is stuff I was alluding to last year, and I kind of wanted to dig through your Pizza code for comparison before posting this, but I think it is time to get this out there and discussed.

The basic idea is that there should never be a reason that your code specifies to take an action on a window, alert or element and expect that action not to happen. These waits do not affect the predicate methods, so flow control should be handled by querying predicate methods, not by rescuing errors like `UnknownObjectException`.

This is the default behavior for most frameworks (including Capybara). I've seen many projects (mine included) that monkey patch these action elements to do `#wait_until_present` first. This is a big part of the notion of a test library "doing what I mean" instead of "what I say." The framework should not be so anxious to give up, but should be patient briefly to make sure that it can't fulfill the intention of the code. This is becoming increasingly essential with more interactive javascript on the page.

I've also created a method for turning off these waits on a per action basis: `browser.without_waiting {}`

This PR is completely backwards compatible. All tests that are passing now will still pass, but some might drastically increase their running time if the test code rescues the errors like `UnknownObjectException` to use for flow control. I've added warnings to the code indicating that there might be performance issues. I've added a warning to the use of `when_present` to point out when it is likely unneeded.

Right now this is implemented with a basic toggle. Once we implement a WatirEventListener class we can make this much more finely tuned (allow specifying specific behavior for specific actions). But what I have here I believe should be the default. I think this feature should be implemented before the Listener class, though.

Oh, and I'm not sold on the idea of `relaxed_locate` as a name if anyone has other suggestions.

Updated specs are in https://github.com/watir/watirspec/pull/87


lib/watir/elements/element.rb
```diff
@@ -495,6 +479,48 @@ def browser
...
+      return assert_exists unless Watir.relaxed_locate?
+
+      timeout ||= Watir.default_timeout
+      return if exists? # Performance shortcut
+
+      assert_element_found unless Watir.always_locate? # don't re-lookup element
+      timeout_off = timeout == 0
+
+      begin
+        @parent.wait_for_exists(timeout)
```
jkotests, line 492: Does this mean that chained element calls can wait longer than the specified timeout? For example, `browser.div.link.click` can:
- Wait up to 29 seconds for the `div` and then (assuming it appears)
- Wait up to another 30 seconds for the `link`?

I would expect the timeout to be applied to the entire action, which to a user is just clicking the link.

p0deje:
That reminds me all the discussions around implicit vs explicit waiting in Selenium 😄 

I definitely need to think about it, but here are few points out of my head:
1.  it slows down failing test execution significantly (i.e. if you have 10 tests failing to locate element that adds 300 seconds to tests execution).
2. Watir starts to swallow all the flaky tests which provide a lot of useful information about SUT stability, performance, random bugs and tests architecture like it provides now (I know that many argue it should work like that).
3. implementing listener/before_hooks should not be harder that this but at the same time more flexible, meaning we can provide strategies that can be enabled/disabled (e.g. present, enabled, etc.) It aligns more to the idea of extensibility we keep talking about.
4. Necessity of `#without_waiting` seems like a smell to me, meaning that the concerns has to reevaluated.

lib/watir.rb
```diff
@@ -14,58 +14,55 @@
...
     #
 
-    def default_timeout=(value)
-      @default_timeout = value
-    end
-
-    def prefer_css?
-      @prefer_css
+    def default_timeout
+      @default_timeout ||= 30
```
jkotests, line 47: Maybe we should differentiate the default timeout for implicit vs explicit waits. I am concerned that a default timeout of 30 seconds punish many users. 
- Consider a new user, who is learning Watir. They will likely get the element locators wrong on many occasions. However, each time it is wrong, they will see the browser sitting there for 30 seconds. This is a long enough time that the users will think that Watir has crashed or is slow.
- For those dealing with simpler pages (ie not JavaScript heavy pages), missing elements will greatly slow down test suites.

For implicit waits, I suggest a value closer to the 2 seconds, which is what Capybara uses.

lib/watir/elements/element.rb
titusfortner: A previous implementation used the Timeout class to set a total limit on it. The problem is when you set the timeout to 0 (or 1 which wouldn't be handled as a special case) and the combination of calls times it out when it shouldn't.

I decided that this is unlikely to be a real world problem. Yes, if it takes 29 seconds for the parent to exist, then 29 seconds for the element to exist, then another 29 seconds for the element to become displayed, then you are in trouble. 

Under most scenarios if the first call passes even after a substantive time, everything else will quickly pass as well. The worst case practical scenario should be an element taking a long time to exist, then never becoming present, which would time out at in under a minute.

All of that being said... I wonder if it would be an issue if we did:

``` ruby
  Watir::Wait.until(timeout) {@parent.wait_for_exists(timeout) && exists?}
```

It should still run through everything at least once before failing if the timeout was set to 0 or 1.

titusfortner:
I want to be very clear that this is not an implicit wait.
An implicit wait is implemented as part of the JSON Wire Protocol to be handled in the driver itself. It applies to both predicate calls and action calls and this only applies to action calls. When calling `@element.exists?` on a non-existent element with an implicit wait, it will wait the full timeout before returning false. This implementation immediately returns false.

This has the benefits of implicit wait without that downside. This is the same as calling `#when_present` or in some cases effectively implements a `#when_exists` method on every element before taking an action. The only time it adds extra wait time is when you expect the element to maybe not be there, try to take an action on it, then rescue the exception as part of your flow control strategy. By adding the warnings, we are encouraging people to not use exceptions as flow control. And most importantly their tests will still pass.

To @p0deje's specific points:
1. If tests fail this way, at lest the user knows they are real failures and not false positives. Results you can rely on is more important than length of test suite which only increases when there are real failures (or as previously discussed exceptions as flow control). Many people rerun failing tests and this would drastically reduce the need to do that. I'd also rather see slower tests that fail less often for users that don't know to add `#when_present?` everywhere on their JS heavy sites.
2. I vehemently disagree that Watir is the right tool for testing performance or the stability of the environment. Most people do not run Watir in production and have a much less powerful and much less stable environment which, while bad, is not testing what they are intending to test.
3. I'm only saying that if we agree on this functionality, I'd rather get it working for 3.0 if we can, then re-implement with more fine-tuned control later. I made a stab at WatirEventListener, and I do not believe it to be trivial.
4. I don't think I'd ever use `#without_waiting` and I'd be completely fine with not including it, I just thought it would be nice to add something to give per-call control rather than implementing things like: `#click_no_wait`. But maybe that is preferable, or maybe no one will want it and we should remove it until/unless people complain about it.

lib/watir.rb
titusfortner: These are not implicit waits (see distinction in main conversation)

The whole point of this is to be able to remove `#when_present` scattered everywhere in a code base, which is how most people have to do it right now. 

`#default_timeout` is the default for how long you wait for an element to show up when you might need to wait for it, and I think it would only be confusing to add a new variable that from a user/non-implementation standpoint does the same thing. Either 30 seconds is too long for everything and we should adjust the default downward, or this implementation should default to the same 30 seconds.

p0deje: These are implicit waits in a way that if element is not present, it waits for it to become present automatically.

titusfortner: I'm only saying that they are not "Implicit Waits" in the context of the conversations that were had about them when WebDriver came out. Several of the concerns here would be more applicable to the WebDriver notion of implicit waits than what I'm suggesting.

titusfortner:
Thanks for all the feedback. Between the conversations here and at the Homebrew meetup, I did a significant rewrite of this for https://github.com/watir/watir/pull/479

Issue #458, opened 2016-09-06
More Versatile Waiting Methods


titusfortner:
As discussed in #452 & #454, we should make it easier to call waits on elements.

Along the lines of:

``` ruby
element.wait_until(&:present?)
element.wait_while(&:visible?)
element.wait_until(&:enabled?)
element.wait_until(&:stale?)
element.wait_until { text == 'foo' }
element.when(&:enabled?).click
```

titusfortner:
@p0deje - I'm not sure how to implement this so this in order to keep the error message without having to pass it in as a parameter. Did you have an idea?

p0deje:
I don't think there is any way to keep error messages the way they are now. Let's just dump them.

Also, one example I've posted is invalid: 

``` ruby
element.wait_until { text == 'foo' }
```

should be

``` ruby
element.wait_until { |e| e.text == 'foo' }
```

titusfortner:
I'm definitely not in favor of dumping the messages. If anything I'd like to see them made more specific and useful for debugging.

What about metaprogramming these? `define_method` the list of predicate methods the given element responds to; then we can use the `method` that is passed in as part of the message.

p0deje:
Guess it's already done.

PR #461, opened 2016-09-09
Wait metaprogrammed

titusfortner:
So this would need documentation and probably some more tests around the edge cases, but this is what I was talking about in our meeting last night as far as supporting `wait_until_*`, `wait_while_*` and `when_*` for the available predicate methods to address #458. 

The first commit passes existing specs just fine with :define_method
The second commit supports dynamic `wait_until_*_and_*` methods with :method_missing
This might be way overkill if the only real use case we have is `present_and_enabled?`

@p0deje - does this look worth pursuing? Or is this down the wrong track?

### DIFF
--- a/lib/watir/alert.rb
+++ b/lib/watir/alert.rb
@@ -1,7 +1,7 @@
 module Watir
   class Alert
 
-    include EventuallyPresent
+    include ConditionalWaits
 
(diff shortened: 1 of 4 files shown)


titusfortner:
Tabling for now in favor of #479

PR #462, opened 2016-09-11
Deprecate Watir#prefer_css

titusfortner:
@p0deje - So, I followed your template for Watizzle to implement [WatirCss](https://github.com/titusfortner/watir_css). It is passing everything in watirspecs.

If this looks good, I'd suggest moving the repo/gem to Watir maintained and then merging this PR to deprecate `Watir#prefer_css` toggle (which I'm guessing people likely don't use, anyway).

At least on Linux, the performance is [usually worse](https://travis-ci.org/watir/watir/builds/159034274), anyway.

We can do more performance tests on Windows if we're really concerned about it.

What do you think?


p0deje:
Looks good to me!

I'm not sure about performance tests, guess it would be great to have them. Also, I think this change worths blog post with explanation on watir.github.io.

titusfortner:
Awesome. I understand what you did with Watizzle better after implementing this. 

Since this looks good, I'll organize things (move watir_css, push the gem, etc), then release a new Watir beta and do a blog writeup.

titusfortner:
merged bf3b9b5b8506655d32fa4ca5ebe83b3d0ffbb2ec

Issue #446, opened 2016-08-20
Image methods


titusfortner:
Is there a reason we have [these methods](https://github.com/watir/watir/blob/master/lib/watir/elements/image.rb#L41-L54) defined if they aren't supported? Are we ok to just delete them?

titusfortner:
Resolved with #463

PR #463, opened 2016-09-11
remove non-supported methods from Image

titusfortner:
As referenced in #446

I think these were implemented in watir-classic, and shouldn't need to stay in the project with exceptions. Goes along with: https://github.com/watir/watirspec/pull/89


p0deje:
:+1:

titusfortner:
merged - e510617887e0934629219c94773c00b0f2da7be4

PR #464, opened 2016-09-12
update #selector_string to include full path of elements

titusfortner:
this is a re-opening #457. 
This will provide better error messages when waiting for an element times out.


p0deje:
👍

titusfortner:
merged - 44d35ff5db623147fc663696e4a679eb01a38e6d

PR #465, opened 2016-09-12
implement generic wait methods

titusfortner:
I'm not sure if this is what we were looking for earlier. I think I like the second one better here if that's what we're looking for. Additional specs in: https://github.com/watir/watirspec/pull/90

```
Watir::Wait.until { element.text == ‘foo’ && element.value == ‘bar’}
element.wait_until { |el| el.text == ‘foo’ && el.value ==  ‘bar’}
```

I should also add that this solves my concern with #454 and that I still think we need `#wait_until_present` `#wait_while_present` and `#wait_until_stale`

This would be an alternative to the metaprogrammed option in #461 
I might like that solution better, though.

### DIFF
--- a/lib/watir/wait.rb
+++ b/lib/watir/wait.rb
@@ -177,7 +177,7 @@ module Watir
       message = "waiting for #{selector_string} to become present"
 
       if block_given?
-        Watir::Wait.until(timeout, message) { present? }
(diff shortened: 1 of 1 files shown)


titusfortner:
Implemented in #479

PR #466, opened 2016-09-13
implement visible filter

titusfortner:
I was going to call this :hidden, but I see that Capybara implements this as :visible, which makes more sense.

This comes up as more and more sites implement as Single Page Apps. For instance there might be a single login page that has modals for both signing in and signing up. They share the same classes and names for elements in the form, but one is visible and the other is hidden. Sometimes it is easy to grab an id from the form or the top element of a modal and do something like:

``` ruby
modal = browser.div(id: "sign-in-modal")
username = modal.text_field(name: "username")
```

Sometimes it is not easy. This would be easier than using indexes, and would allow you to use the same page object (versus subclassing) for both of these when the names of the fields and the actions for each are the same.

Also, I can't figure out how to get the mocks to work in the `element_locator_spec` with this updated implementation, so I'm going to need help with that. :)

The specs need html elements found here: https://github.com/watir/watirspec/pull/91

### DIFF
--- a/lib/watir/locators/element/selector_builder.rb
+++ b/lib/watir/locators/element/selector_builder.rb
@@ -30,6 +30,10 @@ module Watir
             unless what.is_a?(Fixnum)
               raise TypeError, "expected Fixnum, got #{what.inspect}:#{what.class}"
             end
+          when :visible
+            unless what.is_a?(TrueClass) || what.is_a?(FalseClass)
+              raise TypeError, "expected TrueClass or FalseClass, got #{what.inspect}:#{what.class}"
+            end
           else
             unless VALID_WHATS.any? { |t| what.is_a? t }
               raise TypeError, "expected one of #{VALID_WHATS.inspect}, got #{what.inspect}:#{what.class}"
@@ -53,7 +57,7 @@ module Watir
 
         def normalize_selector(how, what)
           case how
-          when :tag_name, :text, :xpath, :index, :class, :label, :css
+          when :tag_name, :text, :xpath, :index, :class, :label, :css, :visible
             # include :class since the valid attribute is 'class_name'
             # include :for since the valid attribute is 'html_for'
             [how, what]
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -93,19 +93,27 @@ module Watir
           selector = selector_builder.normalized_selector
 
           idx = selector.delete(:index)
+          visible = selector.delete(:visible)
+
           how, what = selector_builder.build(selector)
 
           if how
             # could build xpath/css for selector
-            if idx
-              @query_scope.wd.find_elements(how, what)[idx]
+            if idx || !visible.nil?
+              idx ||= 0
+              elements = @query_scope.wd.find_elements(how, what)
+              elements = elements.select { |el| visible == el.displayed? } unless visible.nil?
+              elements[idx] unless elements.nil?
             else
               @query_scope.wd.find_element(how, what)
             end
           else
             # can't use xpath, probably a regexp in there
-            if idx
-              wd_find_by_regexp_selector(selector, :select)[idx]
+            if idx || !visible.nil?
+              idx ||= 0
+              elements = wd_find_by_regexp_selector(selector, :select)
+              elements = elements.select { |el| visible == el.displayed? } unless visible.nil?
+              elements[idx] unless elements.nil?
             else
               wd_find_by_regexp_selector(selector, :find)
             end


p0deje:
That is actually quite interesting. At some point I was thinking about something like that - allowing predicate method locators (`visible: true`, `selected: 'text'`, etc.) but decided to implement it in [watirsome](https://github.com/p0deje/watirsome#locators) and keep using that. Having something like that in Watir might actually help users, so I have nothing against merging it.

lib/watir/locators/element/selector_builder.rb
```diff
@@ -33,6 +33,10 @@ def check_type(how, what)
             unless what.is_a?(Fixnum)
               raise TypeError, "expected Fixnum, got #{what.inspect}:#{what.class}"
             end
+          when :visible
+              unless what.is_a?(TrueClass) || what.is_a?(FalseClass)
```
p0deje: typical Rubymine indentation :D

titusfortner:
@p0deje can you help me figure out how to change the mocks in `element_locator_spec`? I'm bad at mocks. :(

lib/watir/locators/element/selector_builder.rb
titusfortner: 'doh

spec/element_hidden_spec.rb
```diff
@@ -0,0 +1,65 @@
+require File.expand_path('watirspec/spec_helper', File.dirname(__FILE__))
+
+describe Watir::Locators::Element::Locator do
```
p0deje: This should go to watirspec, isn't it?

titusfortner: Yup but again, Travis won't show you if they pass if they are in watirspec. :) I planned to move before merging.

p0deje: Ah, got it.

titusfortner: But thanks for reminding me; I have to keep all these PRs straight. :-D

p0deje:
@titusfortner Can't figure it out quickly, are you okay if I do this later today by myself and commit?

titusfortner:
I'm definitely ok if you do it and commit. I'll take a look at what you do and learn. :)

titusfortner:
It's because I implemented everything as a `#find_elements` with `#select` calls; no more just `#find_element` options.

p0deje:
Right, let's get back to find_element, there is no need to retrieve all elements all the time.

titusfortner:
But it's still a single wire call; I don't think those are significantly slower, though I guess I don't remember testing it.

p0deje:
Pretty sure it's not that slower, but why retrieve everything when there is no need to?

titusfortner:
Because I think we need to grab all of them in order to do filters properly. There are other filters we can use if we grab all the elements and then run `#select` on them. `:visible` is definitely the most important one, though.

p0deje:
Sometimes we need to grab all of them, but sometimes not. Right now, `:visible` is the only selector that requires fetching all elements, but unless it's passed we can stick to old code.

titusfortner:
But the code is so much more simple without that conditional. :-D

p0deje:
You added two `unless blah.nil?` conditions, so it's not that much more simple ¯_(ツ)_/¯

titusfortner:
Yes, I took out one conditional and put in another. You want to add the one I took out back in as well. (also removing it removes the need for that extra parameter and `#__send__`.

p0deje:
Ok, it's up to you to decide which one is better, I just see no point in doing `#find_elements` over `#find_element` when there is no need to.

titusfortner:
I'm not strongly attached to it. Removing an extra conditional seems preferable to doing the precisely correct Selenium call, especially if there is no performance difference. Is there any other reason for it other than correctness? (again, I'm not opposed to correctness just trying to make sure I'm thinking through both sides).

p0deje:
I think the performance will degrade in cases when Watir is used as spider tool. For example, the user will retrieve all the links on the page, iterate through each of them and retrieve hrefs. Currently, only first call to retrieve all the links will do `#find_elements`, all others will call `#find_element`. This PR will do `#find_elements` all the time which might be slower on a huge links collection. Of course, I haven't done any benchmarks.

titusfortner:
Yes, that's true, but it only does a `#select` if `:visible` is defined
For `:index` it is just `elements[index]`
And it is a direct replacement `find_elements(...)[0]` for `find_element(...)`, so it shouldn't be any performance loss.

p0deje:
Let's do simple benchmark:

``` ruby
require 'benchmark'
require 'selenium-webdriver'

begin
  driver = Selenium::WebDriver.for(:chrome)
  driver.get "data:text/html,#{'<div />' * 1000}"

  Benchmark.bm do |x|
    x.report('#find_element ') { 10.times { driver.find_element(tag_name: 'div')     } }
    x.report('#find_elements') { 10.times { driver.find_elements(tag_name: 'div')[0] } }
  end
ensure
  driver.quit
end
```

```
$ bundle exec ruby test.rb
       user     system      total        real
#find_element   0.000000   0.000000   0.000000 (  0.083634)
#find_elements  0.020000   0.000000   0.020000 (  1.832601)
```

titusfortner:
And that answers that. Good call, I'll put the conditional back in. :-D

titusfortner:
@p0deje - This should fix all the things, including the `element_locator_spec.rb` tests

p0deje:
👍

Issue #442, opened 2016-08-17
Validating Elements found #by_id


titusfortner:
We don't validate that an element we locate by id is of the type we specify.

All of these return true for a specific:

```
browser.div(id: 'foo').exists?
browser.span(id: 'foo').exists?
browser.iframe(id: 'foo').exists?
browser.text_field(id: 'foo').exists?
```

Is there a reason we decided to short circuit the validation for id's?

p0deje:
Guess it's just for performance reasons. We can probably change it so that it validates all the time.

titusfortner:
Oh, this was a bug and #467 fixed this, good to close

PR #467, opened 2016-09-13
id should still be validated with the tag name

titusfortner:
This is for issue #442 

I'm not sure why we've allowed id to match any tag name. It seems like if it is specified it should be required to match.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -38,11 +38,11 @@ module Watir
         def locate
           e = by_id and return e # short-circuit if :id is given
 
-          if @selector.size == 1
-            element = find_first_by_one
-          else
-            element = find_first_by_multiple
-          end
+          element = if @selector.size == 1
+                      find_first_by_one
+                    else
+                      find_first_by_multiple
+                    end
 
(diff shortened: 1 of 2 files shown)


p0deje:
👍

titusfortner:
Oh, wow, this actually causes tests to fail. For realsies.
`non_control_elements.html` has 4 different tags with `id="lead"`
I'm not sure how we want to deal with this.
1. If the first matching id value doesn't match the tag, fail it
2. Assume that you may need to test a website with more than one of the same id on it and treat it like any other multiple selector

titusfortner:
Regardless, the wrong elements were getting matched in our specs. :)

titusfortner:
Ok, so it turns our that in `#by_id` we deleted :tag_name to verify no extra parameters, but didn't put it back in when sending it to the element_validation class. This should fix it.

titusfortner:
This also is needed for this PR - https://github.com/watir/watirspec/pull/92

p0deje:
I think it was based on the assumption that IDs are unique, so we don't care about anything else (tag_name).

titusfortner:
merged - 99b5d467de5022d6985a6a494b408fa561da0f2c

PR #469, opened 2016-09-19
implement browser configuration for Watir

titusfortner:
It's odd that we need to require users to do things like this to customize Watir:

``` ruby
http_client = Selenium::WebDriver::Remote::Http::Default.new
http_client.timeout = 120
caps = Selenium::WebDriver::Remote::Capabilities.chrome

Watir::Browser.new :remote, remote_capabilities: caps, http_client: http_client
```

This PR allows:

``` ruby
Watir::Browser.new :remote, browser: chrome, client_timeout: 120
```

After rewriting all the options classes in Selenium, I thought something like this would be nice.
It supports a number of top-level arguments, and then anything else that is put into the opt will get passed along into the Capabilities class.

There is more tweaking that could be done here, but I wanted to get feedback on what I have.
Current master branch on Selenium is required for this code to run.

### DIFF
--- a/lib/watir/browser.rb
+++ b/lib/watir/browser.rb
@@ -43,7 +43,8 @@ module Watir
     def initialize(browser = :chrome, *args)
       case browser
       when ::Symbol, String
-        @driver = Selenium::WebDriver.for browser.to_sym, *args
+        opts = process_capabilities(browser.to_sym, *args)
+        @driver = Selenium::WebDriver.for browser.to_sym, opts
       when Selenium::WebDriver::Driver
         @driver = browser
       else
@@ -51,7 +52,7 @@ module Watir
       end
 
       @after_hooks = AfterHooks.new(self)
-      @current_frame  = nil
+      @current_frame = nil
       @closed = false
     end
 
@@ -331,7 +332,7 @@ module Watir
       when Array
         obj.map { |e| wrap_elements_in(e) }
       when Hash
-        obj.each { |k,v| obj[k] = wrap_elements_in(v) }
+        obj.each { |k, v| obj[k] = wrap_elements_in(v) }
 
         obj
       else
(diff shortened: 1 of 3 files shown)


p0deje:
Let me give you an example of current driver initialization we have in Toptal:

```ruby
proxy_url = [Billy.proxy.host, Billy.proxy.port].join(':')
proxy = Selenium::WebDriver::Proxy.new(http: proxy_url, ssl: proxy_url)

driver_options = {
  args: %w[--disable-translate],
  desired_capabilities: Selenium::WebDriver::Remote::Capabilities.chrome(
    'chromeOptions' => {
      'localState' => {
        'protocol_handler' => {
          'excluded_schemes' => {
            'skype' => false # disable external protocol popups for `skype:' links
          }
        }
      }
    }
  ),
  detach: false,
  http_client: Features::Helpers::RetryingHttpClient.new,
  listener: event_listener,
  prefs: {
    download: {
      prompt_for_download: false,
      default_directory: '~/Downloads/'
    }
  },
  proxy: proxy,
  service_log_path: "log/chromedriver#{ENV['TEST_ENV_NUMBER']}.log"
}

driver = Selenium::WebDriver.for(:chrome, driver_options)
browser = ::Watir::Browser.new(driver)
```

It's pretty huge and requires deep knowledge of driver configuration. I see that we could have simplified this into something like:

```ruby
options = {
  # these are options for Selenium gem
  selenium: {
    proxy: [Billy.proxy.host, Billy.proxy.port].join(':'),
    http_client: Features::Helpers::RetryingHttpClient.new,
    listener: event_listener,
  }
  # these are options for ChromeDriver
  driver: {
    args: %w[--disable-translate],
    detach: false,
    log: "log/chromedriver#{ENV['TEST_ENV_NUMBER']}.log"
    prefs: {
      download: {
        prompt_for_download: false,
        default_directory: '~/Downloads/'
      }
    },
  },
  # these are options for Chrome
  browser: {
    'chromeOptions' => {
      'localState' => {
        'protocol_handler' => {
          'excluded_schemes' => {
            'skype' => false # disable external protocol popups for `skype:' links
          }
        }
      }
    }
  }
}
browser = Watir::Browser.new(:chrome, options)
```

This is still not good enough IMO because it requires searching for Chrome/ChromeDriver options.

Some things can be potentially abstracted. For example, downloading of files:

```ruby
options = Watir::Browser::Options.new
# internally, the following code provides compatibility layer to automatically
# download files across different drivers/browsers
options.download_files = true
options.download_path = '~/Downloads/'
```

That might be actually useful, but I guess there are so many options people are using that might become maintenance nightmare and people will still directly use Selenium configuration.

Maybe we should keep things they are right now and not try to re-implement what Selenium already has.

I don't really have strong preference here, so I guess it's up to you @titusfortner 😄

titusfortner:
@p0deje - fyi default in Selenium for detach is now false

Also, I'm curious about `RetryingHttpClient` and how that was affected by the recent timeout setting changes. Can you share the code you're using? I'm assuming it is a subclass of `Net::HTTP::Persistent`?

I want this functionality included somewhere in the overall project. I'd actually like to create `ChromeOptions` and `FirefoxOptions` classes in Selenium like the other bindings have, especially since things get confusing with some things being strings and some things being more Ruby with symbols, and mixing them is a pain. That being said, even your example above would make it tough when some strings are camel case and some are underscored.

I'll take another look at this implementation. Maybe we want to wait on this to get that figured out. Or maybe I'll add this a Watir Configuration gem instead.

p0deje:
> fyi default in Selenium for detach is now false

Thanks, I missed that.

> Also, I'm curious about RetryingHttpClient and how that was affected by the recent timeout setting changes. Can you share the code you're using? I'm assuming it is a subclass of Net::HTTP::Persistent?

Does ChromeDriver support persistent HTTP connections? The code is a very simple class making sure we retry upon `Net::OpenTimeout`:

```ruby
# Custom selenium-webdriver HTTP client which wraps default implementation
# by rescuing `Net::OpenTimeout` and retrying the requests.
# This error is happening from time to time on CI and we could not find
# a reason for it to happen.
class RetryingHttpClient < Selenium::WebDriver::Remote::Http::Default
  private

  def request(_verb, _url, _headers, _payload, _redirects = 0)
    retries ||= 3
    super
  rescue Net::OpenTimeout
    raise if (retries -= 1).zero?
    retry
  end
end
```

> I want this functionality included somewhere in the overall project. I'd actually like to create ChromeOptions and FirefoxOptions classes in Selenium like the other bindings have, especially since things get confusing with some things being strings and some things being more Ruby with symbols, 
and mixing them is a pain. That being said, even your example above would make it tough when some strings are camel case and some are underscored.

I like the idea of custom `*Options` classes. It can behave like `HashWithIndifferentAccess` so that both symbols and strings can be used as keys. I don't think it should care about camelCase vs snake_case, we can leave it for the user.

lib/watir/browser.rb
```diff
@@ -343,5 +344,81 @@ def wrap_element(element)
...
+        raise Watir::Exception::Error, "browser type #{browser} is not recognized"
+      end
+    end
+
+    def set_path (browser, path)
+      case browser
+      when :firefox, :ff
+        Selenium::WebDriver::Firefox::Binary.path = path
+      when :chrome
+        Selenium::WebDriver::Firefox::Chrome.path = path
```
brianpurgert2, line 415: should this be Selenium::WebDriver::Chrome.path = path

titusfortner: Actually this whole approach needs to be updated since Selenium bindings have been reworked since I made this PR. Binary path should be passed in as a capability. But yes, that's definitely a typo. :)

azohra:
We have a silly approach that makes sense for users driving primarily with chrome -- I cannot imagine how long it would take to flesh this out across browsers with all available flags. 

```ruby
Local.new(headless: true, profile: '/tmp/user1', size: '1900x1200', ignore_ssl_errors: true)

module Ferris
  module Session
    class Local
      attr_reader :browser, :switches, :prefs

      SWITCH_MAP = { headless:          '--headless',
                     cpu_only:          '--disable-gpu',
                     profile:           'user-data-dir=****',
                     size:              '--window-size=****',
                     user_agent:        '--user-agent=****',
                     ignore_ssl_errors: '--ignore-certificate-errors' }.freeze

      def initialize(**keyword_args)
        vendor = keyword_args.fetch(:browser, :chrome)
        @switches = []
        keyword_args.each do |k, v|
          add_switch(k, v) if SWITCH_MAP.include?(k) && v  # dont write false switches
        end
        @browser = Watir::Browser.new(vendor, switches: @switches, prefs: @prefs)
      end

      def add_switch(k, v)
        @switches.push SWITCH_MAP[k].gsub('****', v.to_s)
      end

    end
  end
end```

titusfortner:
I'm amused that people are commenting on this today. I prototyped what I think should be the user API over the weekend. A few pages on my fork's wiki: https://github.com/titusfortner/watir/wiki

I'm closing this PR since it would need an overhaul regardless. I'll add an issue for tracking purposes.

PR #470, opened 2016-09-22
Merge WatirSpec into Watir

p0deje:
That is just an initial version and I'd like to do the followup refactoring after merge.
### TODO
- [x] Fix failing specs
- [x] Pull latest watirspecs


titusfortner:
1. What do you think about starting with #478 to preserve history?
2. When running specs I was getting an error with PersistentBrowserHelper not defined

p0deje:
> What do you think about starting with #478 to preserve history?

Makes sense, but let me do that in this PR.

> When running specs I was getting an error with PersistentBrowserHelper not defined

Yeah, my bad, not tested follow up refactoring.

Generally, how does this look? I'm interested in review of `lib/watirspec`.

titusfortner:
Yes, I agree, do it in this PR, but the stuff is there if you need it.
I haven't had a chance to dig into it beyond briefly trying to figure out the error, then going back to other things. :)

p0deje:
I've pulled WatirSpec with history as you proposed and fixed specs, so this should now be mergeable. Please, take a look at my latter commits as they are the ones that should be reviewed.

titusfortner:
This code isn't my wheelhouse, but I can run local & remote tests, so it's good to go as far as I'm concerned.

Why did we used to make a distinction with persistent browser? Just curious because I agree that we shouldn't need it for what we're doing now. The only reason I can see to have it would be if we were to use something like Saucelabs, but that can be a special implementation later if we decide that makes sense to go that route. :)

titusfortner (approved):

titusfortner:
fwiw - it's look like Capybara does what I suggested: https://github.com/jnicklas/capybara/#setup
Require > 1, and tell users to set to < 2 if they want to use an older version of Ruby.

p0deje:
Nevermind, I've already moved to TCPServer.

titusfortner (approved):
This looks great, let's get this merged in!

lib/watirspec/server.rb
```diff
@@ -54,14 +54,14 @@ def run!
...
-            headers = headers.merge('Server' => 'ruby')
+            headers['Server'] = 'ruby'
             headers = headers.to_a.map { |array| array.join(': ') }
 
             response = ["HTTP/1.1 #{status}"] + headers
             response = response.join("\r\n")
 
             client.puts(response)
-            client.puts("\r\n")
+            client.puts("\r\n\r\n")
```
titusfortner: This change causes Chrome (only one I tested) to often not properly load the page. Reverting this fixes it.

lib/watir/browser.rb
```diff
@@ -72,7 +72,7 @@ def inspect
     #
 
     def goto(uri)
-      uri = "http://#{uri}" unless uri =~ URI.regexp
+      uri = "http://#{uri}" unless uri =~ URI.regexp(%w(http https))
```
titusfortner: This gives errors when trying to use:
`about:blank` and `data:text/html;content-type=utf-8,foobar`

titusfortner:
What do you think about stopping testing with phantomjs? I'm much more concerned that Firefox, Chrome, Safari, IE & Edge can be tested well.

Even putting this code back in for phantomjs: `client.puts("\r\n")`, there are still `Net::ReadTimeout` errors. Otherwise, we might need to just put the rack server back in.

lib/watirspec/server.rb
```diff
@@ -0,0 +1,137 @@
...
+      def run_in_child_process
+        begin
+          require "childprocess"
+        rescue LoadError => ex
+          raise "please run `gem install childprocess` for WatirSpec on Windows + MRI\n\t(caught: #{ex.message})"
+        end
+
+        path = File.expand_path("../../spec_helper.rb", __FILE__)
+
+        process = ChildProcess.build(WatirSpec.ruby, path)
```
titusfortner: The`WatirSpec#ruby` method was removed, and this path no longer exists, so this code won't run on Windows as is.

lib/watir/browser.rb
p0deje: Removed

lib/watirspec/server.rb
p0deje: WAT. I'll take a look but that's actually the correct way ([see spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html)) - there should be one CRLF at the end of header block and one CRLF before body.

lib/watir/browser.rb
p0deje: I have just replaced localhost with 0.0.0.0, just like it was before

lib/watirspec/server.rb
p0deje: Yeah, thanks for catching this. I haven't tested on Windows yet.

lib/watirspec/server.rb
p0deje: I see that Travis fails, but i believe that is because it uses Chrome 55 and ChromeDriver 2.24 which are not compatible.

p0deje: Ah, completely forgot that `puts` appends `\n`. Should be correct now.

p0deje:
> What do you think about stopping testing with phantomjs? I'm much more concerned that Firefox, Chrome, Safari, IE & Edge can be tested well.

We can drop it if you want, but let's keep it in this PR.

> Even putting this code back in for phantomjs: client.puts("\r\n"), there are still Net::ReadTimeout errors. Otherwise, we might need to just put the rack server back in.

This was my mistake and now is fixed.

p0deje:
Looks like the Travis is green now (at least for PR). If you're okay, I'd like to merge it as is and then try to set up CI for Windows using Appveyor. If there will be some specs failing, we'll fix them in followup PRs.

p0deje:
Chrome failure seems to be more related to Chrome 55 rather than WatirSpec itself...

titusfortner:
Sorry for the delay in looking at this. I can't get it to work on my Mac with any browser. I'm getting: "File not found: /forms_with_input_elements.html"

p0deje:
That's strange, I've just did the following and everything went well:

```
$ git clone watir/watir
$ git checkout watirspec
$ bundle install
$ bundle exec rake spec
```

Specs passed without any issues. Can this be some local environment problem?

titusfortner:
This looks good in Linux & Mac, but I'm having an issue on Windows:
https://gist.github.com/titusfortner/92415fb4d642775b8a3673b475dfadd2

p0deje:
Are old watirspecs work fine on Windows?

titusfortner:
It hasn't worked with the new setup; I think this is still an issue: https://github.com/watir/watir/pull/470#pullrequestreview-2462573

I tried playing around with parameters to see if I could get it to work, but I'm mostly just flailing about with it.

titusfortner:
Also, I don't think the failures are chrome version related; I think they are minor race condition issues. This commit gets things consistently passing for me: https://github.com/watir/watir/commit/be7cdcd8dcf3325d135d764f9fe8f80ee5749551

p0deje:
Latest commit fixes an issues on Windows. Now, server is run in a thread (like on JRuby), not in a childprocess. I don't know why it used childprocess before (my best guess is that there were some issues with threads on Windows in old Ruby versions). Now the server is run correctly and specs work fine. I've caught 4 failures for Chrome on Windows 7 (after_hooks, image specs), but I don't think they are relevant for this PR.

If Travis is green and I have your 👍, I'm going to merge this, make proper updates to watirspec repository and move on to setting up AppVeyor CI for automated testing on Windows.

titusfortner:
Images don't seem to be loading on Windows for: `http://127.0.0.1:8181/images.html`

titusfortner:
Also window switching spec is giving:

```
Errno::EPIPE: Broken pipe
./lib/watirspec/server.rb:83:in `write'
```

It appears to be a race condition because putting some pauses in there seems to fix it.

p0deje:
> Images don't seem to be loading on Windows for: http://127.0.0.1:8181/images.html

Thanks for catching, this is now fixed.

p0deje:
> Also window switching spec is giving:
> 
> Errno::EPIPE: Broken pipe 
> ./lib/watirspec/server.rb:83:in `write'
> 
> It appears to be a race condition because putting some pauses in there seems to fix it.

I can't reproduce it, but I joined multiple `#write` into a single one, so can you give it another try?

titusfortner:
Well, either I can't reproduce it any more, or that fixed it. :) Some other Firefox weirdness on Windows that I don't think is related to this. I'm going to call this one good, though I'll probably push the stuff in watirspec_updates as well since there are a couple race conditions it addresses.

titusfortner (approved):

p0deje:
Then let's open a bottle to celebrate!

PR #479, opened 2016-09-26
Wait in the Watir

titusfortner:
After my previous PRs and several discussions, this is my overall proposal. I added [a bunch of tests](https://github.com/watir/watirspec/pull/87)
1. This waits for a set of preconditions (as specified by each method) to be true before failing when taking an action on an element
2. There is a single toggle to make (almost) everything backward compatible
   1. The only non-backward compatible concern is tests that take actions on elements, rescue the `TimeoutError` and use that for flow control
   2. I've added a warning message for people to indicate that might be what is going on
3. There are performance short cuts for everything now, very few extra calls to get this to work
4. It creates and clears a Wait::Timer instance for every `#element_call` to avoid repeating the same timeout value (@jkotests' concern)
5. It ensures that wait actions are always taken at least once (acting like asserts instead of waits), regardless of the time remaining on the overall timer.
6. It deprecates when_\* decorators. 
   1. Because this is a huge change, I'm not removing them, just putting warnings everywhere
   2. There should be minimal need for `when_present` or `when_enabled` with the new code
   3. Any need that does exist can be replaced with the new `wait_until` methods
   4. This gets rid of the unnecessary metaprogramming
   5. Having a Delegator that is defined independently of the future meta-programmed action isn't necessary
   6. Accepting a block that only executes code after taking the action in the method is not good form
7. It improves error messages further 
   1. It gives status on whether element is located but not visible or enabled, etc
   2. It only returns the `selector_string` of the parent if it is the parent method that is actually failing
8. Deprecates ordered parameters of `timeout` and `message` in favor of keywords 
   1. This is nice if you want to pass in a message, but not the timeout value
   2. It also provides context for the number that is getting passed in `timeout: 5`, etc
9. Implement generic `wait_until` and `wait_while` in the manner @p0deje suggested
   1. These return `self` so that actions can be chained after
   2. These do lose some of the extra error messaging, but several people pointed out to me that the stack trace will make the call exceedingly clear what the call was
10. There was some discussion of the `#wait_until` vs `#wait_while` vs `#wait_until_not` options... I'm leaning against including a `#wait_until_not` method. If `#wait_while` seems clunky, it seems like it would be easier to put the negation in the block `wait_until { !condition }` 

**Examples of new wait calls**
Element#wait_until(&:present?).text
Element#wait_until(timeout: 5, &:present?).text
Element#wait_until(message: “Sorry”, &:present?)
Element#wait_until { |el| el.text == ‘foo’ }
Element#wait_until { element.text == ‘foo’ }

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -583,16 +583,9 @@ module Watir
       locator.locate
     end
 
-    protected
-
     def selector_string
       return @selector.inspect if @query_scope.is_a?(Browser)
-      query_scope = if @query_scope.is_a?(IFrame)
-                      @query_scope.element.instance_variable_get("@query_scope")
-                    else
-                      @query_scope
-                    end
-      "#{query_scope.selector_string} --> #{@selector.inspect}"
+      "#{@query_scope.selector_string} --> #{@selector.inspect}"
     end
 
     private
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -110,9 +110,7 @@ module Watir
     #
 
     def click(*modifiers)
-      element_call(:wait_for_present) do
-        assert_enabled
-
+      element_call(:wait_for_enabled) do
         if modifiers.any?
           assert_has_input_devices_for "click(#{modifiers.join ', '})"
 
@@ -304,8 +302,7 @@ module Watir
     #
 
     def send_keys(*args)
-      element_call(:wait_for_present) do
-        assert_writable
+      element_call(:wait_for_writable) do
         @element.send_keys(*args)
       end
     end
@@ -496,47 +493,59 @@ module Watir
 
     protected
 
-    def wait_for_exists(timeout = nil)
+    def wait_for_exists
       return assert_exists unless Watir.relaxed_locate?
       return if exists? # Performance shortcut
 
-      timeout ||= Watir.default_timeout
-      timeout_off = timeout <= 0
-      end_time = Time.now + timeout
-
       begin
-        remaining_time = timeout_off ? 0 : end_time - Time.now
-        Watir::Wait.until(remaining_time) { exists? }
+        @query_scope.wait_for_exists
+        Watir::Wait.until { exists? }
       rescue Watir::Wait::TimeoutError
-        unless timeout_off
+        unless Watir.default_timeout == 0
           warn "This test has slept for the duration of the default timeout. "\
                   "If your test is passing, consider using Element#exists? instead of rescuing this error)"
         end
-        raise unknown_exception, "unable to locate element, using #{selector_string} "\
-                                         "after waiting #{timeout} seconds)"
+        raise unknown_exception, "timed out after #{Watir.default_timeout} seconds "\
+                                         "waiting for #{selector_string} to be located"
       end
     end
 
-    def wait_for_present(timeout = nil)
+    def wait_for_present
       return assert_exists unless Watir.relaxed_locate?
-      return if present? # Performance shortcut
 
-      timeout ||= Watir.default_timeout
-      timeout_off = timeout <= 0
-      end_time = Time.now + timeout
-
+      wait_for_exists
       begin
-        remaining_time = timeout_off ? 0 : end_time - Time.now
-        wait_for_exists(remaining_time)
-        remaining_time = timeout_off ? 0 : end_time - Time.now
-        Watir::Wait.until(remaining_time) { present? }
-      rescue Watir::Wait::TimeoutError
-        unless timeout_off
+        wait_until_present
+      rescue Watir::Wait::TimeoutError => ex
+        unless Watir.default_timeout == 0
           warn "This test has slept for the duration of the default timeout. "\
                   "If your test is passing, consider using Element#present? instead of rescuing this error)"
         end
-        raise unknown_exception, "element located but not visible, using #{selector_string} "\
-                                         "after waiting #{timeout} seconds)"
+        raise unknown_exception, "element located, but timed out after #{Watir.default_timeout} seconds #{ex.msg}"
+      end
+    end
+
+    def wait_for_enabled
+      return assert_enabled if Watir.relaxed_locate?
+
+      wait_for_present
+      begin
+        Watir::Wait.until { @element.enabled? }
+      rescue Watir::Wait::TimeoutError
+        message = "element present, but timed out after #{Watir.default_timeout} seconds waiting for #{selector_string} to be enabled"
+        raise ObjectDisabledException, message
+      end
+    end
+
+    def wait_for_writable
+      return assert_writable unless Watir.relaxed_locate?
+
+      wait_for_enabled
+      begin
+        Watir::Wait.until { !respond_to?(:readonly?) || !readonly? }
+      rescue Watir::Wait::TimeoutError
+        message = "element present and enabled, but timed out after #{Watir.default_timeout} seconds waiting for #{selector_string} to not be readonly"
+        raise ObjectReadOnlyException, message
       end
     end
 
@@ -646,12 +655,14 @@ module Watir
     end
 
     def element_call(exist_check = :wait_for_exists)
+      already_locked = Wait.timer.locked?
+      Wait.timer = Wait::Timer.new(::Time.now + Watir.default_timeout) unless already_locked
       send exist_check
       yield
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
-      @element = locate
-      assert_element_found
       retry
+    ensure
+      Wait.timer.reset! unless already_locked
     end
 
     def method_missing(meth, *args, &blk)
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -498,11 +498,13 @@ module Watir
 
       begin
         @query_scope.wait_for_exists
-        Watir::Wait.until(element: self, &:exists?)
+        wait_until(&:exists?)
       rescue Watir::Wait::TimeoutError
         unless Watir.default_timeout == 0
-          warn "This test has slept for the duration of the default timeout. "\
-                  "If your test is passing, consider using Element#exists? instead of rescuing this error"
+          message = "This code has slept for the duration of the default timeout "
+          message << "waiting for an Element to exist. If the test is still passing, "
+          message << "consider using Element#exists? instead of rescuing UnknownObjectException"
+          warn message
         end
         raise unknown_exception, "timed out after #{Watir.default_timeout} seconds, "\
                                          "waiting for #{selector_string} to be located"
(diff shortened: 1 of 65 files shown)


lib/watir/wait.rb
```diff
@@ -71,20 +79,16 @@ def message_for(timeout, message)
       end
 
       def run_with_timer(timeout, &block)
-        if timeout == 0
```
p0deje, line 74: You should return that cause it will be failing when `timeout = 0` is passed as timeout is reached before block is executed.

titusfortner:
Note - this also implements #458

titusfortner:
I want to make sure we're good on each of these points.

1- Are these the right defaults for each of these?
(the default is to `wait_for_exists`)

`wait_for_enabled`
Element#click

`wait_for_writable`
Element#send_keys
Element#set (UserEditable)
Element#clear (UserEditable)

`wait_for_present`
Form#submit
Element#double_click
Element#right_click
Element#hover
Element#drag_and_drop_on
Element#drag_and_drop_by

2- Are we ok with deprecating ordered pairs in lieu of keywords for message & timeout on `#wait_until` and `#wait_while`?

3- Are we good on deprecating when_present, delegators, method_missing, etc (to be actually removed on future release)?

4- Are we good with the implementation of #458 which results in removing the more specific error messages with updated wait_until & wait_while? As was pointed out to me, which element called the method can be found very easily in the stack trace.

5- What are the specific implementation concerns with the global toggle?

titusfortner:
Also, I blogged about why I want to see this PR - http://watirtight.com/2016/10/13/wait-in-the-watir.html

lib/watir/elements/element.rb
```diff
@@ -395,13 +376,13 @@ def wd
...
     # Returns true if this element is visible on the page.
+    # Raises exception if element does not exist
     #
     # @return [Boolean]
     #
 
     def visible?
-      assert_exists
-      element_call { @element.displayed? }
+      element_call(:assert_exists) { @element.displayed? }
```
p0deje: Why doesn't it wait for element to exist?

lib/watir/elements/element.rb
```diff
@@ -557,6 +596,10 @@ def selector_string
 
     private
 
+    def unknown_exception
```
p0deje, line 593: I'm not sure it's a good idea to make exception polymorphic. Is there really any need to have it a separate method?

lib/watir/elements/element.rb
```diff
@@ -614,12 +657,14 @@ def assert_is_element(obj)
...
+    def element_call(exist_check = :wait_for_exists)
+      Wait.timer ||= Wait::Timer.new.tap { |t| t.end_time = ::Time.now + Watir.default_timeout }
+      send exist_check
       yield
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
-      @element = locate
-      assert_element_found
       retry
+    ensure
+      Wait.timer = nil
```
p0deje: We should not reset timer. The very idea of `Wait.timer` is to allow user to implement his own custom timer class which is set in the beginning of execution. For example, here is [watir-timecop](https://github.com/p0deje/watir-timecop/blob/master/lib/watir-timecop.rb).

spec/element_spec.rb
```diff
@@ -1,135 +1,140 @@
 require 'watirspec_helper'
 
-describe Watir::Element do
+module Watir
```
p0deje: Why do that?

spec/watirspec/wait_spec.rb
```diff
@@ -1,278 +1,305 @@
 require "watirspec_helper"
 
 not_compliant_on :safari do
-  describe Watir::Wait do
-    describe "#until" do
-      it "returns result of block if truthy" do
-        result = 'catter'
-        expect(Watir::Wait.until(0.5) { result }).to be result
-      end
+  module Watir
```
p0deje: Why introduce module?

lib/watir/elements/element.rb
```diff
@@ -614,12 +657,14 @@ def assert_is_element(obj)
       end
     end
 
-    def element_call
+    def element_call(exist_check = :wait_for_exists)
+      Wait.timer ||= Wait::Timer.new.tap { |t| t.end_time = ::Time.now + Watir.default_timeout }
```
p0deje: I don't follow why you need to change end_time here. Isn't it already current time + default timeout?

p0deje: I don't follow why you need to change end_time here. Isn't it already current time + default timeout?

p0deje (dismissed):

lib/watir/elements/element.rb
titusfortner: None of the predicate methods wait by default. Only taking actions on the element uses an automatic wait.

p0deje: I understand, but do you think that's the correct behavior? I mean that I would assume it waits for exists before checking visibility.

lib/watir/wait.rb
titusfortner: I added this logic to the `Timer#wait` method. The code for a given timer will always execute once, regardless of whether the timeout is set to `0` or `0.01`.

lib/watir/elements/element.rb
titusfortner: Hmmm... Because `visible?` throws an exception if it doesn't exist, and `present?` returns false if it doesn't exist... Interesting idea. A few thoughts in no particular order
1. Most people use these methods interchangeably right now
2. Clarifying the meaning of each might even be harder than before. Is `element.visible?` is now equivalent to `element.wait_until(&:exists?).present?` easy enough to understand?
3. Is this a pattern that is likely to be used?
4. I don't like having two methods that are mostly used for the same purpose like we have now

I actually think I like this, but I'm wondering if this shouldn't be made its own thing with its own blog post, etc. At the very least I'd like to get more opinions on it.

lib/watir/elements/element.rb
titusfortner: Yes, because if we don't do this then these fail: https://github.com/titusfortner/watir/blob/cfecfb/spec/watirspec/elements/frame_spec.rb#L74-L84 by getting `UnknownObjectException` instead of `UnknownFrameException`

spec/element_spec.rb
titusfortner: In classes we should prefer:

```
module Foo
  class Bar
...
```

to 

```
class Foo::Bar
...
```

This change was along those lines, and also means we don't have to specify the Watir namespace everywhere in the specs.

That being said, I didn't make sure this was consistent in other files, so I should probably remove it for now.

lib/watir/elements/element.rb
titusfortner: The logic is that the `#wait_for_present` & `#wait_for_exist` methods are only ever accessed by the `#element_call` method. We want to make sure that in the process of verifying parent exists, then parent present, then child exists, then child present, etc, that we do not compound the timeouts. So we create a `Timer` instance, set it as the class variable in Wait and that is what gets used for all calls until the timer expires. We use `||=` such that it only will ever override a `nil` value in case `#element_call` is nested somewhere.

titusfortner: Yes, but it needs to be set once before the multiple waits.

titusfortner: The `||=` didn't make sense because the `#timer` method always returns a value. I committed a fix.

lib/watir/elements/element.rb
titusfortner: Yes, you were right, this didn't do what I thought it was doing. Committed a fix.

spec/element_spec.rb
p0deje: We can do that separately, but I think it would be better to remove it for current PR. The changeset is already huge and it creates unnecessary noise.

titusfortner:
@p0deje - Everything is passing now; let me know what else needs to be discussed.

p0deje:
Let's rebase, get rid of `fixup!` commit and merge.

I haven't found a way to make this simpler, so let's just push this forward and then reiterate.

azohra:
Wooooooo! Nice work guys.

azohra:
Hey @titusfortner 

URL: `https://www.loblaws.ca`
Browser: `chrome`
Call: `@browser.element(class: 'ui-mobile-rendering').wait_while{|el| el.exists? }`
Return: 

``` ruby
Watir::Wait::TimeoutError: timed out after 60 seconds, waiting for false condition on {:class=>"ui-mobile-rendering"}
from /Users/fbarann/.gem/ruby/2.3.1/gems/watir-6.0.1/lib/watir/wait.rb:72:in `while'
```

In the example above, we have a `class` which only exists during page load. We wait for it to no longer be in the DOM. 

The initial call is timing out and not catching changes in the DOM. If we send another call to `@browser.element(class: 'ui-mobile-rendering').wait_while{|el| el.exists? }` after the page has loaded, it immediately returns as expected 

More interestingly, If we use `Watir::Wait.while { @browser.element(class: 'ui-mobile-rendering').exists? }` then things work as expected first time around. 

I imagine the intention of this PR is to have the element reference constantly re-evaluated during wait operations, thoughts?

p.s: we have eliminated `page-object` from the conversation to ensure we are dealing with Watir as you intended.

titusfortner:
@tk8817 
It looks like it is the html tag itself that has this temporary class value.
This is an interesting side effect of Watir moving away from "always locate" (which happened a year and a half ago).
Essentially the code locates the element that matches the selector, but while the selector goes away, the underlying element is still there. 
You don't want to wait for the element  not to exist, but for the attribute not to exist:
`@browser.element.wait_while{|el| el.attribute_value('class').include? 'ui-mobile-rendering'? }`

azohra:
@titusfortner - very interesting, thank you for the explanation. To really drive it home, we will shift our mindset to define elements with permanent locators and use temporary attributes in our waits/calls. 

I imagine this will crop up for a few folks who using the new element waiters that were getting lucky with non-element waiting methods up till now.

titusfortner:
@tk8817 - So, I'm starting to think this might should be treated like a bug, because it is not Principle of Least Surprise.

titusfortner:
@tk8817 - how does this show up in 6.0, but not in watir-webdriver 0.9.x? I don't think the wait_while should be different... If you want to discuss this more, find the Watir channel in the Selenium Slack: https://seleniumhq.herokuapp.com/

acappelgryn:
@titusfortner - I am currently using the last version of watir-webdriver released and looking to move my framework to watir 6.0.1. My current automation framework uses the `page_ready_element.when_present($timeout).some_action` format to wait for page elements. I have done some testing and found both `page_ready_element.wait_until(timeout: $timout, &:present?).some_action` and `page_ready_element.wait_until_present(timeout: $timout, &:present?).some_action` execute successfully. The `when_present` depreciation and passing arguments messages both do not show either. Which method, `wait_until` or `wait_until_present`, is best to use moving forward?

titusfortner:
You should be able to completely remove the `when_present` code and just do:
`page_ready_element.some_action`

The `when_present` is now being applied universally in Watir itself.

acappelgryn:
@titusfortner - That is excellent news! Thank you for the update sir.

PR #483, opened 2016-10-12
run after_hooks only if non-empty

colintsteele:
Previously, when the code was `run_checkers`, if an element had no error_checkers -or after_hooks- an empty array would be returned.  Now, instead, `nil` is returned.  This propagates to hangups in test_helper methods that provide wait functionality based on the `click` method

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -130,7 +130,7 @@ module Watir
         end
       end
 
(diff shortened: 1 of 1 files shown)


titusfortner:
This is already how the code works: https://github.com/watir/watir/blob/master/lib/watir/after_hooks.rb#L68-L72

PR #485, opened 2016-10-19
Initial support for AppVeyor CI

p0deje:
This PR adds AppVeyor as part of our CI which is a Windows machine.

Currently, the following tests are implemented:
- Chrome
- Firefox + geckodriver

Shortly we can add support for IE11 which is pre-installed in VM and PhantomJS (if we want to).

The tests are currently only run on Ruby 2.3. I don't see much point in testing different Ruby versions (like on Travis) mainly because in AppVeyor the builds are not run in parallel so the bigger matrix we have, the longer we'll have to wait.

Doctests are failing (seems to be running out of ephemeral ports) and IE11 needs some stabilization. In any case, I think it's good for a start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 Watir Powered By Selenium!
 
 [![Gem Version](https://badge.fury.io/rb/watir.svg)](http://badge.fury.io/rb/watir)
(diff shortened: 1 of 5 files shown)


p0deje:
@titusfortner Please take a look!

PR #494, opened 2016-11-10
have `#wait_while_present` always relocate

titusfortner:
@p0deje - you ok with this feature?

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -488,6 +488,10 @@ module Watir
       true
     end
 
+    def reset!
+      @element = nil
+    end
+
     protected
 
     def wait_for_exists
@@ -555,7 +559,7 @@ module Watir
     def assert_exists
       if @element && @selector.empty?
         ensure_context
-        @element = nil if stale?
+        @element.reset! if stale?
       elsif @element && !stale?
         return
       else
--- a/lib/watir/wait.rb
+++ b/lib/watir/wait.rb
@@ -190,7 +190,10 @@ module Watir
         warn "Instead of passing arguments into #wait_while_present method, use keywords"
         timeout = deprecated_timeout
       end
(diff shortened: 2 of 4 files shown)


p0deje:
Why we need that?

titusfortner:
https://github.com/watir/watir/pull/479#issuecomment-259735670

titusfortner:
Another option... 

```
def not_present?
  self.reset!
  !present?
end

element.wait_until(&:not_present?)
```

Is there actually any other scenario in which `wait_while` makes sense for an element?
Bret commented before that he didn't see a compelling reason for a separate `wait_while` method, and thought that it made more sense to just have `wait_until` with people passing in their true conditionals even if that meant using `!=` instead of `==`, etc.

If we implement a `not_present?` maybe we can also get rid of `wait_while` as well?

azohra:
@titusfortner in terms of elegance, having an ability to avoid `!` is always nice - especially as it allows us to take advantage of `&:` notation. 

While `not_present` sounds like a good alternative (which preserves `&:` notation), I see a slippery slope where folks may want/expect negative decorators for all the popular methods (`not_exist`,`not_visible`,`not_enabled`).

``` ruby
loading_modal.wait_while(&:present?)
vs
loading_modal.wait_until {|e| !e.present?}
```

the former is just so wonderful to read. /my 2 cents

p0deje (dismissed):
Can we add a spec for this behavior so that we don't have regression if we try to refactor?

titusfortner:
@p0deje - added the spec... kind of unfortunate that it only applies to `#wait_while_present` instead of `#wait_while(&:present?)`... but it would break implementations if people use: `element.wait_while(&:stale?)`

p0deje (approved):

titusfortner:
Merged - 21fe9325913dfe1e3a4f81f2ea63fa6b78886398

Issue #403, opened 2016-02-07
generate default test suite


titusfortner:
Possible new feature:

As we're working on updating the websites, I was thinking it could be useful to have a default github template that people could fork, rename and use as a starting point for creating their test suites with Watir.

Then I got to thinking that maybe we could do it the same way jekyll does it: 
`jekyll new my_jekyll_site`
and do:
`watir new my_watir_testsuite`
Which requires the things and creates RSpec examples. Maybe this would make more sense if we collaborated on a watir-sanctioned page object model. Either way, I wanted to throw this out there as an idea.

titusfortner:
Revisiting this in a separate gem.

PR #496, opened 2016-11-18
generate default test suite - issue 403 - start

joesho112358:


titusfortner:
@joesho112358 Thanks for this PR! 

I don't think we want to add this functionality to watir directly at this time, but I'm working on a more fully featured approach to doing this via a separate gem. What I have so far is pretty rough, though. If you'd like to collaborate on that or provide input, please contact me titusfortner on gmail / twitter / pretty much wherever.

Thanks again.

joesho112358:
@titusfortner I am not sure having a separate gem just for setting up suites makes sense to me. Nevertheless, if you are going the separate gem route, then you should close the issue in this gem.

titusfortner:
Because Watir by itself isn't really enough to write a good test suite, so we need something to put together a more comprehensive default setup with page objects and session configuration and data modeling support, etc. Stuff beyond just Watir code.

joesho112358:
I get what you are going for: basically a mirror to rails, right? 

But just in the models alone, do you set up active record for database access and/or one of the many web service gems?
If database, then an independent suite is probably not supporting it's own database for an app and so the user would need to go in and configure the proper connection. If it is services, that is easy! So easy that you could just do a `require whatever` and off they go.

Either way, I don't see anyone getting a whole lot out of it, and that is just the models. I like the idea, but I think simpler is better for it.

titusfortner:
Errr, not really. Tests don't need (and ideally shouldn't use) databases. I'm referring to models as merely a place to store the data that is needed to interact with a page or an API through the course of a test.

I'm thinking more along the lines of creating a new project with boilerplate setup along the lines of `bundle gem your_project`

Watir by itself has no stake in how test data is organized or abstracted, and having intelligent defaults should not preclude configuration to support additional options. For instance, I would be very unlikely to encourage the use of gauge as a first choice for running tests. But if you wanted to use that, the solution provided should allow that to be configured if desired.

joesho112358:
Sure, so now you should close issue 403 because it has little to do with Watir.

Moreover, you should push up what you have to a different repository if you haven't already, and link to it in the read me if you really want to do this- that is all that needs done for issue 403. Correct?

titusfortner:
Oh, I forgot I added #403 so I wasn't following your references to it.
What I have is in a separate repo, but it is more proof-of-concept than usable still. Hopefully I'll be able to publicize it more soon.

PR #501, opened 2016-12-02
allow interval to be set in wait methods

titusfortner:
A feature Selenium has had that's been missing in Watir. 0.1 second pings for 30 seconds is a lot of pings. Would also be up for changing these defaults, or allowing the defaults to be configurable.

@p0deje what do you think?


p0deje:
Why not.

Issue #161, opened 2012-08-24
Disable bubbling of fire_event


moll:
Hey,

I noticed an older version of Watir had a bubble attribute to `fire_event`. Why was that removed?

I was just writing a test to test the `paste` event, which on the latest Chrome seems not to bubble. At least it's not caught by jQuery's delegates. But, when doing the same thing with Watir, that is setting the input via JavaScript to not trigger keyups and then `fire_event 'paste'` on that input, jQuery catches that — so it's a false positive.

Disabling bubbling in that case to make sure production code binds _on_ the input and not via delegates, would make writing this test possible. I mean, unless somehow using the fireEvent.js directly that's in Watir's repo.

titusfortner:
After 5+ years no obvious way for us to fix this within the current code; as well as not having an obvious use case for testing actual user interactions on a site, I'm going to close it as "won't fix"

PR #502, opened 2016-12-02
#select_text shouldn't be an extension and atom needs to get loaded

titusfortner:
Is there a reason this was in the extension folder? Doesn't seem necessary...


p0deje:
I don't know 😄 

What does this method do? Can you add specs for it?

titusfortner:
It highlights text inside an element. Like double clicking on a word. I'm honestly not sure why it would be needed, but @cheezy uses it in page-object gem, and I couldn't figure out why it needed to be in its own folder if we were going to support it.

The only way I could think to test it is to also implement `Browser#select_text`. I just grabbed code off of Stack Overflow so I'm not sure how robust it is, but it also seems like this would potentially be more useful, anyway.

This also brings up the question I have of how we do the atoms in the first place. Are we supposed to be periodically building them in Selenium and then updating them in our project? Do I need to figure out how to add this one to the Selenium project so it also can get built there? I couldn't even figure out how to build them in Selenium. I mean, there's a `build.desc` file for crazyfun, but there's no reference to it in Rakefile, and I'm not even sure how to use it since I don't even know what mechanism picks up that `desc` file in the rb/ directory. Do you know how this works, or is this something we should ask Jari?

p0deje:
> It highlights text inside an element. Like double clicking on a word. I'm honestly not sure why it would be needed, but @cheezy uses it in page-object gem, and I couldn't figure out why it needed to be in its own folder if we were going to support it.

Oh wow, I had no idea watir could do that. I don't know why it's needed too, so maybe @cheezy can explain it.

> This also brings up the question I have of how we do the atoms in the first place. Are we supposed to be periodically building them in Selenium and then updating them in our project? Do I need to figure out how to add this one to the Selenium project so it also can get built there? I couldn't even figure out how to build them in Selenium. I mean, there's a build.desc file for crazyfun, but there's no reference to it in Rakefile, and I'm not even sure how to use it since I don't even know what mechanism picks up that desc file in the rb/ directory. Do you know how this works, or is this something we should ask Jari?

I don't know how atoms are build neither in Selenium nor in Watir, that's why I didn't ever finished #161. I guess we need to figure it out somehow :)

titusfortner:
Status update: I'm working on updating/fixing the atoms from Selenium right now. I'll request a code review when everything is happy.

titusfortner:
This is getting handled with #603

PR #515, opened 2017-01-02
put #visible? and #enabled? back into element calls

titusfortner:
Nesting shouldo longer give issues with our Timer implementation

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -382,8 +382,7 @@ module Watir
     #
 
     def visible?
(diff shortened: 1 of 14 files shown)


p0deje (approved):
Reasonable, though having specs would also be great.

titusfortner:
@p0deje for the record, here's the spec: https://github.com/watir/watir/pull/515/commits/b65d82a9#diff-766c2ed149795115273e7f75480bb843R260

I'm getting better at these mocks. :)

Issue #503, opened 2016-12-05
Select#select and Select#select_value should wait for element to be present


jkotests:
There was a recent Stack Overflow question, [Any new arrangement for select_list in WATIR 6.0.2 waits until options got populated?](http://stackoverflow.com/q/40968543/1200545), which asked about selections in one dropdown populating another dropdown. I recall there being other users with similar problems in the past.

The new wait functionality for actions does not appear to apply to selecting options from select lists. A Watir::Exception::NoValueFoundException is immediately thrown when the option is not present.

~~~~~~~~
browser.select_list.select('changed_text')
#=> Watir::Exception::NoValueFoundException
~~~~~~~~

It would be nice if the `#select` and `#select_value` methods waited - ie equivalent to the following, which does currently wait:

~~~~~~~~
browser.select_list.option(text: 'changed_text').select
~~~~~~~~

titusfortner:
This should fix it: 512f848

PR #516, opened 2017-01-03
rewrite the select_list to match other elements

titusfortner:
I started writing this to address #503 
Then I ended up refactoring the whole class because I found it confusing...

This is much cleaner and all the specs pass, but it probably needs a once or twice over to make sure I'm not missing the point of all the methods I deleted. :)


p0deje (approved):
Let's merge if specs are passing and wait for bug reports :)

titusfortner:
merged - 512f848

PR #520, opened 2017-01-07
implement keyword support

titusfortner:
Being able to set a keyword value for an element is especially useful for page object implementations. Rather than trying to match the selector output to a defined method, the user can now just look up the name of the element assigned. It would need to be implemented in the Page Object gems, but this provides the necessary hook for that.

Also, I changed the error messages to use `#selector_string` when really they should be using `#inspect` like was there before. `#selector_string` is still recursive, but `#inspect` includes information about "located", "class name" and the new "keyword." (I removed the memory information; that is still available with `Element#to_s`if anyone cares about it).

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -12,6 +12,8 @@ module Watir
     include EventuallyPresent
     include Waitable
 
+    attr_accessor :keyword
+
     #
     # temporarily add :id and :class_name manually since they're no longer specified in the HTML spec.
     #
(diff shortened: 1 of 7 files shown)


lib/watir/elements/element.rb
```diff
@@ -49,10 +51,13 @@ def exists?
     alias_method :exist?, :exists?
 
     def inspect
+      string = "#{self.class}: "
```
p0deje: Let's keep it look like a usual Ruby inspect which wraps class in `#<Class>`:

```ruby
string = "#<#{self.class}: "
# ...
string << '>'
```

spec/element_spec.rb
```diff
@@ -106,33 +106,53 @@
     end
   end
 
-  describe "#selector_string" do
-    it "displays selector string for regular element" do
-      browser.goto(WatirSpec.url_for("wait.html"))
+  describe "#inspect" do
+    before(:each) { browser.goto(WatirSpec.url_for("nested_iframes.html")) }
```
p0deje, line 110: please add empty line below

p0deje (approved):

PR #521, opened 2017-01-09
implement MixedElementCollection class for siblings and children

titusfortner:
here's a potential supplemenatal idea to remove the need for generic HTMLElementCollection

### DIFF
--- a/lib/watir/mixed_element_collection.rb
+++ b/lib/watir/mixed_element_collection.rb
@@ -1,0 +1,53 @@
+module Watir
+
+  #
+  # Base class for mixed element collections.
+  #
+
+  class MixedElementCollection < ElementCollection
+
+    #
+    # This collection as an Array.
+    #
+    # @return [Array<Watir::Element>]
+    #
+
+    def to_a
+      @to_a ||= elements.map.with_index do |e, idx|
+        Watir::HTMLElement.new(@query_scope, @selector.merge(element: e, index: idx)).to_subtype
+      end
+    end
+
+    #
+    # Creates a MixedElementCollection containing elements of two collections.
+    #
+    # @example
+    #   (browser.select_list(name: "new_user_languages").children + browser.select_list(id: "new_user_role").children).size
+    #   #=> 8
+    #
+
+    def +(other)
+      case
+      when to_a.empty?
+        other
+      when other.to_a.empty?
+        self
+      else
+        other.to_a.each { |el| to_a << el }
+        @to_a = to_a
+        self
+      end
+    end
+
+    private
+
+    def element_class_name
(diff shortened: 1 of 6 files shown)


lib/watir/mixed_element_collection.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def element_class_name
+      "HTMLElement"
+    end
+
+    def element_class
+      Watir::HTMLElement
+    end
+
+  end # ElementCollection
```
p0deje, line 52: Wrong comment

p0deje (commented):
I'm not sure if it's really needed, so up to you to decide should this be added or not.

p0deje:
Maybe we should instead automatically cast element to subtypes all the time, so that user never has to work with `Watir::Element` and then we can have only one collection class which will be a mixed one.

titusfortner:
I could get behind that idea. Might slow things down if someone did a `browser.elements` call?

p0deje:
We already retrieve tag name for all elements when locating them in `lib/watir/locators/element/validator.rb` so it shouldn't be slower.

titusfortner:
We only validate when retrieving a single element with a tag specified. `browser.element(id: 'foo')` will return an `HTMLElement` and `browser.elements` will return an `HTMLElementCollection` of `HTMLElement`. Likewise `browser.divs` returns a `DivElementCollection` of `DivElement` which is never validated. (I'm ok with this last one, and we probably validate more than we need to in general).

titusfortner:
Yeah, I'm not sure what makes the most sense here, either.

@jkotests or anyone else have opinions on this one?

jkotests:
It would be great to eliminate the need for users to call `#to_subtype` as its need is not obvious.

However, I am not sure I understand the details of the `MixedElementCollection`. Why would `#children` return a `MixedElementCollection` when `#elements` does not? To me, the only difference should be the depth of element search. The return type should be consistent.

titusfortner:
I guess I rarely use `#elements` directly, or `#to_subtype` for that matter, so it wouldn't seem to me to be much of an issue, but maybe others have a use case for it. 

You're right, though, there is no real difference here because you can specify the tag name either way.

p0deje:
> We only validate when retrieving a single element with a tag specified. 

Not exactly, we retrieve tag name for any element (except when they are looked up using `:id`) but validate only when tag name is specified.

See https://github.com/watir/watir/blob/b2b485f8deb2232891d939602bf72f15abebe30b/lib/watir/locators/element/validator.rb#L6-L11

titusfortner:
Closing in favor of #531 please add any comments you have on this idea there. Thanks!

PR #531, opened 2017-02-09
To subtype

titusfortner:
Removes support for `#siblings` since combining HTMLCollections seems like more trouble than it is worth. Doing it "correctly" would mean figuring out how to handle multiple selector signatures, etc, and if users really want it they can iterate over both the previous & next collections themselves.

Also uses `Element#to_subtype` to objects in `ElementCollection` in lieu of #521

### DIFF
--- a/lib/watir/adjacent.rb
+++ b/lib/watir/adjacent.rb
@@ -1,5 +1,3 @@
-require "active_support/inflector"
-
 module Watir
   module Adjacent
 
@@ -73,19 +71,6 @@ module Watir
     alias_method :next_siblings, :following_siblings
 
     #
-    # Returns collection of all sibling elements of current element.
-    #
-    # @example
-    #   browser.text_field(name: "new_user_first_name").following_siblings.size
-    #   #=> 52
-    #
-
-    def siblings(opt = {})
-      raise ArgumentError, "#siblings can not take an index value" if opt[:index]
-      preceding_siblings(opt) + following_siblings(opt)
-    end
-
-    #
     # Returns element of direct child of current element.
     #
     # @example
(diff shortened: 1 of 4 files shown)


lib/watir/adjacent.rb
```diff
@@ -122,10 +107,14 @@ def xpath_adjacent(direction = '', opt = {})
         raise ArgumentError, "unsupported locators: #{opt.inspect} for ##{caller} method"
       end
 
-      xpath = "./#{direction}#{tag_name || '*'}"
-      tag_name ||= 'element'
-      return self.send(tag_name.to_s.pluralize, {xpath: xpath}) unless index
-      self.send(tag_name, {xpath: "#{xpath}[#{index + 1}]"})
+      klass = self.send(tag_name).class if tag_name
```
p0deje:
```ruby
if index
  klass = tag_name ? self.send(tag_name).class : HTMLElement
  klass.new(self, xpath: "./#{direction}#{tag_name || '*'}[#{index + 1}]")
else
  klass = tag_name ? Object.const_get("#{klass}Collection") : HTMLElementCollection
  klass.new(self, xpath: "./#{direction}#{tag_name || '*'}")
end
```

p0deje (changes requested):
Overall looks good:

1. There is a small comment I've left.
2. Let's add spec for getting collection of `elements` which return subtypes.

titusfortner:
The `adjacent_spec.rb` essentially already tests that functionality, but it makes sense to make it more explicit and add specs to `collection_spec.rb`.

titusfortner:
Merged - 7bd172cca9dbddb6d16a61016b8ac86218121d19 & 786fb126482495ef7aa2fd6453908c290430bd65

Issue #532, opened 2017-02-21
Retry Element Not Clickable


titusfortner:
The W3C spec is defining [element interactability](https://w3c.github.io/webdriver/webdriver-spec.html#element-interactability), and requiring it to be checked before clicks & send keys. It is even getting [its own error code](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-element-not-interactable).

A user on Slack had an issue with a css animation covering the element in a way that it was temporarily visible, but not clickable. We should consider rescuing & retrying similar to a stale element error.

titusfortner:
I think we decided that this isn't the right solution to this problem, so I'm closing.

titusfortner:
So, we need some kind of solution to this. I think it requires `Watir::Executor` and us figuring out which the default needs to be. :)

Rajagopalan-M:
@titusfortner Hi, This is one of the essential feature, whats the problem in adding this feature? Recently Firefox has given the exception class for this error. 

Selenium::WebDriver::Error::ElementClickInceptedError

So we can rescue as we are rescuing for some of the other errors.

titusfortner:
Because there are multiple reasons for a click to be intercepted, so we don't want to force wait on this always. You can write your own method that rescues this exception and does the wait you need. The long term fix is to implement an Executor that can be customized by the user as desired.

Rajagopalan-M:
@titusfortner  I know the various conditions, there are permanent overlay and there are temporary overlay,Its not suitable for permanent overlay, but what's the harm in rescuing this? If it's permanent overlay, user will anyhow come to know the error after 30 seconds, he is not going stuck there indefinitely. Yes?

titusfortner:
In principle I completely agree with you. We just need to allow the option for users with a permanent overly to rescue the exception and handle it differently. If Watir auto rescued and retried, then they'd be dealing with a Timeout exception instead of a click intercepted exception.

Rajagopalan-M:
@titusfortner  but timeout execption in WATIR is clearly telling us about timeout because of what! It's telling us it's timeout because unable to locate element or timeout for element to be present ,yes? So we need to include the error for this as well, for an example timeout because of element overlay. Actually either we can create one more rescue section or we may include this exception class along with other exception class by simply putting comma and we need to receive the error by =>e and then we need to inform the user about the specific error.

titusfortner:
[...]

PR #534, opened 2017-02-28
#532 - Retry click when element is not clickable

BluePhoenix:


lib/watir/elements/element.rb
```diff
@@ -640,6 +640,10 @@ def element_call(exist_check = :wait_for_exists)
         yield
       rescue Selenium::WebDriver::Error::StaleElementReferenceError
         retry
+      rescue Selenium::WebDriver::Error::UnknownError => ex
+          # Chromedriver and previous versions of Geckodriver throw UnknownError when element not clickable
```
titusfortner: s/previous versions of Geckodriver/Legacy Firefox Driver

titusfortner:
Addresses #532 

@p0deje - I think I proposed this once a couple years ago, but either you or Jari were against it because it was a browser specific error message (Chrome only at that time). Since this is going to be a part of the W3C, I think we should add this as is, then update it to rescue the specific W3C error class when Watir moves to Se 4.

Hmm, and actually @BluePhoenix we might want to see if we can create a spec for this as well. I think there is a Ruby spec in Selenium for getting this error, and we'd need to duplicate that html and add javascript code for it to go away after a brief time (we have that code in the wait specs). Let me know if we should pair on this.

BluePhoenix:
@titusfortner Yeah, pairing might be required. But for now, I could do an initial attempt myself first. You think the spec would be inside https://github.com/watir/watir/blob/master/spec/click_spec.rb or it would be a new _spec.rb file?

This is the one @p0deje mentioned earlier, but not sure if this is what you want us to implement in the spec: https://github.com/SeleniumHQ/selenium/blob/master/rb/spec/integration/selenium/webdriver/element_spec.rb#L30-L40 .

titusfortner:
Yeah, find the html that allows that spec to work (`click_tests/overlapping_elements.html`) in the Selenium project. I forget where the html files are stored somewhere in the common code. Then recreate it in Watir. Probably should put it wherever the stale element rescue code is? Maybe?

p0deje:
In Toptal we deal with a bunch of various "unclickable" elements and most of them require different handling. Here is an example of a code:

```ruby
click = instance_method(:click)
define_method(:click) do |*args|
  Features::Helpers::FlashMessagesHandler.close_all_notice_messages(browser.driver)
  begin
    click.bind(self).call(*args)
  rescue Selenium::WebDriver::Error::UnknownError => error
    message = error.message
    if element_not_clickable?(message)
      element_received_click = message[/Other element would receive the click: (.+)$/, 1]
      element_received_click = Nokogiri::HTML(element_received_click)

      if nps_overlay?(element_received_click)
        discard_nps
        retry
      end

      if scrolling_overlay?(element_received_click)
        wait_for_scrolling_overlay_to_disappear
        retry
      end

      if sticky_flash_message?(element_received_click) || sticky_header?(element_received_click)
        scroll_half_page_up
        retry
      end

      if tooltip?(element_received_click)
        hide_tooltip
        retry
      end

      if bounce_modal?(element_received_click)
        close_bounce_modal
        retry
      end

      if side_menu?(element_received_click)
        wait_for_menu_animation
        retry
      end
    end

    raise error
  end
end
```

with that said, I personally see no point in blindly retrying click upon this error (also, this PR will make an infinite loop).

titusfortner:
Oh, right, because we can get recursive calls, the loop break is based on the timer in the Wait class which won't apply here. Ok.

But wouldn't blindly retrying fix your use cases for `wait_for_scrolling_overlay_to_disappear` & `wait_for_menu_animation`?

p0deje:
> But wouldn't blindly retrying fix your use cases for wait_for_scrolling_overlay_to_disappear & wait_for_menu_animation?

They are close but not similar (waiting for DOM changes on different elements). However, even if we retry clicking 3 times with a small delay between them that won't be reliable (just like any other hardcoded sleeps).

titusfortner:
I guess I see a fundamental difference between needing to take an action for something to work and something that will work if only given more time. The idea behind this is explicitly to avoid hard coding a sleep by polling for the blocking item to go away. I get that this implementation gets in the way of more complex solutions, though. 

I'm not sure what `self` is in your code, but it looks like that's essentially an extensive monkey patch of `Watir::Element`. I'm wondering if there is a way we can do this to provide a default retry (because the site will eventually synchronize with the state of the test code), but allow it to be overridden with a more complicated handling of things that require more active adjustments.

p0deje:
> I'm wondering if there is a way we can do this to provide a default retry (because the site will eventually synchronize with the state of the test code)

So if Watir provides `Element#clickable?` we can easily implement "relaxed_click", but I'm not sure how hard it would be. Otherwise, I don't think we can provide default retry because there is a big chance that site will *not* synchronize with the test code.

titusfortner:
Yeah, I don't think we can do `Element#clickable?` without actually clicking it, even with updated implementations in w3c spec...

Spitballing... what would it look like to do something like...

```
@element.click
rescue Exception => ex
        browser.after_hooks.run(ex)
```

and let `#run` do:
```
def run(ex=nil)
      if @after_hooks.any? && @browser.window.present?
        each { |after_hook| after_hook.call(@browser, ex) }
      end
     raise ex if ex
end
```

Might be too complicated and/or not do what I actually want... I'd need to write a bunch of specs for it because scopes of the Procs might prevent this from being useful.

BluePhoenix:
Maybe this is short sighted of me, but the way that I see this would ideally work is similar to how click waits for elements to exist.
But in this case, 
1. click would wait for element to exist if not covered (like it does now)
2. click would wait for element to not be covered if it exists **and is covered**

Seems one of the bigger challenges is that there really is no way to know before actually clicking.

I'll update this branch with the specs because at least it gives us something to try things against. There is a current issue (as mentioned before) that the test will run infinitely with the current code, so I also added a spec to cover that scenario.

titusfortner:
@BluePhoenix - the problem with this approach is that it would get in the way of alternative handling. We would need to find a more complicated way to make a retry the default behavior, but still allow for more complicated things like what Alex is doing. I'm not sure what that would look like. Sorry for sending you down a dead end road.

enkessler:
Just today I had to create a wrapper method for clicking elements for the sole purpose of catching non-clickable errors, retrying a few times, and then giving up and re-raising the exception if the problem persisted.

Between automatically waiting for and element to exist, not be stale, or become clickable, I don't see much difference such that the first two would be supported but the last one would not. All of these states should be temporary and when they are not, it is perfectly appropriate for them to time out and cause an error because there is clearly a problem with either the application or the test and that problem should be addressed.

It is true that having a timeout instead of an immediate exception would get in the way of trying to fix things on the fly like in the example provided by @p0deje, but that kind of exception handling indicates a lack of understanding of the flow of the application being tested or an inability to control the test environment (i.e. stuff really shouldn't randomly happen during your test). Which is itself a larger and separate problem beyond the scope of Water.

That being said, an immediately returning `#clickable?` method analogous to the `#exists?` method would be pretty important to this kind of change.

p0deje:
It would be awesome to have `#clickable?` but unfortunately WebDriver doesn't provide this command so the only correct way to check if element is clickable is to actually perform click.

enkessler:
Hmm. Given the choice between automatic waiting or the ability to know immediately if an element is clickable, I'd choose the later (even if that immediate knowledge comes from it immediately blowing up in my face). Because I can always add extra wait time myself but I cannot remove wait time that is baked in and having `#clickable?` available is what would prevent me from otherwise needing to remove wait time.

So...both or neither, is what I am saying and it sounds like the second one isn't technically possible right now. So neither it is.

titusfortner:
@enkessler the problem is that whether an element is "clickable" is even less knowable than whether it is actually present. The driver can make a better stab at the former than the latter. Check out [David Burns presentation](https://codeandtalk.com/v/seleniumconf-uk-2016/is-it-or-is-it-not-really-visible-seleniums-flawed-vision-david-burns) that details some of these issues. It was also specifically [discussed with respect to the w3c standard](https://github.com/w3c/webdriver/issues/395).

Watir can only support the things the drivers themselves can support. What framework are you using? It is possible to wait for the javascript action to complete before attempting the click in the first place.

I'm closing this issue because I don't think there is any way we're going to be able to implement something that will address this for reasons beyond our control.

PR #538, opened 2017-03-06
Fix exception thrown due to looking for `reset!` method on wrong object

DJCecil2:


DJCecil2:
Hey all. Been using Watir for about a year, first time contributor. Found an exception lurking here and I think this is the fix.

I have this overridden in my local codebase for now. Seems to be working. Tests are passing.

Let me know if you'd like a unit test for this?

titusfortner:
Whoops, yup, I did the wrong thing when implementing the new logic for `#wait_while_present`. Thanks for catching it and fixing it!

DJCecil2:
👍 Thank you, sir, for such a fine ruby gem! Been thoroughly impressed with it.

titusfortner:
Glad you're getting good use from it. We have a somewhat active Slack channel if you want to participate in any of the conversations / coding discussions. (join the #Watir channel from Selenium Slack: http://seleniumhq.herokuapp.com/)

PR #540, opened 2017-03-11
use javascript to get selected options

titusfortner:
This PR is based on [this Stack Overflow question](http://stackoverflow.com/questions/42724129/watir-web-driver-taking-too-long-to-get-the-selected-item-from-drop-down), and @jkotests's response.

The reason to use Selenium calls instead of a javascript call is typically because Selenium attempts to evaluate the visibility of an option. In the case of obtaining selected options, I don't think it is possible for it to be hidden. As such I think using javascript here for performance reasons is a reasonable alternative.

@p0deje what do you think?

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -112,7 +112,7 @@ module Watir
     #
 
     def selected_options
-      options.select { |e| e.selected? }
+      browser.execute_script("return arguments[0].selectedOptions;", self)
     end
 
     private


p0deje:
According to quick googling and MDN, it's not supported in Internet Explorer and is not very reliable, so I don't think we should switch to it.

titusfortner:
***shakes fist at Microsoft***
Thanks for checking on that.

abotalov:
By the way, there is `selectedIndex` that can be arguably used to speed up `#value` or not existing `#selected_option`.

titusfortner:
@abotalov - can you show the code for what you are suggesting? Thanks.

abotalov:
See e.g. http://stackoverflow.com/a/610344/841064

```ruby
def value
  selected_option && selected_option.value
end

def selected_option
  browser.execute_script("el = arguments[0]; return el.options[el.selectedIndex];", self)
end
```

imohdahmed:
Hello Folks,

Im new to Watir. Please help me to do a automation data entry to website for www.facebook.com page.
I'm not able to select the dropdown list. while I'm doing i get the error as shown

#<Watir::Select: located: false; {:class=>"b_5dba", :tag_name=>"select"}>

jkotests:
I believe that `selectedIndex` only returns the index of the first selected option, which means it won't work for multi-select lists.

We might need to iterate over the options;

~~~~~~~~ruby
def selected_options
  script = <<-SCRIPT
    result = [];
    options = arguments[0].options;
    for(i = 0; i < options.length; i++) {
      opt = options[i]
      if(opt.selected) { result.push(opt) }
    }
    return result;
  SCRIPT
  browser.execute_script(script, self)
end
~~~~~~~~

It's not as fast as `selectedOptions`, but still a lot faster than the existing method. Seems to work in at least Chrome and IE11.

PR #542, opened 2017-03-15
Optimize Select#selected_options and Select#value using JS

abotalov:
Continuation from #540

### DIFF
--- a/lib/watir/elements/image.rb
+++ b/lib/watir/elements/image.rb
@@ -23,8 +23,9 @@ module Watir
     #
 
     def width
-      wait_for_exists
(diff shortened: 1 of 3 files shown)


titusfortner:
If we don't call `#options` we still need to add a line for `#wait_for_exists` to make sure select list is there before working with it.

abotalov:
Done.

p0deje:
There is always a chance element goes stale during script execution, so this should probably be wrapped with `element_call(:wait_for_exists)`.

titusfortner:
merged: 1f22570e0cb46e66ca9296f5837fb78792d99e20

PR #554, opened 2017-05-03
Updated the flash function

Paul3816547290:
Now the caller can control the color, number of flashes, and duration between flashes

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -215,12 +215,15 @@ module Watir
     #
     # Flashes (change background color to a new color and back a few times) element.
     #
-    # @examples
+    # @example
     #   browser.text_field(name: "new_user_first_name").flash
     #   browser.text_field(name: "new_user_first_name").flash("green",3,0.05)
+    #   browser.text_field(name: "new_user_first_name").flash(color:"red")
+    #   browser.text_field(name: "new_user_first_name").flash(flashes:4)
+    #   browser.text_field(name: "new_user_first_name").flash(delay:0.1
     #
 
-    def flash(color="yellow",flashes=5,delay=0.1)
+    def flash(color: 'yellow', flashes: 10, delay: 0)
       background_color = style("backgroundColor")
       element_color = driver.execute_script("arguments[0].style.backgroundColor", @element)
 
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -213,26 +213,28 @@ module Watir
     end
 
     #
-    # Flashes (change background color far a moment) element.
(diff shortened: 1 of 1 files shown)


Paul3816547290:
https://github.com/watir/watir#note-on-patchespull-requests
Suggests I should write tests - happy to do that but dont know where these are or how they are executed

p0deje:
Thank you for pull request!

> Suggests I should write tests - happy to do that but dont know where these are or how they are executed

I think this is fine assuming you've added new example in documentation which is automatically tested.

lib/watir/elements/element.rb
```diff
@@ -213,26 +213,28 @@ def drag_and_drop_by(right_by, down_by)
...
+    # Flashes (change background color to a new color and back a few times) element.
     #
-    # @example
+    # @examples
     #   browser.text_field(name: "new_user_first_name").flash
+    #   browser.text_field(name: "new_user_first_name").flash("green",3,0.05)
     #
 
-    def flash
+    def flash(color="yellow",flashes=5,delay=0.1)
```
p0deje: Let's use keyword arguments so that you can pass just one of them (e.g. you only want to control number of flashes):

```ruby
def flash(color: 'yellow', flashes: 5, delay: 0.1)
  ...
end
```

lib/watir/elements/element.rb
```diff
@@ -213,26 +213,28 @@ def drag_and_drop_by(right_by, down_by)
     end
 
     #
-    # Flashes (change background color far a moment) element.
+    # Flashes (change background color to a new color and back a few times) element.
     #
-    # @example
+    # @examples
```
p0deje: Do not change it

lib/watir/elements/element.rb
```diff
@@ -213,26 +213,28 @@ def drag_and_drop_by(right_by, down_by)
...
     #   browser.text_field(name: "new_user_first_name").flash
+    #   browser.text_field(name: "new_user_first_name").flash("green",3,0.05)
```
p0deje: Please, keep the code style:

```ruby
browser.text_field(name: "new_user_first_name").flash("green", 3, 0.05)
```

lib/watir/elements/element.rb
p0deje: Also, let's keep old defaults:

* flashes - 10
* delay - 0

p0deje (changes requested):
Overall it looks great, thank you for contribution.

However, let's fix few things before we merge this.

Paul3816547290:
All amendments suggested completed and remerged

p0deje:
Merged in c30132eb46f570fb50d70c96de9e977ba9e4772c, thank you for contribution!

PR #558, opened 2017-05-24
support for map of text of elements in ElementCollection

titusfortner:
Prompted by this: https://stackoverflow.com/questions/44154398/ruby-code-to-display-table-element-details

But I use map a lot in debugging and I pretty much always forget to do #to_a first. This should make it easier to do what I expect is a common action.

I'm not sure I have the comment documentation correct, though.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -74,6 +74,28 @@ module Watir
     end
 
     #
+    # List of block calls on each element in this collection
+    #
+    # @param [Block]
+    # @return [Proc]
+    #
+
+    def map &block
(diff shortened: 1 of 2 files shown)


jkotests:
Isn't `#map` already supported without the need for `#to_a`? For example, Watir already allows `browser.links.map(&:text)`.

titusfortner:
huh, I would have sworn that didn't work when I was debugging things last month. For some reason I thought that was an Array thing and not an Enumerable thing. Ok, I can remove that. :)

p0deje:
Agree with Justin, it already works.

titusfortner:
`#map(&:text)` is close enough to `#text` for this to be unnecessary.

PR #559, opened 2017-05-28
Validate that path has valid extension when saving a screenshot (closes #400)

### LINKED ISSUE #400: Screenshot#save should validate file extension
Currently this code works:

``` ruby
browser.screenshot.save "foobar.jpg"
```

The problem is, however, that (at least with chrome-driver) the screenshot will be saved in `png` format instead. I'm not sure if it differs between drivers, but I suggest either to add extension validation into `watir-webdriver` or convert files into proper format in the end.

Code above caused me many problems in different software - for example even `Photoshop CC 2015` didn't open that kind of files because it thought it is `jpg` although it was a `png` and `GoPro Studio` just crashed.

Yes, all this software could be smarter, but to avoid that kind of a problem we could have a safe-guard in `watir-webdriver`.

`watir-classic` for example does use `win32screenshot` to take screenshots and it will validate extensions and save screenshots into different formats depending of the given extension (`jpg` is among of these formats).

Thoughts?

abotalov:
Webdriver spec [says that only PNG is supported](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-encode-as-base64-a-canvas-element).

### DIFF
--- a/lib/watir/screenshot.rb
+++ b/lib/watir/screenshot.rb
@@ -1,6 +1,8 @@
 module Watir
   class Screenshot
 
+    PNG_EXTENSIONS = ['.png']
+
     def initialize(driver)
       @driver = driver
     end
@@ -12,9 +14,14 @@ module Watir
     #   browser.screenshot.save "screenshot.png"
     #
     # @param [String] path
+    # @raise [ArgumentError] if path has a non-png extension
     #
 
     def save(path)
+      extension = File.extname(path).downcase
+      if !extension.empty? && !PNG_EXTENSIONS.include?(extension)
+        raise ArgumentError, "#save can only save PNG files so paths with non-PNG extensions are not supported"
+      end
       @driver.save_screenshot(path)
     end
 
--- a/lib/watir/screenshot.rb
+++ b/lib/watir/screenshot.rb
@@ -20,7 +20,7 @@ module Watir
     def save(path)
       extension = File.extname(path).downcase
       if !extension.empty? && !PNG_EXTENSIONS.include?(extension)
-        raise ArgumentError, "#save can only save PNG files so paths with non-PNG extensions are not supported"
+        warn "#save will save an image in PNG format because Webdriver supports only PNG screenshots"
       end
       @driver.save_screenshot(path)
     end
--- a/spec/watirspec/screenshot_spec.rb
+++ b/spec/watirspec/screenshot_spec.rb
@@ -19,28 +19,28 @@ describe "Watir::Screenshot" do
 
   describe '#save' do
     it 'saves screenshot to given file' do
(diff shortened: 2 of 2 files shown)


p0deje:
Looks good to me, however I see that there is a failing spec now - https://ci.appveyor.com/project/p0deje/watir/build/%23175/job/nx6srh2qh8csua9v, can you fix it?

titusfortner:
Except Watir isn't claiming to be W3C compliant right now. jpg works just fine in chromedriver & geckodriver, and I don't think this version of Watir should prevent users from that format if they want.

p0deje:
#400 says that even if you specify .jpg it will still be PNG image when used with chromedriver.

abotalov:
@titusfortner Relevant commands in bridges don't take any format parameters, see https://github.com/SeleniumHQ/selenium/blob/ceaf3da79542024becdda5953059dfbb96fb3a90/rb/lib/selenium/webdriver/common/driver_extensions/takes_screenshot.rb#L46
.NET has an ability to save screenshot as jpg by converting png to jpg - https://github.com/SeleniumHQ/selenium/blob/1e221937594c8ac4895899024927575aa6015a60/dotnet/src/webdriver/Screenshot.cs

abotalov:
I have now changed code to warn about invalid extension rather than raising an error because of Semver.

titusfortner:
Oh, interesting. Shouldn't this verification happen in Selenium, then?

titusfortner:
Also, defining a constant array with one value seems unnecessary to me.

abotalov:
Defining a constant has the aim to allow users to add additional extensions if they would like to save PNG screenshots with different extensions.

p0deje:
We should not allow mutating constants, we should instead freeze it.

abotalov:
> Oh, interesting. Shouldn't this verification happen in Selenium, then?

I don't think so.
Other language bindings don't do validation.
Also some Selenium does its job - it saves a PNG screenshot to a path provided by the API user. It's not Selenium's fault that API user expected media format to inferred from file extension.

titusfortner:
That's what I'm saying, we shouldn't have mutable constants, as if it isn't mutable then it is overkill for one item. We could create a method that could be subclassed, or create a keyword to override exception, or we leave it as a warning.

I think the method itself should just set it to png & PNG. Or maybe we go the other way and complain if it is jpg, jpeg, etc, since that's probably the real concern.

As for Watir vs selenium, Watir is just a pass through here. If selenium users think they are getting a jpg, but aren't, they should at least be warned.

abotalov:
This is now handled in Selenium - https://github.com/SeleniumHQ/selenium/issues/4108

titusfortner:
Fixed in ruby or just python?

abotalov:
@titusfortner I meant "It's planned to be handled in Selenium" :)

PR #560, opened 2017-05-31
Add #text method for select lists

tnypxl:
If I want to access the text of a selected option in a select list that doesn't use the multiple attribute, I have to do this:

`some_select_list(name: 'list').selected_options.first.text`

While its not the most egregious thing ever, its just inconvenient if the majority of your select lists don't use the multiple attribute. This PR allows the following:

`some_select_list(name: 'list').selected_option.text`

**UPDATE**:

After some discussion with @titusfortner and other folks on Slack, `selected_option` was a bit dangerous when used with select lists that use the `multiple` attribute. Following the same pattern as the `value` method, I opted for a `text` method instead.

If you need to output the text of everything inside a select_list, I would just map/collect them.

`select_list.options.map(&:text)`

Let me know your suggestions.

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -106,7 +106,7 @@ module Watir
 
 
     #
-    # Returns an array of currently selected options.
+    # Returns a single selected option.
     #
     # @return [Array<Watir::Option>]
     #
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -104,25 +104,25 @@ module Watir
       o.value
     end
 
-
     #
-    # Returns a single selected option.
+    # Returns the text of the first selected option in the select list.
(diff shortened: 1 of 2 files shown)


titusfortner (approved):
Looks good to me. 
I was going to suggest some changes, but I see you copied what `#value` did, which means we'd want to update syntax across the class/project, which is out of scope for this.
Thanks!

tnypxl:
@titusfortner There wasn't anything special about `element.text` that warranted a different approach from `element.value`. I wanted there to be so it wasn't such a blatant dup of `value`. lol. Thanks for review!

p0deje (approved):

titusfortner:
merged with 7c530eb29d21c8de02e28a9f51a1896cbbee60aa

PR #562, opened 2017-06-16
reduce number of iterative calls in waits

titusfortner:
To improve performance, this code changes how many wire calls are made verifying state before attempting to take action.

On an already located element, `#click` currently makes 9 wire calls:
1. enabled
2. window
3. windows
4. frame
5. enabled
6. displayed
7. enabled
8. enabled
9. click

With this PR:
1. enabled
2. click

As a non-representative example:
https://github.com/titusfortner/address_book/blob/master/spec/2watir_spec.rb
This went from 485 wire calls to 308 (and Sauce labs execution time of 150 seconds to 100 seconds).

The tradeoffs
1. No longer see order of precedence for children/parents in failures
2. Removes the existing warning messages about why tests may be timing out when upgrading from watir-webdriver
3. I honestly have no idea if this code is properly reusing the timeout at this point, but I don't think it matters

From an implementation standpoint the tricky parts are:
1. Drivers do not error when user clicks on disabled things, but Watir needs an error thrown, so Watir needs to check that the element is not disabled before attempting to click. W3C spec says that "enabled" end point is only applicable for form controls, so I've restricted clicks requiring this check to just those elements. 
2. Drivers do not error when user sends text to a read-only text field, but Watir needs an error thrown, so Watir needs to check that the element is read-only before attempting to send text.


vveliev:
hmmmmm , never checked that before.... what is the need of making all this checks ?! I don't think that since selenium 2.45+  you can click an element that is not visible......it will give you an error

p0deje:
I don't fully follow what's going on here, but let's ship it and see!

p0deje (approved):

titusfortner:
@vveliev 
Here's a quick run down of what Watir is doing with each wire call:

enabled - checking to make sure that the cached element isn't stale
window & windows - making sure that the current window handle is in the list of open windows
frame - switch to default context (necessary for how we do our frame handling)
enabled & displayed- verifies the element is visible (if not, it will wait for it to become visible)
enabled & enabled - verifies the element can be clicked (if not it will wait for it to become enabled)
click

Alternately we can just make the click and do a better job at handling errors that the driver returns. Made more tricky by the two caveats above.

abotalov:
There could be a race when element is e.g. visible, then it quickly becomes invisible and then it becomes visible again.
Capybara wraps action methods using `synchronize` to solve such issues (and rescues a number of errors - https://github.com/teamcapybara/capybara/blob/db779f3b0d32cb787fd1a3329f68afdf5b4235d2/lib/capybara/selenium/driver.rb#L281)

titusfortner:
This code should actually better handle that use case. If an action is taken and it is in the not-displayed state, it will rescue that exception and retry. The current code is what checks for displayed, assumes it is in a good condition and continues without rescuing exceptions.

titusfortner:
merged with #568

PR #563, opened 2017-06-17
Performance context

titusfortner:
This rescues window not found error instead of making two wire calls on each element location to make sure it is there.

It also uses `@default_context` variable to allow only sending null to frame endpoint when it isn't already in the default browsing level context.

Coupled with #562 I'm actually seeing a 50% reduction in test run times on Sauce, and the number of wire calls in my example tests went from 485 —>  308 —> 207


titusfortner:
merged with #568

PR #564, opened 2017-06-17
don't unnecessarily validate elements

titusfortner:
@p0deje - are we ok to do do this since text_field no longer locates both `input` and `textarea`?

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -44,10 +44,8 @@ module Watir
                       find_first_by_multiple
                     end
 
-          # This actually only applies when finding by xpath/css - browser.text_field(:xpath, "//input[@type='radio']")
-          # We don't need to validate the element if we built the xpath ourselves.
-          # It is also used to alter behavior of methods locating more than one type of element
-          # (e.g. text_field locates both input and textarea)
+          # Validation not necessary if watir builds the xpath
+          return element unless @selector.key?(:xpath) || @selector.key?(:css)
           element_validator.validate(element, @selector) if element
         rescue Selenium::WebDriver::Error::NoSuchElementError, Selenium::WebDriver::Error::StaleElementReferenceError
           nil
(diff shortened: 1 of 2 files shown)


lib/watir/locators/element/locator.rb
```diff
@@ -44,10 +44,8 @@ def locate
                       find_first_by_multiple
                     end
 
-          # This actually only applies when finding by xpath/css - browser.text_field(:xpath, "//input[@type='radio']")
-          # We don't need to validate the element if we built the xpath ourselves.
-          # It is also used to alter behavior of methods locating more than one type of element
-          # (e.g. text_field locates both input and textarea)
+          # Validation not necessary if watir builds the xpath
+          return element unless @selector.key?(:xpath) || @selector.key?(:css)
```
p0deje, line 48: You should move this to lib/watir/locators/element/validator.rb so that watizzle and watir-css also don't validate elements.

p0deje (approved):
There is a small change to be made, but overall let's merge!

lib/watir/locators/element/locator.rb
titusfortner: Both of those will be affected by line 48 still; are you saying you want this in the `Validator#validate` method so that it can be overridden by watir-css / watizzle / etc? (rather than always short circuiting here?)

p0deje: Yeah, that's what I meant. For example, watizzle does not implement custom validations and uses the one that is provided by Watir, which means that this fix would work there as well if it is done inside Validator class.

titusfortner: Oh, I see what you are saying now, sorry, was confused. Yes, that's the correct place for this code, I will update. Thanks.

titusfortner: Ugh, except this won't work without also updating [this line](https://github.com/p0deje/watizzle/blob/master/lib/watizzle/locators/element/locator.rb#L33) since we also want to validate `:id` parameters, but we can't check it in `Validator#validate` because we delete the key. :-/

I think the subclasses of `Locator#locate` will need to be updated for this to work.

titusfortner:
merged with #568

PR #565, opened 2017-06-18
6 2 2

titusfortner:
I'd like to get a release out there before the major performance overhaul that is in the works.

As for using `webdrivers` gem: it simplifies Travis code, but depending on your ruby environment setup you might need to run `gem uninstall webdrivers` when running Selenium projects that don't include it in the gemfile after running watirspecs.

Regardless, our code is now happy on Travis with the latest drivers

@abotalov I had to tweak `#selected_options` things a bit to get it to work inside frame contexts

PhantomJS: any problems stopping running specs with phantomjs? Maybe "Deprecate PhantomJS Support" is too severe for the changelog?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 ### 6.2.2 (Unreleased)
 
 * Allow locating elements by attribute presence/absence (#345)
+* Element#flash configurable by color, number, and delay (thanks Paul3816547290)
+* Implement Select#text like Select#value (thanks Arik Jones)
+* Optimize Select#selected_options, Select#value, and Select#text with javascript (thanks Andrei Botalov)
+* Support #execute_script from inside frame context
+* Deprecate PhantomJS Support
 
 ### 6.2.1 (2017-03-22)
 
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -468,6 +468,13 @@ module Watir
     end
 
     #
+    # Delegates script execution to browser or frame.
+    #
+    def execute_script(script, *args)
+      @query_scope.execute_script(script, *args)
+    end
+
+    #
     # Returns true if a previously located element is no longer attached to DOM.
     #
     # @return [Boolean]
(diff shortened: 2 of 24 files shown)


lib/watir/elements/element.rb
```diff
@@ -468,6 +468,13 @@ def browser
     end
 
     #
+    # Delegates script execution to browser or frame.
+    #
+    def execute_script(script, *args)
```
p0deje, line 473: I don't follow, why do we introduce `Element#execute_script`? Does it make any sense?

CHANGES.md
```diff
@@ -1,6 +1,11 @@
 ### 6.2.2 (Unreleased)
 
 * Allow locating elements by attribute presence/absence (#345)
+* Element#flash configurable by color, number, and delay (thanks Paul3816547290)
+* Implement Select#text like Select#value (thanks Arik Jones)
+* Optimize Select#selected_options, Select#value, and Select#text with javascript (thanks Andrei Botalov)
+* Support #execute_script from inside frame context
+* Deprecate PhantomJS Support
```
p0deje, line 8: We're not really deprecating, just dropping from specs :)

p0deje (approved):
I've left 2 comments, but overall have good to release!

lib/watir/elements/element.rb
titusfortner: Right now we support `Browser#execute_script` and `IFrame#execute_script` and require the user to pick the one they want to use. 

Unfortunately this gets complicated with this example: `browser.div(id: 'a').iframe(id: 'b').div(id: 'c').select(id: 'd').selected_options` 

Now that we use `#execute_script` to get `#selected_options`, the call needs to be made by the `IFrame`, not the `Browser`, and the resulting elements need to have the frame included in their scoping. The only way I know how to do this is to use the `@query_scope` variable. So in the above example the `Select` instance calls `@query_scope.execute_script`, which means the div element at id: 'c' needs to respond to `#execute_script` by passing it up to its `@query_scope`, and that will happen until the query scope is either an instance of `Browser` or `IFrame`.

titusfortner:
merged

PR #568, opened 2017-06-21
6.4

titusfortner:
I'm going to merge this to master and bundle as a release candidate for people to try out before an official release. Here is the diff to evaluate all of the performance changes so we can have a discussion about any issues or changes we want to make before an official release.

@p0deje do I need to do something to fix the doctest issues? I'm seeing a number of`Net::ReadTimeout`s.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+### 6.4.0.rc1 (2017-06-21)
+
+* Significant performance updates
+
 ### 6.3.0 (2017-06-20)
 
 * Allow locating elements by attribute presence/absence (#345)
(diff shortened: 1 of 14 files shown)


titusfortner:
Looks like we're going to need to update the gemspec to require Selenium 3.4.1 since that's where the new error messages.
```
  * Renamed ElementClickIntercepted to ElementClickInterceptedError.
  * Renamed ElementNotInteractable to ElementNotInteractableError.
```

p0deje:
> @p0deje do I need to do something to fix the doctest issues? I'm seeing a number ofNet::ReadTimeouts.

I need to fix those *sigh*

PR #575, opened 2017-07-13
allow collection elements to be relocated

titusfortner:
Instead of just doing `#to_subtype`, this stores the index of the tag name so it can be relocated.
The alternative is to figure out when to throw an exception when the only locator is `@element`.


p0deje (approved):
Heh, nice catch.

titusfortner:
merged 4e3b54578133acf16c29640df21b0df9d4bd8611

Issue #578, opened 2017-07-18
Ability to match on multiple/absence of class


titusfortner:
This comes from the discussion on this issue:
https://github.com/cheezy/page-object/issues/429

Theoretically we could implement something like this with code like
```
//*[contains(concat(' ', @class, ' '), ' yes ') and not(contains(concat(' ', @class, ' '), ' no '))]
```

But this could cause issues with backward compatibility because right now Watir lets you select with more than one class (which Selenium's `:class` selector does not) based on XPath's string comparison.

Right now this:
```html
    <div class="a b c">Match me!</div>
```

gets matched by:
```ruby
browser.div(class: 'a')
browser.div(class: 'b')
browser.div(class: 'c')
browser.div(class: 'a b')
browser.div(class: 'b c')
browser.div(class: 'a b c')
```

but does not get matched by:
```ruby
browser.div(class: 'c b a')
browser.div(class: 'c b')
browser.div(class: 'c a')
browser.div(class: 'b a')
browser.div(class: 'a c')
```

I mostly think this is bad, but I also think it would break a lot of people's stuff if we changed it to raise exception if spaces are found. I guess this is the difference between treating "class" in html as separate objects vs an ordered string.

but what if we implemented a `:classes` locator (raise exception if both `class` & `classes` used at the same time)

```ruby
# 1
browser.element(classes: {true => ['a', 'b'], false => ['d']})
# 2
browser.element(classes: ['a', 'b', 'd' => false])
# 3
browser.element(classes: ['a', 'b', '!d'])
# 4
browser.element(classes: "a b !d")
 ```

Anyone have opinions? @chrismcmahon?

chrismcmahon:
I like the :classes locator idea. I find myself working with elements sharing a mix of classes quite often.

jkotests:
I was under the impression, in terms of applying CSS, that the order of class names in a class attribute value is irrelevant (https://stackoverflow.com/a/29695391/1200545). Personally, I would prefer to make the :class locator order insensitive. To me, all of the current not matching examples should match. It's always strange to have to explain that
`browser.div(class: 'a b')` will be different than `browser.div(class: 'b a')` when applying the CSS doesn't care.

While there could be backwards incompatibility, I would hope it does not affect many people. At least, I would hope that people are not relying on an order that a developer could change at any time. For the scenarios that are impacted, wouldn't they be able to switch to a regular expression and get the same behaviour? 

[...]

PR #579, opened 2017-07-19
implement classes locator

titusfortner:
Here's an implementation for #578 
I haven't spent a lot of time in the SelectorBuilder code for XPath, so let me know if I'm putting things in wrong places.

```ruby
browser.div(classes: ["a"])
browser.div(classes: ["c", "a"])
browser.div(classes: ["a", "!c", "b"])
```

### DIFF
--- a/lib/watir/locators/element/selector_builder.rb
+++ b/lib/watir/locators/element/selector_builder.rb
@@ -2,7 +2,7 @@ module Watir
   module Locators
     class Element
       class SelectorBuilder
-        VALID_WHATS = [String, Regexp, TrueClass, FalseClass].freeze
+        VALID_WHATS = [Array, String, Regexp, TrueClass, FalseClass].freeze
         WILDCARD_ATTRIBUTE = /^(aria|data)_(.+)$/
 
         def initialize(query_scope, selector, valid_attributes)
@@ -35,6 +35,9 @@ module Watir
               raise TypeError, "expected TrueClass or FalseClass, got #{what.inspect}:#{what.class}"
             end
           else
+            if what.is_a?(Array) && how != :class
+              raise TypeError, "Only :class locator can have a value of an Array"
+            end
             unless VALID_WHATS.any? { |t| what.is_a? t }
               raise TypeError, "expected one of #{VALID_WHATS.inspect}, got #{what.inspect}:#{what.class}"
             end
--- a/lib/watir/locators/element/selector_builder/xpath.rb
+++ b/lib/watir/locators/element/selector_builder/xpath.rb
@@ -26,7 +26,9 @@ module Watir
           # @todo Get rid of building
           def attribute_expression(building, selectors)
             f = selectors.map do |key, val|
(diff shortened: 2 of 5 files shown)


titusfortner:
@jkotests & @chrismcmahon 
While I'm waiting for Alex on the correctness of this implementation, do you have any feedback on the desirability of this syntax?

jkotests:
I'm good with specifying an Array for the value and the "!" for exclusions.

Like I said before, I'd prefer to stick with :class instead. The fact that you will be passing an Array vs a String would make the codes intentions clear. This would also be more similar to the absence/presence functionality, which has a behaviour based on the type of value (ie Boolean). To me, the key is meant to reflect the attribute we are matching, which doesn't change based on how many classes we want to match.

p0deje:
Seems good to me, though I agree with @jkotests that we should use `:class` instead. If we stick to it, this locator can be generalized to support arrays for any other locator (though it makes most sense for classes).

P.S. Sorry for getting back that late - I was on vacation with limited access to the internet!

titusfortner:
Yeah, `:class` makes sense, it's just much harder to implement based on how we're currently doing `SelectorBuilder::XPath#attribute_expression`. Let me see how not-ugly I can make it. :)

titusfortner:
Ok, that wasn't as hard as I thought it would be. @p0deje if you don't have any complaints with this implementation, I'll do a release with it in the next day or two for Chris.

jkotests:
I did not realize the that the XPath builder already supported Array values. It looks like it would only be available to direct calls - ie not available to users, which get a TypeError. The builder will do an "or", where as the class attribute will do an "and".

For example, given:

~~~~~
<input data-value="a b"></div>
~~~~~

With the proposed changes:

~~~~~ruby
b.text_field(data_value: ['a']).exists? #=> false
b.text_field(data_value: ['b']).exists? #=> false
b.text_field(data_value: ['a', 'b']).exists? #=> false
b.text_field(data_value: ['a b']).exists? #=> true
b.text_field(data_value: ['invalid', 'a b']).exists? #=> true
~~~~~

Do you think this inconsistency is okay? It would be ideal if we can be consistent across attributes, but I'm not sure what API to propose. Maybe we should constrain user-specified Array values to the :class locator for now?

titusfortner:
It previously supported Array values to the extent that it checks aliases for Button types, etc; so implementation specific. Honestly, classes are the only things that I think should be treated as space-separated objects. Everything else I believe deserves to be treated like a string.

Ah, but I changed the list to allow Array to be a valid type that can be passed in directly by the user without throwing an exception, even when it isn't just in the context of the class. Good catch. Let me figure out where to throw that exception.

p0deje (approved):

titusfortner:
@jkotests & @p0deje I added another commit that will raise an exception if an Array is passed in as a value for a parameter that isn't `:class`. Let me know if we need to rethink this more.

jkotests:
Looks good to me.

PR #580, opened 2017-07-19
use default Travis Chrome

titusfortner:
Alex, I see your comment about Chrome & shared memory on Travis, but I don't really understand it. When I ran on the container everything passed, and they seem to be about the same speed between the runs, though I didn't do a thorough comparison.

Regardless, can we use default Travis package, or do we need the current sudo libs?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,7 @@ rvm:
   - 2.4.1
 addons:
   firefox: latest
-  apt:
-    packages:
-      - unzip
-      - libxss1
-      - google-chrome-stable
+  chrome: stable
 cache: bundler
 notifications:
   recipients:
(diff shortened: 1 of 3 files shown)


p0deje:
1. The problem with shared memory is that container has it limited, so Chrome might crash. If in never images on Travis everything passes in sudo-less environment, we can use it.
2. Please don't disable gems on Travis because it safeguards us from dependency collisions (we've already faced them at least once and Travis builds were failing).

For everything else, good to merge.

titusfortner:
@p0deje thanks. Is this better?
Also, if Chrome crashes due to shared memory issue, isn't that a Travis bug that we should get them to fix if they are trying to get everyone to switch over? I guess it doesn't really matter.

titusfortner:
disregard, I'm trying something new.

PR #581, opened 2017-07-20
run watir specs on Sauce Labs

titusfortner:
@p0deje - I'm looking for feedback on how I implemented this in Watirspec. This is working but I'm not sure I'm approaching it the right way.

The heavy lifting for Sauce integration is done with the Saucer gem that is still very much in a beta mode. (I need to see if knapsack works better because I'm struggling with how parallel split test is doing things right now).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -152,4 +152,27 @@ namespace :spec do
       Rake::Task[:spec].execute
     end
   end
+
+  desc "Run on Saucelabs"
+  task :sauce do
+    require 'saucer'
+    ENV['USE_SAUCE'] ||= 'true'
+
+    # Note - This needs to be cleaned up
+    ENV['WATIR_BROWSER'] ||= 'remote'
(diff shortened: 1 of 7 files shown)


p0deje:
1. Why do you need to change hooks from `before(:all)` to `before(:each)`?
2. Why do you need to run specs in parallel? Isn't 1 process enough for us?

titusfortner:
I'm not sure; for some reason that spec wasn't working with this code and I changed it to `before(:each)` and now it is working. I didn't figure out the cause, just changed the code. I could look into it more.

I'm not looking to replace what we have now. Partly this is an exercise to make sure we can run specs on Sauce and get a working example with the saucer gem; though, it would also be nice if it were simple to run tests on any platform/browser.

titusfortner:
This is now out of date and needs reworking, will resubmit later.

PR #582, opened 2017-07-28
Spec updates

titusfortner:
1. Stop testing legacy Firefox
2. Use Tech preview for Safari tests since tests actually pass using it
3. Test using the latest released Standalone Server
4. I copied Alex's approach to Selenium Travis here to minimize the number of things that get run. I think this is the right cross section.
5. Does anyone read the emails from Travis results? If so, I'll add back the notifications (@jarmo?)
6. I moved the server startup code into lib/watirspec because `watirspec_helper` was cluttered and because I was going to create a rake task that started the server... Then I decided not to do the rake task. But I think that code is better off in lib than in the spec directory.

Also, any ideas on the failing doctest for `Selenium::WebDriver::Error::UnknownError: unknown error: failed to change window state to maximized, current state is normal`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,31 +3,65 @@
 # use host's /dev/shm.
 sudo: required
 dist: trusty
-rvm:
-  - 2.2.7
-  - 2.3.4
-  - 2.4.1
-addons:
-  firefox: latest
-  apt:
-    packages:
-      - unzip
-      - libxss1
-      - google-chrome-stable
 cache: bundler
 notifications:
-  recipients:
-    - p0deje@gmail.com
-    - jarmo.p@gmail.com
-    - titusfortner@gmail.com
   slack:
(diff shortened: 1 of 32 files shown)


p0deje (approved):
> 1. Stop testing legacy Firefox.

👍 

> 2. Use Tech preview for Safari tests since tests actually pass using it

👍 

> 3. Test using the latest released Standalone Server

👍 

> 4. I copied Alex's approach to Selenium Travis here to minimize the number of things that get run. I think this is the right cross section.

👍 

> 5. Does anyone read the emails from Travis results? If so, I'll add back the notifications (@jarmo?)

I'm fine with not reading them :)

> 6. I moved the server startup code into lib/watirspec because watirspec_helper was cluttered and because I was going to create a rake task that started the server... Then I decided not to do the rake task. But I think that code is better off in lib than in the spec directory.

👍 

> Also, any ideas on the failing doctest for Selenium::WebDriver::Error::UnknownError: unknown error: failed to change window state to maximized, current state is normal?

🤷‍♂️ Does it happen in master?

jarmo:
I'm okay not receiving any notifications for this project anymore. Thanks for confirming!

titusfortner:
This passed 9 days ago: https://travis-ci.org/watir/watir/builds/255369312
I just reran the doctest and it is failing. Same code, so...

Oh, just checked, chromedriver was updated between passing/failing results, so this is likely Google's fault. Though, not sure why it would just be an issue in docs and not in specs. I'll try to run on my linux box this weekend to see if I can reproduce.

Issue #436, opened 2016-08-06
Watir::Capabilities class


titusfortner:
Would it make sense to have Watir wrap the Selenium Remote::Capabilities class?

As I'm writing up documentation on how to set marionette: false for the upcoming Watir 6.0, this is what needs to be done:

```
caps = Selenium::WebDriver::Remote::Capabilities.firefox(marionette: false)
Selenium::WebDriver::Firefox::Binary.path = "/path/to/desired/firefox"
Watir::Browser.new desired_capabilities: caps
```

It seems like it would be much nicer to be able to do:

```
Watir::Browser.new(marionette: false, firefox_binary: "/path/to/binary")
```

But maybe we need to figure out the right approach on the Selenium side first, because I hate how we combine allowing both first class parameters & desired_capabilities parameters.

ambirag:
@titusfortner 
just to add to what you are trying to do, currently we also need another way to pass Firefox profile, args and prefs for the new moz:firefoxOptions of marionette , if anyone is using selenium webdriver 3.\* series. afaik, this is not supported in watir, please let me know otherwise.

titusfortner:
@ambirag - it is supported in Watir. For Firefox with geckodriver, Selenium 3 converts `:firefox_options` to `"moz:firefoxOptions"` before creating the new session. Let me know if you are finding issues.

ambirag:
@titusfortner thanks, let me try and get back to you

ambirag:
if i pass the "profile" in profile = Selenium::WebDriver::Firefox::Profile.new
    profile['accept_untrusted_certs']=true
    profile['assume_untrusted_certificate_issuer']=false
in 
capabilities = Selenium::WebDriver::Remote::Capabilities.firefox(marionette: true,firefox_options: {profile: "profile"}) , i get "Failed to unzip profile"
Is there any example how to create base64 encoded profile directory?

ambirag:
@titusfortner

titusfortner:
This was implemented with #583

PR #583, opened 2017-07-28
implement Watir::Capabilities

titusfortner:
Whew. Almost a year to actually implement #436 

The important points:

1. Remove the need to create Selenium classes to initialize `Watir::Browser`; this means Watir users create hashes for all of the things and Watir plugs them into the correct objects for Selenium to use.
2. Remove the need to specify ":remote" as a browser option, just pass in browser you want, along with the server url


lib/watir/capabilities.rb
```diff
@@ -0,0 +1,85 @@
+module Watir
+  class Capabilities
```
p0deje, line 2: It would be great to have some docs and examples for this.

lib/watirspec/implementation.rb
```diff
@@ -8,6 +8,11 @@ def initialize
       @guard_proc = nil
     end
 
+    def initialize_copy(orig)
+      super
+      @browser_args = Marshal.load( Marshal.dump(@browser_args))
```
p0deje, line 13: Unnecessary space after `(`.

p0deje: Why you need to marshal? Why not just `#clone` or `#dup`.

p0deje (commented):
Overall it's good. Just keep in mind we planned to rework capabilities in the future releases of selenium-webdriver.

p0deje (approved):

titusfortner:
I held off on working on this in Watir because I thought selenium updates should be done first. Except I got burned out on the selenium work. I recognize that this will keep the Watir and selenium versions more tightly coupled for a while.

lib/watirspec/implementation.rb
titusfortner: This was a pain to figure out. A clone or dup is a different object referencing the same objects inside the browser_args array. I was really struggling especially when I was trying to do this in the specs themselves. Now that it is implemented inside the class with `#initialize_copy`, I see that I can update it to be: `@browser_args = browser_args.map(&:dup)`. Thanks for making me check again, this is much better than Marshal.

PR #584, opened 2017-07-29
Select List updates

titusfortner:
A few performance updates for Selects & Options (24% fewer wire calls in `select_list_spec`. This should be more significant when working with a large number of options....

Though, now that I think about it, when using Regex it could be slower... I'll create a separate PR to address that.

This also removes the direct Selenium `#find_elements` calls from the Select class, which makes this much cleaner.

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -10,7 +10,7 @@ module Watir
       raise Error, "you can only clear multi-selects" unless multiple?
 
       options.each do |o|
-        click_option(o) if o.selected?
+        o.click if o.selected?
       end
     end
 
@@ -21,7 +21,7 @@ module Watir
     #
 
     def options(*)
-      element_call(:wait_for_exists) { super }
+      super
     end
 
     #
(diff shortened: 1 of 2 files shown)


lib/watir/elements/select.rb
```diff
@@ -75,21 +71,15 @@ def select_value(str_or_rx)
     #
 
     def selected?(str_or_rx)
-      match_found = false
+      by_text = options(text: str_or_rx)
+      return true if by_text.find { |e| e.selected? }
```
p0deje: `by_text.find(&:selected?)`

lib/watir/elements/select.rb
```diff
@@ -75,21 +71,15 @@ def select_value(str_or_rx)
...
-        @element.find_elements(:tag_name, 'option').each do |e|
-          matched = str_or_rx === e.text || str_or_rx === e.attribute(:label)
-          if matched
-            return true if e.selected?
-            match_found = true
-          end
-        end
-      end
+      by_label = options(label: str_or_rx)
+      return true if by_label.find { |e| e.selected? }
```
p0deje: `by_label(&:selected?)`

p0deje (approved):

PR #585, opened 2017-07-30
Select all

titusfortner:
1. Removes a few reasons to use `ElementCollection#to_a`

2. Implements `Select#select_all` and deprecates, but does not remove functionality for selecting multiple options with `Select#select`

The issue here is performance. If we only want to click one thing, we can search for it directly, instead of iterating over a collection, which is very slow, especially for large select lists.
Note that the performance will not actually improve until we change the implementation of `#select`, but I thought we should provide a deprecation warning for some set amount of time before we pull the trigger on the change.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -40,6 +40,22 @@ module Watir
     alias_method :size, :length
 
     #
+    # Returns true if no elements are found
+    #
+    # @example
+    #   browser.select_list(name: "new_user_languages").options(class: 'not_here').empty?
+    #   #=> true
+    #
+    # @example
+    #   browser.select_list(name: "new_user_languages").options(id: 'danish').empty?
+    #   #=> false
+    #
+
+    def empty?
+      length == 0
+    end
+
+    #
     # Get the element at the given index.
     #
     # Also note that because of Watir's lazy loading, this will return an Element
@@ -96,6 +112,7 @@ module Watir
             end
           end
     end
+    alias_method :locate, :to_a
 
     #
     # Returns true if two element collections are equal.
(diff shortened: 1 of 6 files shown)


lib/watir/element_collection.rb
```diff
@@ -40,6 +40,39 @@ def length
...
+    # @example
+    #   browser.select_list(name: "new_user_languages").options(class: 'not_here').exist?
+    #   #=> false
+    #
+    # @example
+    #   browser.select_list(name: "new_user_languages").options(id: 'danish').exist?
+    #   #=> true
+    #
+
+    def exist?
```
p0deje: Is there any reason to prefer this method name over `#any?`?

p0deje (approved):

lib/watir/element_collection.rb
titusfortner: Wouldn't that override enumerable method?

p0deje: Isn't it the method that you want? Some method that is the opposite of `#empty?`.

If class already has `#any?` - let's just use it instead of adding another one.

titusfortner: Hmm, not near an irb, doesn't any require a block?

titusfortner: Hmm, not near an irb, doesn't any require a block?

p0deje: No.

```
irb(main):003:0> [1].empty?
=> false
irb(main):004:0> [1].any?
=> true
```

titusfortner: That's awesome. :-) sure, we don't need exist. :-)

PR #587, opened 2017-07-31
implement Watir::Executor

titusfortner:
I believe this is closer to the implementation we discussed a year ago for the waiting strategies.

This code is just providing parity with existing code. I actually like pulling all of this out of the Element class since it removes so much clutter.

There are several additional features I'm thinking about, but for now I'd like to consider this a private api while we figure out what things we want to open up for users to do. 

Looking for input on syntax, naming, implementation, anything. Thanks!

-------------------------

Future features I'm imagining:
```
Watir.executor.add_hook(:before_click, ->(o) { puts "I'm so excited to click #{o.inspect}!" }
Watir.executor.add_hook(:after_text, ->(o) { puts "Retrieving text from #{o.inspect} made me happy" }
Watir.executor.add_hook(:rescue_submit, ->(o) { puts "Uh Oh: #{o.browser.span(class: 'error').text}" }
```
Changing wait strategies can currently be done by subclassing
Something like these could be fun:
```
Watir.executor.change_strategy(:element, :set, :exist, timeout: 4, include: [:text_field, :textarea])
Watir.executor.change_strategy(:alert, :ok, :none, timeout: 4)
```
We could also do these as arrays so you can take more than one action at a time, essentially re-implementing after_hooks in an extensible way.

These things are all likely low priority, though

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -14,6 +14,7 @@ require 'watir/cookies'
 require 'watir/browser'
 require 'watir/screenshot'
 require 'watir/after_hooks'
+require 'watir/executor'
 
 module Watir
 
@@ -21,7 +22,7 @@ module Watir
 
   class << self
 
-    attr_writer :relaxed_locate, :always_locate, :default_timeout, :prefer_css, :locator_namespace
+    attr_writer :always_locate, :default_timeout, :prefer_css, :locator_namespace
 
     #
     # Whether or not Watir should wait for an element to be found or present
@@ -31,6 +32,11 @@ module Watir
 
     def relaxed_locate?
       @relaxed_locate
+    end
+
+    def relaxed_locate=(val)
+      warn "`Watir#relaxed_locate=` is deprecated, use `Watir#executor#no_wait!` instead"
+      executor.no_waits! if val
     end
 
     #
@@ -82,6 +88,14 @@ require the watir_css gem - https://github.com/watir/watir_css
 
     def locator_namespace
       @locator_namespace ||= Watir::Locators
+    end
+
+    #
+    # Stores the Watir executor
+    #
+
+    def executor
+      @executor ||= Watir::Executor.new
     end
 
     #
--- a/lib/watir/alert.rb
+++ b/lib/watir/alert.rb
@@ -3,6 +3,8 @@ module Watir
     include EventuallyPresent
     include Waitable
 
+    attr_reader :browser
+
     def initialize(browser)
       @browser = browser
       @alert = nil
@@ -19,8 +21,7 @@ module Watir
     #
 
     def text
-      wait_for_exists
-      @alert.text
+      Watir.executor.go(self) { @alert.text }
     end
 
     #
@@ -33,9 +34,7 @@ module Watir
     #
 
     def ok
-      wait_for_exists
-      @alert.accept
-      @browser.after_hooks.run
+      Watir.executor.go(self) { @alert.accept }
     end
 
     #
@@ -48,9 +47,7 @@ module Watir
     #
 
     def close
-      wait_for_exists
-      @alert.dismiss
-      @browser.after_hooks.run
+      Watir.executor.go(self) { @alert.dismiss }
     end
 
     #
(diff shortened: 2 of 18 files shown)


lib/watir.rb
```diff
@@ -33,6 +34,11 @@ def relaxed_locate?
       @relaxed_locate
     end
 
+    def relaxed_locate=(val)
+      warn "`Watir#relaxed_locate=` is deprecated, use `Watir#executor#no_wait!` instead"
```
p0deje, line 38: no_wait => no_waits

p0deje:
This is interesting, let me think about that and get back later.

titusfortner:
Let me know your ideas. I don't necessarily love this implementation, but I keep finding ways to make `#element_call` more messy, so I figured it might make sense to do something more like what we talked about doing a year and a half ago.

p0deje (commented):
I quite like the idea. So far, it looks like a refactoring of private API for me because I don't see how I could have used it for now, but that might be the foundation of the future extensibility.

I vote for merging this with 1 change - let's extract alert/element/window strategies to their own classes. Executor is messy now, so that will make it more readable and maybe it will show how we can organize API for extensibility in the future.

Donavan:
I have a gem called [Captain Hook](https://github.com/Donavan/cpt_hook) that adds before/after hooks to arbitrary objects.  I built it for the purpose of DRYing up my automation code.  I've tied it in at the PageObject layer but there's no reason it couldn't be tied into Watir.

I have a [gist](https://gist.github.com/Donavan/502e103d08b821741c97c69d527f2819) up of the PageObject extension if anyone wants to take a look.  I've been testing it in a large project for the past few weeks before actually releasing the extension.  It works be replacing the "_element" method PageObject adds to return a Watir element wrapped by a Captain Hook "Hookable" wrapper (A small SimpleDelegator wrapper).

With it I can declare a set of hooks like so:

```ruby
WFA_HOOKS ||= CptHook.define_hooks do
  after(:click).call(:wait_for_ajax)
  after(:set).call(:wait_for_ajax)
  after(:value=).call(:wait_for_ajax)
  after(:check).call(:wait_for_ajax)
  after(:uncheck).call(:wait_for_ajax)
end
```

Then use them when I declare the elements:

```ruby
button_hooked(:cancel_dialog, data_bind: /cancelDialog/, hooks: WFO_HOOKS)
```

Hooks can be combined for composition:

```ruby
  SEARCH_HOOKS ||= define_hooks do
    before(:value=).call(:ensure_search_element_visible)
  end
  text_field_hooked(:search, id: 'cmboJobReference', hooks: SEARCH_HOOKS.merge(WFA_HOOKS))
```

The DSL allows for more complicated hooks where parameters are passed as well as using procs/lambdas.  It also has the notion of a "context" for the method call, allowing the hook to make calls on other objects.

```ruby
  DASHBOARD_MENU_HOOKS ||= define_hooks do
    before(:click).call(:ensure_visible).with(:dashboard).using(:page)
  end
```

titusfortner:
@Donavan thanks for the link to the code. I saw you post it to PO gem, and I almost mentioned our intention to add Watir::Executor to Watir 7.

Of course, this PR is way out of date, and was mostly a proof of concept anyway. I need to redo it with the latest logic and throw down some example code.

I like the syntax of your usage, though I'm not sure what it would take to implement it using Watir, or how that would compare to the approach I'm investigating. I will also admit that I have a personal bias against the delegator class because I find it harder to debug through. It's another reason I'm not a huge fan of the PO gem itself, it over-complicates a lot of otherwise straightforward things.

I'll dig into this more when we get closer to Watir 7. I have a lot of work in the `Watir::Capabilities` class before I can get to it, though. :-/

Donavan:
With PageObject I'm making their element wrappers subclasses of CptHook::Hookable.  For Watir I'd just skip the whole SimpleDelegator base.  Maybe turn the core functionality into a module that could just be included into Watir::Element.

akostadinov:
Hi, I'm wondering if this Executor can help with error handling as well. Sometimes operations over elements can raise strange errors because of browser/page non-deterministic behavior depending on browser window size or just performance. e.g. clicking an element results in `element is not clickable at point` error.

I have implemented some begin/rescue logic around `click` and other operations to scroll element to the middle, close menus, retry, etc. in such occasions. But would be interesting if custom error handling can be added through this Executor.

titusfortner:
Since this is now 250 commits behind master, I'm closing this and tracking it with #835

PR #593, opened 2017-08-01
implement direct support for headless

titusfortner:
Not sure we *need to run headless tests on Travis, but it is interesting to see what is/isn't supported so far. (pretty much alert/window handling and javascript clicks)

`Watir::Browser.new :chrome, headless: true`
or
`Watir::Browser.new :chrome, headless: true, url: "#{remote_server}/wd/hub"`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,36 +32,39 @@ _browsers:
 
 matrix:
   include:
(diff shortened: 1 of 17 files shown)


p0deje (approved):
Cool.

PR #596, opened 2017-08-01
support for finding multiple classes with regular expressions

titusfortner:
This code is kind of ugly. Do we even care about this feature? Someone pinged me on Slack saying they thought they needed to convert everything to an Array, and their single regex broke. I had him remove the array and he was cool with it. But thought it would be interesting to get it to work.

I'm ok with not doing this if we don't care about the support for it.

@chrismcmahon, would you use it?


p0deje:
I don't really think it makes sense. You can have a single regexp containing OR/AND conditions, so what's the point of having array?

titusfortner:
yup that totally makes sense.

PR #598, opened 2017-08-02
Allow user-specified :type when locating input buttons

jkotests:
When locating input buttons, the user-specified :type is ignored. For example, in the page:

~~~~~~~~html
<input type="reset">
<input type="submit">
~~~~~~~~

You get the wrong element if you locate by type:

~~~~~~~~ruby
browser.button(type: "submit").html
#=> "<input type="reset">"

browser.button(type: false).html
#=> "<input type="reset">"
~~~~~~~~

With these changes:

1. Match the user-specified type when specified
2.  Ignore input tags when {type: false} (since that would mean they are text fields)
3. Continue to select only valid button types when {type: true} or not specified

### DIFF
--- a/spec/watirspec/elements/button_spec.rb
+++ b/spec/watirspec/elements/button_spec.rb
@@ -75,6 +75,23 @@ describe "Button" do
       expect(browser.button(xpath: "//input[@type='button']")).to exist
     end
 
+    it "matches the specific type when locating by type" do
+      aggregate_failures do
+        expect(browser.button(type: "button").type).to eq "button"
+        expect(browser.button(type: "reset").type).to eq "reset"
+        expect(browser.button(type: "submit").type).to eq "submit"
+        expect(browser.button(type: "image").type).to eq "image"
+      end
+    end
+    
+    it "matches valid input types when type is boolean" do
+      aggregate_failures do
+        expect(browser.buttons(type: false).map(&:tag_name)).to all eq("button")
+        
+        input_buttons = browser.buttons(type: true).select { |e| e.tag_name == "input" }
+        expect(input_buttons.map(&:type).uniq).to match_array(Watir::Button::VALID_TYPES)
+      end
+    end
 
     it "raises TypeError when 'what' argument is invalid" do
       expect { browser.button(id: 3.14).exists? }.to raise_error(TypeError)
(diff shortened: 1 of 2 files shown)


titusfortner:
I'm not following what the use case is for passing in true/false value here.

spec/watirspec/elements/button_spec.rb
```diff
@@ -75,6 +75,23 @@
       expect(browser.button(xpath: "//input[@type='button']")).to exist
     end
 
+    it "matches the specific type when locating by type" do
+      aggregate_failures do
```
titusfortner, line 79: This is a nifty feature, but I think since we don't do this anywhere else in our specs, we shouldn't do it here.

p0deje:
Why not use it like this?

```ruby
browser.input(type: "submit")
browser.input(type: "reset")
```

jkotests:
@p0deje, for that specific example, yes, your suggestion would work. However, there will still be problems. For example, the original time I noticed the problem, I had a page with a mix of inputs and buttons:

~~~~~~~~html
<input type="button" id="Search">
<button type="submit">
~~~~~~~~

I tried to locate the button tag using the type attribute since I knew it was the only submit button on the page:

~~~~~~~~ruby
browser.button(type: 'submit').html
#=> "<input type=\"button\" id=\"Search\">"

browser.buttons(type: 'submit').count
#=> 2
~~~~~~~~

jkotests:
@titusfortner, I don't have a use case for true/false. I just noticed the issue when trying to fix my original problem, so I figured I'd fix it while I was there. The type attribute is optional for button tags, so in theory, someone could want to differentiate buttons with/without the attribute.

I can take that part back out if you don't think it's worth maintaining.

spec/watirspec/elements/button_spec.rb
jkotests: I'm all for consistency, but if that's the only opposition, would it be better to start switching the existing specs to use aggregate_failures? It just seems weird that if we have a number of independent checks that we have to work through one failure at a time rather than knowing all of the failures.

p0deje:
I see. Well, let's add `type: String` to button then 👍

p0deje:
I agree that it doesn't make sense to have `type: Boolean` though.

titusfortner:
Ah, ok, I see what you are doing. This keeps it feature compatible with all the other forms of attributes. Right. I'm ok with this code then, just to keep it all the same, even if it doesn't always make sense to use it.

spec/watirspec/elements/button_spec.rb
titusfortner: For how I work with Watir specs, it is just unnecessary. I fix things one at a time and rerun failing tests until they pass. Plus I'd rather see fewer blocks in our specs rather than adding a bunch more throughout all the files. :)

titusfortner:
I merged this to get it out with 6.6 release today. Thanks!

Issue #600, opened 2017-08-03
Attribute Location


titusfortner:
So, while looking at #207, we have a big issue with how attribute location is working. I guess this means that no one really uses it? :)

I think this is a result of #261. We need to get the generator working again and be able to update the code. I don't really understand it well enough to know what it is doing. :-/

For instance, with this html:
```html
 <div tabindex="1">Foo</div>
```
To select the element, the xpath needs to contain "tabindex", which is precisely the way that the current code does not let you use.

```ruby
# underscore (selector builder converts to dash)
browser.div(tab_index: "1")).exist?
# 2017-08-03 00:07:02 DEBUG Watir Converted {:tab_index=>"2"} to [:xpath, ".//div[@tab-index='2']"]

# downcase
browser.div(tabindex: "1").exist?
# Watir::Exception::MissingWayOfFindingObjectException: invalid attribute: :tabindex

# camelcase
browser.div(tabIndex: "1").exist?
# 2017-08-03 00:07:02 DEBUG Watir Converted {:tabIndex=>"2"} to [:xpath, ".//div[@tabIndex='2']"]
```

Just aliasing the method between `tabindex` & `tab_index` will still raise the exception for invalid attribute.

This looks like it is an issue with every single camel case attribute in `html_elements.rb`
Which I guess means we aren't testing any of these attribute location strategies right now?

p0deje:
#207 and #28 is about attribute methods:

```ruby
browser.div.tab_index
```

Nobody really looked into selectors, so I'm not sure what works and doesn't work there.

Tricky part here is that you have to translate some of them to dashed and some to camelCase. For example, `tab_index` should be `tabindex`, while `http_equiv` should be `http-equiv`. There are many cases so #601 should take into account.

titusfortner:
#207 talked about both... But this is a good question. Do we need underscores for any of these? Doesn't `#itemtype` make more sense than `#item_type` if the actual attribute that someone sees in the html is the latter?

titusfortner:
fixed!

Issue #28, opened 2010-08-22
inconsistent: IDL attribute vs ruby method vs content attribute


jarib:
We're currently not translating between these correctly. 

e.g.:

<pre>
  def accept_charset
    @element.attribute("accept_charset")
  end
</pre>

should really be:

<pre>
  def accept_charset
    @element.attribute("accept-charset")
  end
</pre>

There's way too many of these to maintain by hand, so hopefully we can find a way to extract it from the HTML spec.

titusfortner:
@p0deje did #261 fix this issue?

p0deje:
@titusfortner It would still be great to allow to handle content attributes as well, but I believe IDL is fine in 99% of cases and watir-webdriver has been using them for a long time already, so I'm closing this.

PR #601, opened 2017-08-03
fix attribute location

titusfortner:
This addresses #600 

I don't think there is any reason to store the camel case version of the attribute in the attribute list.
Also, for the vast majority of items, we don't want underscores to be replaced with dashes. I might be missing some here, though. There are so many elements they are hard to parse.

Update:
This PR now does 

1. Allows *locating* an element by any of its valid content attribute names. Currently it converts underscores to dashes only for:
    * data_
    * aria_
    * http_
    * accept_

2. Returns the value of any valid content attribute when called as a method, with the exception of "tfoot" and "thead" which will return a `TableSection` instance. To obtain the attribute value of either of these two, call directly with `#attribute_value` instead.

3. For both locating and obtaining values, the snake case and down case are now interchangeable.


p0deje:
Have you read #28?

titusfortner:
Yes, but there are a *ton of attributes that come from IDL in camel case that get translated to snake case and then get converted to dash case in the xpath, when they need to be in downcase. In fact there are only a handful that should be dash case, and none of them seem very useful to me (no one is going to locate an element by "accept-charset"). This code supports many more options.

p0deje:
I'm not talking about selectors here, only attribute methods. Most of them are kept in camelCase because IDL attributes are camel cased. Since they are camel cased, how do you want to drop this from attributes list? Don't mix it up with content attributes, which we're not using in such methods.

Moving on, for selectors we actually need to use content attributes which are either dashed or downcased. The vast majority are downcased so we can apply this change. However, we also need to add exceptions for accept-charset, http-equiv and maybe some other attributes. I don't agree with `no one is going to locate an element by "accept-charset"` and I see no reason to break existing API. You can easily make it work with downcased version by default and exceptions for dashed attributes.

titusfortner:
Ok, so I added a line of code that will allow for using attribute methods both as:
`#tabindex` and `#tab_index`, etc

I guess I don't understand the inherent value of storing IDL attribute. If it isn't what is displayed in the DOM, it isn't what we want to use to locate an element, and isn't what we want to name the method we're calling, of what use is it in the Watir code? :)

There's already an exception in the code to allow for http-equiv. It'll be easy enough to add one for accept-charset as well. It's hard to see what all else would apply. Obviously it would be nice to have something that did this programatically.

titusfortner:
Well, that doesn't work, `#tfoot` - should it return the value of the attribute, or the TableSection element. :) #tryingagain

titusfortner:
@p0deje I've updated this PR which I think addresses the concerns you had. Let me know what you think of it. This is going to be much more robust.

p0deje:
> Allows locating an element by any of its valid content attribute names. Currently it converts underscores to dashes only for:

👍 

> Returns the value of any valid content attribute when called as a method, with the exception of "tfoot" and "thead" which will return a TableSection instance. To obtain the attribute value of either of these two, call directly with #attribute_value instead.

I don't think that adding 1 more attribute method which is not documented anywhere is a good idea.

The more I think about that, the more I'm convinced that we should change generator to try and figure out content attribute name for method out of IDL attribute. So, for example,

```ruby
  class TextArea < HTMLElement
    attribute(Integer, :max_length, :maxLength)
    attribute(Integer, :min_length, :minLength)
  end
```

will be changed to 

```ruby
  class TextArea < HTMLElement
    attribute(Integer, :maxlength, :maxLength)
    attribute(Integer, :minlength, :minLength)
  end
```

That should be more-less correct with a few exceptions. What do you think?

titusfortner:
> I don't think that adding 1 more attribute method which is not documented anywhere is a good idea.

What is getting added? This is just preventing things that have method names getting overridden by attributes. You can always call `#attribute_value`, but you can't do anything to get the method.

Oh, you're talking about adding both with and without underscore. Part of that is backward compatibility. Though, I think all of this is/has been broken for a while. So, yes, I'm all for removing the unnecessary underscores with generator. Are there some underscores we want to keep, or would we just convert them all?

p0deje:
> Are there some underscores we want to keep, or would we just convert them all?

I'd say this needs to be carefully reviewed and checked first :)

titusfortner:
boring ;)

p0deje:
I can do that, but again I'm not sure when, so if you are wiling to wait...

titusfortner:
Closing in favor of #615

Issue #547, opened 2017-04-08
JavaScript text entry


titusfortner:
Entering text via drivers tends to be slow. [(for reference)](http://stackoverflow.com/questions/25583641/set-value-of-input-instead-of-sendkeys-selenium-webdriver-nodejs)
Once we've verified a textfield or textarea can accept text, there isn't really a reason to type all of the text since that doesn't provide additional testing value.

I'm thinking perhaps `#set!` or `#set_text`, which would do the normal present/enabled/writeable checks, clear, perhaps type the first letter, use javascript to fill everything but the last character, then `send_keys` the final character. We can test to verify performance improvements for different drivers to see if it is worth doing.

This might solve the shadow dom issue people have had recently with React frameworks not recognizing text entry. I've also had issues in the past where the full text was not properly entered. Not sure if it is worth having an `#ensure_text` get added as well.

azohra:
Holy Jeeze is JS ever fast. Just gave the code below a try across a form of 10 text fields. I see huge benefit to this if we mix in your ideas above to keep interaction somewhat 'real'. 

```ruby 
    def do(v)
        set v
    end

# VS

    def do!(v)
        browser.execute_script("arguments[0].value = '#{v}';", self)
    end
```

``` ruby 
  it 'do! better than do' do
    site.home_page.visit
    watir = Benchmark.measure { site.home_page.input.do(input_region_fields) }
    site.home_page.visit
    js = Benchmark.measure { site.home_page.visit.input.do!(input_region_fields) }
    expect(js.real).to be < watir.real
  end
```

In Chromedriver
```ruby 
       user     system      total        real
   0.080000   0.020000   0.100000 (  1.636390)  #watir
   0.030000   0.000000   0.030000 (  0.499667) #js-direct

titusfortner:
is this chromedriver?

azohra:
@titusfortner - yes apologies for missing that. I am expanding the test to encompass more element types, will keep adding data here.

p0deje:
What is the problem with shadow DOM?

This sacrifices correctness (making sure all events like keyDown and keyUp are emitted) for performance, which is a bit too much IMO.

titusfortner:
http://stackoverflow.com/questions/42774205/text-entered-with-watir-keeps-disappearing

But I see that as a side benefit more than a major feature.

[...]

PR #603, opened 2017-08-04
implement bang methods with JavaScript

titusfortner:
I'm not a javascript person, but these seem to work.

Any feedbacks? @tk8817 you are doing something similar already? How does this compare?

This implements #547


lib/watir/elements/select.rb
```diff
@@ -161,6 +196,23 @@ def select_by(how, str_or_rx)
       select_matching(found)
     end
 
+    def select_by!(how, str_or_rx, number)
+      script = how == :text ? text_script : value_script
+
+      if str_or_rx.is_a? Regexp
+        js_rx = str_or_rx.inspect.sub('\\A','^').sub('\\Z','$').sub('\\z','$').sub(/^\//,'').sub(/\/[a-z]*$/,'').gsub(/\(\?#.+\)/, '').gsub(/\(\?-\w+:/,'(')
```
p0deje: wat

lib/watir/elements/select.rb
```diff
@@ -198,6 +250,33 @@ def matches_regexp?(how, element, exp)
         raise Error, "unknown how: #{how.inspect}"
       end
     end
+
+    def text_script
+      <<SCRIPT
```
p0deje: I propose to make these atoms or at least put into separate file and read on load. E.g. like https://github.com/p0deje/watir-dom-wait/blob/master/lib/watir/dom/elements/element.rb#L3

p0deje (commented):
*sigh* I guess this day has come - we're adding JavaScript to Watir. From this moment, we're stepping into the hell of JavaScript engines compatibility.

I'm not against it, so we can give it a shot. My other thought is that we might somehow reuse Selenium atoms which already handle a ton of cross-browser issues.

lib/watir/elements/select.rb
titusfortner: JavaScript complains with Ruby Regex. This is code I found on Stack Overflow. :)

p0deje: Don't get me wrong, but regexp engines in JavaScript and Ruby are much more different. For example, there is no support for negative lookbehind, POSIX expressions, etc. Also, there are different flags.

With that said, I'd rather not support it at all than saying "you could try, but there is no way we can guarantee it will work" :)

titusfortner: Oh yes, that's what I was reading. That being said, I feel like that's why we're doing these with bangs. They are a slimmed down, use at your own risk kind of thing. Hopefully others who find this interesting can improve on it. This is just the extent of my JS and/or Stack Overflow searching capabilities. :)

titusfortner:
Well, the day already came when we implemented `#selected_options` in JavaScript. That one isn't even a bang method. :)

This being said, I should fire up my windows machine to do a full set of runs on them to see where we stand, anyway. Mac & Linux are both happy with the limited specs I wrote. IE is probably the one to be concerned with if any of them.

titusfortner:
I started looking into the Selenium atoms. the Watir atoms in the Selenium project currently don't even work as compiled. I was trying to figure out how to update what we have. I have another branch I was playing around with before. I'll take a look again. If there are things in there that you think it makes sense let me know, I don't really understand any of the Google Closure things. :)

titusfortner:
@p0deje I re-implemented this moving all the atoms and the js snippets into a `js_snippets` directory with a single `#execute_js` interface. Let me know if the naming needs to be different, or if there are any other issues here.

p0deje (approved):

titusfortner:
@p0deje did a significant refactor when rebasing and went ahead and overhauled the specs. Not sure if you are interested in another pass through on `Select#select_by!` and its private `Select#matching_option`? I tried to make it more compatible with non-! by throwing an exception if the option is disabled. 

Also, I changed the order of precedence to: `:text` then `:label` then `:value`. I think this probably more closely matches how users will approach this, anyway. Some weird gotchas here. If there is no value specified on an option, our regex will locate based on text. So it makes sense (to me) that we always base it off of text first since it will never pull from value. Additional weirdness with label in that calling `#text` on an `Input` instance will return the value of the `label`, not the text, which is why some of the spec returns were changed to match the label instead of the text.

p0deje (approved):

Githraine:
What version will this go out in?

titusfortner:
Already released. 6.8.0

Issue #550, opened 2017-04-25
Window Management


titusfortner:
I'd like to see an alternative to either using a block to manage windows
```
browser.window(title: 'Foo').use { browser.element(id: 'foo').click
```
or using variables to store original window information:
```
original_window = browser.window
browser.window(title: 'Foo').use
browser.element(id: 'foo').click
original_window.use
```

Something where the original window gets stored:
```
browser.window(title: 'Foo').use
browser.element(id: 'foo').click
browser.reset_window
```
Other possible names: `browser.original_window.use` or `browser.previous_window.use`
or maybe we store it in a window variable: `window.previous`

Mostly I don't like the idea of requiring block usage to control code execution order when the block is not needed for scoping, and it would be nice not to have to store the variable explicitly in order to switch back.

titusfortner:
fixed by PR

PR #604, opened 2017-08-04
Support switching back to Original Window without using a block

titusfortner:
implements #550 

This gives more flexibility if you want to be in a different context than a single block to return to the original window (which is most use cases when dealing with multiple windows.

Allows this:
```ruby
browser.window(title: "foo").use
do_stuff
browser.reset_window!
```
instead of this:
```ruby
browser.window(title: "foo").use do
  do_stuff
end
```
or this:
```ruby
original_window = browser.window
browser.window(title: 'foo').use
do_stuff
original_window.use
```

### DIFF
--- a/lib/watir/window.rb
+++ b/lib/watir/window.rb
@@ -3,8 +3,9 @@ module Watir
     include EventuallyPresent
     include Waitable
 
-    def initialize(driver, selector)
-      @driver = driver
+    def initialize(browser, selector)
+      @browser = browser
+      @driver = browser.driver
       @selector = selector
 
       if selector.empty?
@@ -151,6 +152,7 @@ module Watir
     #
 
     def close
+      @browser.original_window = nil if self == @browser.original_window
       use { @driver.close }
     end
 
@@ -190,6 +192,7 @@ module Watir
     #
 
     def use(&blk)
+      @browser.original_window ||= current_window
       wait_for_exists
       @driver.switch_to.window(handle, &blk)
       self
--- a/lib/watir/browser.rb
+++ b/lib/watir/browser.rb
@@ -9,7 +9,7 @@ module Watir
     include HasWindow
     include Waitable
 
-    attr_writer :default_context
+    attr_writer :default_context, :original_window
     attr_reader :driver
     attr_reader :after_hooks
     alias_method :wd, :driver # ensures duck typing with Watir::Element
--- a/lib/watir/has_window.rb
+++ b/lib/watir/has_window.rb
@@ -11,7 +11,7 @@ module Watir
     #
 
     def windows(*args)
-      all = @driver.window_handles.map { |handle| Window.new(@driver, handle: handle) }
+      all = @driver.window_handles.map { |handle| Window.new(self, handle: handle) }
 
       if args.empty?
         all
@@ -30,11 +30,26 @@ module Watir
     #
 
     def window(*args, &blk)
-      win = Window.new @driver, extract_selector(args)
+      win = Window.new self, extract_selector(args)
 
       win.use(&blk) if block_given?
 
       win
+    end
+
+    #
+    # Returns original window if defined, current window if not
+    # See Window#use
+    #
+    # @example
+    #   browser.window(title: 'closeable window').use
+    #   browser.original_window.use
+    #
+    # @return [Window]
+    #
+
+    def original_window
+      @original_window ||= window
     end
 
     private
--- a/spec/watirspec/after_hooks_spec.rb
+++ b/spec/watirspec/after_hooks_spec.rb
@@ -38,7 +38,7 @@ describe "Browser::AfterHooks" do
 
   describe "#run" do
     after(:each) do
-      browser.window(index: 0).use
+      browser.original_window.use
       browser.after_hooks.delete @page_after_hook
     end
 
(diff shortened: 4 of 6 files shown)


p0deje:
I honestly don't see this bringing any value (blocks are Rubyish, saving state to variable is one of the basics), but if you think it is - let's merge.

titusfortner:
I agree that blocks are Rubyish, but I'm going through a phase of considering them unnecessarily over-used. :)

Good reasons for using blocks:
Maintaining context (specifying variables that won't be defined until execution)
Dynamically passing in values to code at execution time
Code executed repeatedly (waits, retries)

Bad reasons for using blocks:
Code execution order 
Like the `#when_present` code we got rid of, where block was essentially "and do this after"
Window / Frame switching makes a little more sense because there is both a before and an after action, but it is still just code execution order without any need to maintain context or reuse or dynamically accept parameters at run time.

p0deje:
That's essentially a context switch, so all the code is executed in specific window (context) and then return back.

titusfortner:
Ah, I meant to be saying "scope" when I was saying "context". If variables or methods aren't yet in scope until the block is called it is different than just passing along the code to run once inside very simple wrapper code.

p0deje:
Until you combine blocks with instance_eval or something like to, there is
no scope in it, so I don't think it's really true for Ruby blocks.

As I said, I am for keeping API simple so I think we need it.

titusfortner:
Ah, here we go: "closures allow a function to access non-local variables even when invoked outside of its immediate lexical scope." – Wikipedia

So, yes, exactly. If it isn't using `#instance_eval` or `#instance_exec` or similar, I think it is a good reason to think twice about using it. I've seen people write some extremely hard to read code with nested windows & frames using the page object gem.

For the record, I'm having a similar frustration with Ruby's DSL-All-The-Things! approach as well. :)

titusfortner:
Blocks are ridiculously powerful and I love using them in contexts like this: https://github.com/titusfortner/watir_drops/blob/master/spec/support/test_page.rb

p0deje:
I don't think blocks should only be used when you need to execute it within a different scope. I actually try to avoid it as much as I can because you constantly need to do a mental switch about what is `self` in block and outside. I prefer code which yields self, a typical example is configuration:

```ruby
RSpec.configure do |configuration|
  configuration.blah_blah
end
```

versus

```ruby
RSpec.configuration do
  blah_blah
end
```

However, that's more of a taste, so let's keep it more pragmatic. What's the use case here?

titusfortner:
@p0deje 

So for instance in our specs we do this in several places:
```ruby
    browser.window(index: 0).use
    browser.windows[1..-1].each(&:close)
```

Technically this code isn't correct since windows specifically do not have to get stored in an array in the order in which they were opened thanks to IE. This PR will make the first line more correct by returning to the original window regardless of where it is in the array.

Would it also make more sense for `Browser#reset_window!` to take care of the second line above as well? Maybe it should be `#reset_windows!` (plural)?

lib/watir/window.rb
```diff
@@ -3,8 +3,14 @@ class Window
     include EventuallyPresent
     include Waitable
 
-    def initialize(driver, selector)
-      @driver = driver
+    def initialize(browser, selector)
+      if browser.is_a?(Watir::Browser)
+        @browser = browser
+        @driver = browser.driver
+      else
```
p0deje: When does this happen?

lib/watir/window.rb
```diff
@@ -190,6 +196,9 @@ def url
     #
 
     def use(&blk)
+      if @browser
```
p0deje: How is it possible that there is no `@browser`?

p0deje (commented):
Ok, it now begins to be look like magic for me. When you name method `#reset_window!`, what would you expect it to do? What does reset mean? What is the window here if the method receiver is `browser`? If I would meet such a method in API, I'd have to read its code to understand what's going on there.

`#reset_windows!` ain't much better because it's still not clear what windows are going to be reset, and what does reset means. For me, honestly, this brings no value and I still think that people should use blocks and variables.

In regards to IE and ordering, I'd prefer something like scopes/enumerables that would allow to deal with both ordering and usage. For example,

```ruby
browser.window(index: 0).use
browser.windows.reject(:current?).each(&:close)
browser.windows.ordered[1..-1].each(&:close)
```

lib/watir/window.rb
titusfortner: currently this object is initialized by a driver, not a browser. Most likely no one initializes a window directly and goes through Browser, but I don't think this is a private API. which is why I put the deprecation notice.

lib/watir/window.rb
titusfortner: same as above. The plan is to remove this conditional & the deprecation notice whenever we do Watir 7

lib/watir/window.rb
p0deje: boring :)

that's essentially private API, nobody uses it, just remove pointless conditions

titusfortner:
You are totally correct, I was overthinking this.
`browser.original_window.use` is totally acceptable versus `browser.reset_window!`
Plus this adds resetting of `#original_window` when it is closed.

This is much better code.

p0deje (approved):
```ruby
browser.windows.reject(:current?).each(&:close)
```

This is so funny. I just made up this example as to show how API might look like, but then it turned it out to be a real code 😄 

Anyways, I like `#original_window` much better!

titusfortner:
Hah, I thought you were saying we should change it to that, and I'm all, "looks good to me!" :)

PR #605, opened 2017-08-04
implement TableCell#column_header method

titusfortner:
Based on a feature conversation in Slack channel.
This is easy to implement now that we have adjacent calls.
This assumes that the `th` elements are in the first `tr` which I think is required.

Not sure if there are other similar convenience methods we want to do?
`TableDataCell#row_value(index)` with `td.previous_sibling(index: index)` ?

### DIFF
--- a/lib/watir/elements/table.rb
+++ b/lib/watir/elements/table.rb
@@ -9,22 +9,16 @@ module Watir
     #
 
     def hashes
-      all_rows   = rows.to_a
-      header_row = all_rows.shift or raise Exception::Error, "no rows in table"
+      all_rows = rows.locate
+      header_row = all_rows.first || raise(Exception::Error, "no rows in table")
 
-      headers = header_row.ths.map { |header_cell| header_cell.text  }
-      result = []
+      header_type = header_row.th.exist? ? 'th' : 'td'
+      headers = header_row.send("#{header_type}s").map(&:text)
 
-      all_rows.each_with_index do |row, idx|
-        cells = row.cells.to_a
-        if cells.length != headers.length
-          raise Exception::Error, "row at index #{idx} has #{cells.length} cells, expected #{headers.length}"
-        end
-
-        result << headers.inject({}) { |res, header| res.merge(header => cells.shift.text) }
+      all_rows.entries[1..-1].map do |row|
+        cell_size_check(header_row, row)
+        Hash[headers.zip(row.cells.map(&:text))]
       end
-
-      result
     end
 
     #
@@ -35,7 +29,21 @@ module Watir
     #
 
     def [](idx)
-      row(:index, idx)
+      row(index: idx)
+    end
+
+    #
+    # @api private
+    #
+
+    def cell_size_check(header_row, cell_row)
+      header_size = header_row.cells.size
+      row_size = cell_row.cells.size
+      return if header_size == row_size
+
+      index = cell_row.selector[:index]
+      row_id = index ? "row at index #{index - 1}" : 'designated row'
+      raise Exception::Error, "#{row_id} has #{row_size} cells, while header row has #{header_size}"
     end
 
   end # Table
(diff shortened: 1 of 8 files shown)


lib/watir/elements/table_cell.rb
```diff
@@ -2,5 +2,12 @@ module Watir
   class TableCell < HTMLElement
     alias_method :colspan, :col_span
     alias_method :rowspan, :row_span
+
+    def column_header
+      index = self.previous_siblings.size
+      table = self.parent(tag_name: 'table')
+      TableHeaderCell.new(table.tr, {tag_name: 'th', index: index})
```
jkotests: This can be inconsistent with [Table#hashes](https://github.com/watir/watir/blob/master/lib/watir/elements/table.rb#L19-L22). `#hashes` checks for `td` and `th` elements, as well as makes sure the table has a basic layout.
~~~~~~~~ruby
cells = row.cells.to_a
if cells.length != headers.length
  raise Exception::Error, "row at index #{idx} has #{cells.length} cells, expected #{headers.length}"
end
~~~~~~~~

I don't have a strong opinion on which one is correct. But it would be nice to be consistent.

titusfortner: I've never #hashes, but it looks like it makes a ridiculous number of wire calls, most of which won't be needed to handle the basic use cases.

I'll take a look, though on how it is structured differently. Thanks!

titusfortner:
Also, this user's table doesn't actually have `th` cells, so we probably need the option to use the first `tr` as a header row: `browser.td(text: '1 331').column_header('tr')`. :-/

p0deje (approved):

p0deje (approved):
It seems strange to me to retrieve header from the cell. Naturally, what I would do is get table element and try to  get headers out of it.

However, if that works for users, let's merge.

lib/watir/elements/table_cell.rb
titusfortner: @jkotests so are you just saying that we should do a check to make sure:
```ruby
if self.parent(tag_name: 'tr').tds.size != table.tr.ths.size
    raise Exception::Error, "row at index #{idx} has #{cells.length} cells, expected #{headers.length}"
end
```

titusfortner:
@p0deje & @jkotests - ok, I made some more substantive changes to try to match it up better with what we already have in the Table class.

1. Oops the `ElementCollection#locate` method I implemented last time was supposed to return a collection, not an Array, specifically so you can use the collection instead of the array while working with it. (as shown in the `Table#hashes` implementation).

2. `#hashes` refactor still makes the same number of wire calls, but is implemented with Watir objects instead of arrays (the exception being the places where it uses `Array#shift` which I don't think we should implement in `ElementCollection`)

3. Allow `Element#selector` to be read (in this implementation it gives access to index in ElementCollection without needing `#to_a`).

Let me know if I made too many changes in here.

lib/watir/elements/table.rb
```diff
@@ -38,5 +34,16 @@ def [](idx)
       row(:index, idx)
     end
 
+    # @private
```
p0deje:
```ruby
#
# @api private
#
```

spec/watirspec/elements/td_spec.rb
```diff
@@ -70,4 +70,12 @@
     end
   end
 
+  describe "#column_header" do
+    it "returns the corresponding column header" do
+      td = browser.td(text: '1 331').column_header
+      expect(td.text).to eq 'Income tax'
+    end
+
```
p0deje: Remove this empty line

spec/watirspec/elements/td_spec.rb
```diff
@@ -70,4 +70,12 @@
...
+  end
+
```
p0deje: Remove this empty line

p0deje (approved):
Please see my tiny comments.

spec/watirspec/elements/td_spec.rb
titusfortner: I think how we are doing this one is that if there is an empty line after the `describe` line, then there is an empty line before the closing `end`. At least I've been trying to do that. Keeping up with style things is hard. :-D I should do another pass with rubocop at some point.

PR #607, opened 2017-08-07
implement Element#text! with nokogiri

titusfortner:
`Element#text!` should be implemented with nokogiri, not JavaScript. The main use case for this (for good or for bad) is doing a bunch of text verification on a page.

Rather than a bunch of wire calls for each element, this will grab the page source once and allow you to continue to use it for every `#text!` method you call.

The cached source is reset with an After Hook (which is only added if `#text!` is used) to prevent user from accessing old data.

`@element` is also reset after `#text!`, since the element needs to be relocated for any non-`#text!` method.

Gets rid of the `extensions/nokogiri.rb` file, since that is now accomplished with `browser.text!`

This feature does not support elements nested more than one layer below the browser or an iframe.

Implementation uses `#find_first_by_multiple` to force the selector to return XPath (or CSS) that Nokogiri can use. The only tricky part was pulling out the `#locate_element` & `#locate_elements` code. The rest of the locator logic is exactly correct, just need to override the actual location mechanism.

I ran the specs with `#text!` aliased on `#text`. Other than the nested elements, and the text that is not visible still showing up, there were only a couple edge cases that I don't think will matter to users who might want to take advantage of this feature.

It did cause me to add `afterhooks.run` after several more methods when it was obvious that the doc needed to be updated but was not.


p0deje:
What bothers me most in this is that location is different based on methods you use, so it's possible that `#text` and `#text!` will produce different behavior.

I do understand that there is a need for people parsing a lot of text, but it seems too much to have a custom locator strategy that works for only 1 method. Why do we want this to be a part of Watir? It fits more into watir-nokogiri gem.

jkotests:
I started trying to re-write the Watir-Nokogiri gem as a [driver](https://github.com/jkotests/static_html_driver/tree/master/lib/static_html_driver) instead. My hope is then to monkey patch a method into the Browser and Element that allows for switching to static HTML parsing. For example, it might be:

~~~~~~~~ruby
element.static.text
~~~~~~~~

The code is a bit of a mess since I am still unfamiliar with the Selenium code base and webdriver specs. However, I do believe that it will be a more consistent and maintainable static HTML parsing approach than my original hacking of the Watir-Nokogiri.

titusfortner:
But that's the thing, I don't want to use a separate nokogiri driver. Doing a custom locator strategy also doesn't work if I only want it to apply to one method. All other options would require monkey patching, which is specifically what the location code extensibility was designed to prevent. 

I want to do all of the normal Watir things with the normal Watir code, I just sometimes want to get text via nokogiri. I don't think it is a big deal that different methods (`#text` and `#text!`) sometimes return different things; we already know that some of our location strategies are off in just this way (#342), this would actually be more explicit about the fact that there is a difference. In fact based on how our Regex matching works right now, it might make more sense for *it to use this nokogiri implementation since it should return the same thing as current with a huge speed boost...

What if we used `#static_text` instead of `#text!`? I just thought rolling everything out with bangs would be symmetric and such. The reason it belongs in the Watir project is the same reason the javascript would belong there. I don't see any real difference in the approach to non-default non-selenium implemented performance enhancing code.

titusfortner:
As a weekend exercise I decided to implement this with a separate gem: [Watigiri](https://github.com/titusfortner/watigiri) (you know to match "Watizzle"). Anyway.

Overall I'm pleased with it, but my implementation requires monkey patching several things onto Watir that I'm not super excited about. (Also, this is what the code from #630 uses).

That gem would essentially let people just do `Element#text!` without needing to do anything more than `gem 'watigiri'` & `require 'watigiri'` in their code. No instantiating Nokogiri objects, no re-implementing a ton of Watir code. It's essentially just [4 methods in Locator class](https://github.com/titusfortner/watigiri/blob/master/lib/watigiri/locators/element/locator.rb) and then the monkey patches.

So, this solves my use case, but not @jkotests's where the concern is speed of element location especially with regular expressions. I have an idea for implementing this. I thought perhaps Watizzle might be a better solution for handling this, but @p0deje expressed concerns with the universal compatibility of Ruby -> JS Regular Expressions.

I'm thinking we could something like this...
```ruby
noko_elements = @query_scope.browser.doc.xpath(what).to_a
index = noko_elements.find_index { |e| matching_function(e) }
watir_element = @query_scope.elements(xpath: what)[index]
```

Issues would be:
1. does `#matching_function` include checking for visibility?
2. Do we implement a separate locator for this gem? (I'm thinking `:html` as I referenced in #631)
3. Does it only apply to selectors that include regular expressions? All selectors with regex or a subset?

titusfortner:
@p0deje I still kind of think I'd like this to be built-in functionality in Watir. At the very least we probably should look into what we talked about at the last Test Automation Bazaar - allowing users to set multiple locator strategies by order of precedence. But, I'll close this PR for now.

@jkotests do you have any interest in collaborating on the watigiri project to support faster location for regex? It leverages the extensibility code Alex wrote last year, and my goal would be to make usage completely transparent to the user after they require it. I'm curious what your core requirements are based on how you'd want to use nokogiri: is there more to it than just regex location? Or are you interested in continuing watir-nokogiri since it would still support watir-classic? :)

jkotests:
@titusfortner the goal of Watir-Nokogiri was to provide static HTML parsing using the Watir syntax. Watir(-WebDriver) can be slow for a number of things, not just locating by Regexp. The example that got me started is parsing large tables. In the application I work with, we generate some large table reports - eg 2,500 cells. Parsing the table, eg `table.strings` takes several minutes to execute. A number of years back, Zeljko had suggested using Nokogiri to solve the problem. However, this meant learning XPath/CSS, which we tend to avoid with Watir. Watir-Nokogiri was meant to solve this problem.

Anyways, I am open to any collaboration (whether it be just making Regexp searching faster or full HTML parsing).

titusfortner:
@jkotests - yeah I saw Zeljko's post from after the talk given at the Test Automation Bazaar in 2012.
 
Oh weird I didn't know `#strings` existed. Why is it in `RowContainer` instead of a method on the `Table` class? Oh huh, so it will work with `TableSection` as well. Interesting.

Right, so my thought was to implement any and all repeated text heavy stuff using Watigiri. So, at this point that looks like implementing & monkey patching in: `Element#text!`, `Table#string!` and `Table#hashes!`. These would all just use `#text_content` instead of `#text`. Plus the overriding of the default regex parsing for location. Would this address all of your use cases, or am I missing something?

Or... should using this gem override the default `#text`/`#string`/`#hashes` without the bangs?

Do you currently use Watir-Nokogiri in production for anything right now? Hard to see who else might be using it since there's only one release of it 4 years ago. :-D

jkotests:
I don't currently use Watir-Nokogiri. At the time I created it, I didn't have a use for it since Watir-Classic was quite fast. When I finally ran into a scenario that would benefit from it, Watir-Nokogiri was too out-of-date with Watir that I just used Nokogiri. But in terms of usage, I know it's come up as a request for using it with Cheezy's Page-Object gem. Apparently a [wiki page was even written](https://github.com/cheezy/page-object/wiki/Use-Page-Object-On-Any-HTML-Fragment-Or-Page-(with-WatirNokogiri)).

Interesting. The wiki suggests that Capybara might have a static HTML parsing capability -https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/simple.rb

I think monkey-patching in a couple of methods is still a bit short of where I would want to go. If you take the large table example again, what if I want to get an attribute value from each row of a 100 row table? That's a minimum of 100 wire calls. Watir-Nokogiri (or the newer static HTML driver) would be just a single wire call to get the element HTML.

titusfortner:
Then we can create a `#value!` and/or `#attribute!` method? That's what I'm looking to figure out, what all makes sense to add, because I don't think it is actually that many methods that make sense to implement like this.

I guess I'm trying to find a solution that:

1. Provides a simple interface that abstracts away the implementation details (i.e. same interface could be used by Watizzle, or another JS implementation, etc). User is *always working with a Watir element, never a Nokogiri node directly.
2. Takes advantage of Watir's existing Selector engine so everything supported by Watir is supported by the ~giri implementation
3. Takes advantage of after hooks to minimize likelihood of using stale document
4. Focused so that it is only used in the places that make sense (e.g. used for regex location, but not id location; called from a text method but not a click method, etc)

Interesting on the `page_object_watir_nokogiri`. It isn't on rubygems so it looks like it is "single use?" :)

jkotests:
Would the solution allow locating nested elements without wire calls?

titusfortner:
I hadn't considered that (it would only save a couple wire calls), but I think it is doable.

jkotests:
In that case, sounds like Watirgiri might achieve the same thing. Give me some time to better understand the code, but I still feel like there's more duplication required than faking out a driver.

Issue #602, opened 2017-08-03
Fix Generator


titusfortner:
We need to get the generator code working again. 

My cursory investigation also leads me to wonder if we shouldn't be basing our code off of [W3C HTML 5.1](https://www.w3.org/TR/html/) instead of the [HTML Living Standard](https://github.com/whatwg/html) like we're currently doing. If we are going to use the latter, we probably should be running the generator regularly (before releasing). There appears to be considerable disdain toward W3C from WHATWG members, but I also kind of like the idea of saying we are "version X" compatible. 🤷‍♂️ 

Living Standard also has a [raw file on github](https://github.com/whatwg/html/blob/master/source) we can parse instead of working with a much larger html file.

This might also mean updating Jari's IDL code? Or forking it? Or rewriting something similar but different?

titusfortner:
For giggles I tried to implement something with some complicated nokogiri parsing, completely avoiding the webidl gem. It.. mostly works, but obviously I needed to use hard coded variables and things that aren't desirable, and I'm stuck on getting the `GlobalEventHandlers` and `WindowEventHandlers` to work, since they aren't at all delineated in the xml. And multiple elements implement them; right now now everything is getting put in `HTMLElement` where only `GlobalEventHandlers` belong. :-/

Anyway, this is it. Not sure how worthwhile it is to keep going with it. I think it'd get us 90% of what we want, but then there's a chance that it's going to miss things. https://github.com/watir/watir/commit/6de374

There have been a bunch of changes to the spec in the past 2 years, though it's hard to say how much because the order of the old and new spec files is also very different.

p0deje:
I'm not sure when I'll have time to take a look at it, but generally usage of WebIDL gem is simple, we just need to extract IDL interfaces correctly and pass them in.

In past, most of the issues were coming from the fact that HTML spec had bugs and/or invalid IDLs.

titusfortner:
Fixed now.

PR #610, opened 2017-08-08
Fix generator and re-generate HTML elements from spec

p0deje:
The fix relies on jarib/webidl#6 so I could only re-generate with `LOCAL_WEBIDL=1` environment variable.

@titusfortner Please have a look. This should solve #602.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ end
 
 if ENV['LOCAL_WEBIDL']
   gem 'webidl', path: File.expand_path('../webidl')
+else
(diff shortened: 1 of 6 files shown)


titusfortner (approved):
So you think we should move from whatwg to w3c? I think that probably makes the most sense for us. "HTML 5.1 Compliant"

I would think the parser should also be able to handle both of these as well
https://html.spec.whatwg.org/
https://www.w3.org/TR/2017/PR-html51-20170803/single-page.html

But I guess if we have it working for the current published spec that's probably good enough.

p0deje:
Merged in 3dd6c24,

PR #612, opened 2017-08-08
implement extra methods and add aliases from page object gem

titusfortner:
Now that Page Object gem is entirely implemented by Watir, I think we should be moving some of the things that are implemented there that aren't technically related to Page Objects into Watir directly. Many of these things should already have been done in Watir and I just didn't think to implement them.

1. Several obvious aliases
2. For attributes that have Watir method equivalents, the default behavior should be the method, not the attribute. We can put these into a constant if that makes more sense. (except `Font#size` which overrides the superclass method because it is different)
3. `html_element.rb` file to load after all `#attribute` calls are made to alias `#attribute` to `#attribute_value`. The alternative is to rename `#attribute` in the helper file to `#set_attribute` which I might actually prefer, but thought I'd get feedback on it before changing it.
4. Six new element methods with documentation & tests
5. Not sure that's the right error to throw when trying to click a hidden element. thoughts?

I'm still digging into the Page Object code, so there will likely be more things, but these are the ones that I got in the first pass.

@AlexisKAndersen or @cheezy let me know if you have any input on this as well.

### DIFF
--- a/lib/watir/elements/image.rb
+++ b/lib/watir/elements/image.rb
@@ -16,18 +16,6 @@ module Watir
       )
     end
 
-    #
-    # Returns the image's width in pixels.
-    #
-    # @return [Integer] width
-    #
-
-    def width
-      element_call do
-        driver.execute_script "return arguments[0].width", @element
-      end
-    end
-
   end # Image
 
   module Container
--- a/lib/watir/attribute_helper.rb
+++ b/lib/watir/attribute_helper.rb
@@ -38,6 +38,10 @@ module Watir
     #  @return [$1] value of $3 property
     #
     def attribute(type, method, attr)
+      if %i(size selected? clear style width height).include? method
+        Watir.logger.debug "#{self}##{method} will use direct Element method not #attribute_value by default"
+        return
+      end
       typed_attributes[type] << [method, attr]
       define_attribute(type, method, attr)
     end
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -112,7 +112,17 @@ module Watir
             end
           end
     end
-    alias_method :locate, :to_a
+
+    #
+    # Locate all elements and return self.
+    #
+    # @return ElementCollection
+    #
+
+    def locate
+      to_a
+      self
+    end
 
     #
     # Returns true if two element collections are equal.
--- a/lib/watir/elements/checkbox.rb
+++ b/lib/watir/elements/checkbox.rb
@@ -18,6 +18,7 @@ module Watir
     def set(bool = true)
       set? == bool ? assert_enabled : click
     end
+    alias_method :check, :set
 
     #
     # Returns true if the element is checked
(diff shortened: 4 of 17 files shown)


lib/watir/elements/image.rb
```diff
@@ -16,18 +16,6 @@ def loaded?
       )
     end
 
-    #
-    # Returns the image's width in pixels.
-    #
-    # @return [Integer] width
-    #
-
-    def width
```
kapoorlakshya, line 25: Just curious why this is removed? Is this part of the change in point 2?

titusfortner: Because it returns the same value as the new `Element#width` method. No need to override in a subclass.

kapoorlakshya: Ah, got it. Thanks!

titusfortner: Actually, for the Elements that include a valid attribute for `width` (`Table`, `Input`, etc), it matches the 'y' value returned by `WebDriver::Element#size`. The Selenium call can be made on any element, though, so I implemented the code to always use it instead of `#attribute_value`.

lib/watir/elements/html_element.rb
```diff
@@ -0,0 +1,7 @@
+module Watir
+  class HTMLElement < Element
+
+    alias_method :attribute, :attribute_value
```
p0deje: Why not do this in element.rb?

p0deje (approved):
I don't understand why you decided to add `HTMLElement` class and what's the need for `attribute` alias, but everything else is fine.

lib/watir/elements/html_element.rb
titusfortner: `element.rb` needs to be loaded before `html_elements.rb` and the alias needs to happen after `html_elements.rb`. The alternative is renaming to `#set_attribute` which I think I might actually prefer to this.

p0deje: But why loading order matters? Can you explain, please?

titusfortner: When we're actually loading the elements we need `#attribute` to use this: https://github.com/watir/watir/blob/master/lib/watir/attribute_helper.rb#L40
This is kind of a hacky workaround in case you don't want to rename the one in attribute helper. Or disagree on aliasing attribute in general. We can always keep the alias in PO gem. But matching `Watir::Element#attribute` to `Selenium::WebDriver::Element#attribute` makes sense to me.

lib/watir/elements/image.rb
p0deje: But `AttributeHelper` is extended, so element has `.attribute` class method and here you alias an instance method. There shouldn't be a collision.

titusfortner: Hmmm, I would have sworn I had an issue and this was the fix. Maybe I screwed up something else during the testing. I'll give it a try in elements and report back.

titusfortner: And again, you are completely correct. I must have screwed something else up when getting this to work and thought that part was necessary. Thanks!

PR #615, opened 2017-08-09
use downcase instead of snake case for attributes

titusfortner:
per our discussion in #601 

This gets all the specs passing.

There are currently 83 different attributes that use snake case: https://gist.github.com/titusfortner/032d80553aa1ff10061de7eb0fe2919b

The only one this gives special consideration for is `:http_equiv`. Are we ok with the rest being downcased?


lib/watir/attribute_helper.rb
```diff
@@ -38,6 +38,10 @@ def attribute_list
     #  @return [$1] value of $3 property
     #
     def attribute(type, method, attr)
+      if %i(thead tfoot).include? method
```
p0deje: Can we remove this?

lib/watir/elements/html_elements.rb
```diff
@@ -76,8 +76,8 @@ class HTMLElement < Element
     attribute(String, :oncopy, :oncopy)
     attribute(String, :oncut, :oncut)
     attribute(String, :onpaste, :onpaste)
-    attribute(String, :content_editable, :contentEditable)
-    attribute("Boolean", :content_editable?, :isContentEditable)
+    attribute(String, :contenteditable, :contentEditable)
+    attribute("Boolean", :iscontenteditable?, :isContentEditable)
```
p0deje: This doesn't look correct, it should be `contenteditable?`

lib/watir/elements/html_elements.rb
```diff
@@ -436,34 +436,34 @@ class LabelCollection < ElementCollection
   end
 
   class Form < HTMLElement
-    attribute(String, :accept_charset, :acceptCharset)
+    attribute(String, :acceptcharset, :acceptCharset)
```
p0deje: This one should probably stay snake cased

p0deje (changes requested):
Please see my comments.

Have you re-generated this from spec? If yes, than we should do the same for svg_elements.

lib/watir/attribute_helper.rb
titusfortner: If we remove it, then this code:
```
browser.table(id: 'table').thead
```
will be the same as this code: 
```
browser.table(id: 'table').attribute_value(:thead)
```
Instead of returning a `TableSection` instance as would be expected

titusfortner:
rebased and updated with corrections

titusfortner:
nevermind, have more work to do.
Also, Generator is failing on SVG :(

lib/watir/attribute_helper.rb
p0deje: Then those attribute methods should be be generated in the first place. Do something like this to achieve that:

```diff
diff --git a/lib/watir/generator/html/generator.rb b/lib/watir/generator/html/generator.rb
index 8a7ed73..8f07633 100644
--- a/lib/watir/generator/html/generator.rb
+++ b/lib/watir/generator/html/generator.rb
@@ -16,7 +16,9 @@ module Watir
       end

       def ignored_attributes
-        %w(cells elements hash rows span text)
+        attributes = %w(elements hash span text)
+        # table elements should not be accessible as attributes
+        attributes += %w(cells tHead tFoot rows)
       end

       def generator_implementation
```

titusfortner: I tried using `#ignored_attributes` first. It is only called in the cleanup step, and it didn't seem to do anything.

p0deje: Just tried it and it works correctly.

You probably used `thead` instead of `tHead`. Ignored attributes should be IDL ones.

titusfortner: why, yes, that is exactly what I did :)

titusfortner: Should I move these there as well? https://github.com/watir/watir/blob/master/lib/watir/attribute_helper.rb#L41

p0deje: Yes please!

titusfortner:
Any easy fixes with the failing SVG generation?

lib/watir/attribute_helper.rb
p0deje: I don't understand, did ignored_attributes work for you or not?

p0deje:
So, there is a bug in current SVG spec. IDL grammar is not valid for ShadowAnimation interface. I've sent an email to SVG W3C mailing list about this problem. but until it's fixed we'll have to edit spec:

```
sed -i.bak "s/Constructor<\/i> (/Constructor<\/i>(/" support/svg.html
```

After that, you should be able to generate elements from spec.

titusfortner:
All requested changes have been made; before merging it, though...

These are the arguments that are now set to get ignored
`%w(cells elements hash rows span text size selected? clear style width height tHead tFoot)`

We are using attributes both for custom locators and for custom methods on the applicable elements. I think some of these should be allowed as locators, but not as methods, not entirely removed from `html_elements.rb`.

This PR is better code than what we have now, so I'll push forward on it, assuming you don't see any issues with it on your next look. But I still wanted to get your opinion on changing it again. :)

p0deje (approved):
> We are using attributes both for custom locators and for custom methods on the applicable elements. I think some of these should be allowed as locators, but not as methods, not entirely removed from html_elements.rb.

I don't think there will be a need for that, but let's see.

lib/watir/elements/svg_elements.rb
```diff
@@ -634,15 +580,15 @@ def text_paths(*args)
     end
     Watir.tag_to_class[:textPath] = TextPath
 
-    # @return [TSpan]
+    # @return [Text]
```
titusfortner: I can't figure out why the generator is making this `Text` instead of `TSpan`. Everything works if I manually change it back. We can:
1. Manually change it and create issue to investigate more later
2. Add it to the ignored list (like I also ended up doing with unknown tag) 
3. Figure out how to fix the generator code

I'd like to get this code released out pretty much right away since there are a couple bug fixes that people need. So, @p0deje, if 3 isn't an option for you right away, let me know if you prefer 1 or 2 so I can do the release. Thanks.

p0deje: The following should do what you want:

```diff
diff --git a/lib/watir/generator/svg/spec_extractor.rb b/lib/watir/generator/svg/spec_extractor.rb
index a5b5ff8..74db8ac 100644
--- a/lib/watir/generator/svg/spec_extractor.rb
+++ b/lib/watir/generator/svg/spec_extractor.rb
@@ -24,15 +24,21 @@ module Watir
           list.each do |node|
             tag_name = node.css('a span').inner_text.strip
             id = node.css('a').attr('href').to_s
+            next if external_interface?(id)

-            # Some interfaces are actually defined in different specs
-            # (for example, clipPath), so we ignore them for now
-            if id =~ /^#.+/
-              interface_css = 'div.element-summary a.idlinterface'
-              interface_name = @doc.css("#{id} #{interface_css}, #{id} ~ #{interface_css}").first.inner_text.strip
+            interface_css = 'div.element-summary a.idlinterface'
+            interface_definitions = @doc.css("#{id} #{interface_css}, #{id} ~ #{interface_css}")

-              result[tag_name] = interface_name
-            end
+            interface_definition = if tag_name == 'tspan'
+                                     # TSpan is defined along with Text
+                                     # so the first IDL definition is SVGTextElement
+                                     interface_definitions[1]
+                                   else
+                                     interface_definitions[0]
+                                   end
+            interface_name = interface_definition.inner_text.strip
+
+            result[tag_name] = interface_name
           end
         end
       end
@@ -41,6 +47,12 @@ module Watir
         %w(SVGHatchElement SVGHatchPathElement SVGSolidColorElement)
       end

+      # Some interfaces are actually defined in different specs
+      # (for example, clipPath), so we ignore them for now.
+      def external_interface?(id)
+        id !~ /^#.+/
+      end
+
     end # SVG::SpecExtractor
   end # Generator
 end # Watir
```

p0deje:
FTR, here is an email I've sent to SVG working group to address the problem with IDL interface: https://lists.w3.org/Archives/Public/www-svg/2017Aug/0005.html.

titusfortner:
merged

PR #616, opened 2017-08-09
implement TableCell#sibling_from_header

titusfortner:
An example use case for this is being able to locate a specific cell in a table (e.g. by text) and wanting to check a checkbox in an adjacent cell. This would allow:

```
browser.td(text: 'Jason Smith').sibling_from_header(text: "Opted In").checkbox.check
```
Can also use index, but not sure if it makes more sense than what is available now.

Now:
```
browser.td(text: 'Jason Smith').parent(tag_name: 'tr').td(index: -1).checkbox.check
```
PR:
```
browser.td(text: 'Jason Smith').sibling_from_header(index: -1).checkbox.check
```

It's the most descriptive name I could come up with. Alternate naming options welcomed. :)

### DIFF
--- a/spec/watirspec/elements/td_spec.rb
+++ b/spec/watirspec/elements/td_spec.rb
@@ -77,4 +77,20 @@ describe "TableCell" do
     end
   end
 
+  describe "#sibling_from_header" do
+    it "returns the corresponding sibling cell by text" do
+      local_td = browser.td(text: '1 331')
+
+      td = local_td.sibling_from_header(text: 'After income tax')
+      expect(td.text).to eq '4 532'
+    end
+
+    it "returns the corresponding sibling cell by index" do
+      local_td = browser.td(text: '1 331')
+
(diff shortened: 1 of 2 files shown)


spec/watirspec/elements/td_spec.rb
```diff
@@ -77,4 +77,20 @@
     end
   end
 
+  describe "#sibling_from_header" do
+    it "returns the corresponding sibling cell by text" do
+      local_td = browser.td(text: '1 331').tap(&:exist?)
```
p0deje: Why tap?

p0deje (approved):

spec/watirspec/elements/td_spec.rb
titusfortner: because I had a debug point right after it to check on things as I was testing it. :-D
They don't belong in the final code, thanks.

PR #618, opened 2017-08-10
provide more collections options

titusfortner:
More stuff from comparing code to Page Object gem.

We unnecessarily override things Enumerable gives us for free, and there are a few additional ways we can take advantage of elements that can act like collections.

Oddly, splitting out `Table#headers` will be directly useful to page object gem implementation.

Also, `#siblings`!


p0deje (commented):
Looks good.

titusfortner:
Merged: https://github.com/watir/watir/commit/78ef6ad343c7d00bd5647d1563bb5f5abb30190c

PR #619, opened 2017-08-10
fix bug of not waiting for Select List

titusfortner:
Even more tweaking the select list code...

Mostly timing related...
`#options` needs to go though an `#element_call` to get the specified waiting behavior 

I'm also removing the waits for the options, since they didn't respect the `#relaxed_locate` setting and I'm not sure how to check both text and labels without it causing problems somewhere. I think we're ok to assume that if the select list is present that the options should be ready to go right away. Maybe I'll have a better thought on this after I sleep more, but this should be good.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -593,7 +593,7 @@ module Watir
       return if present?
 
       begin
-        @query_scope.wait_for_present
+        @query_scope.wait_for_present unless @query_scope.is_a? Browser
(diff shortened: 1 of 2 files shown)


p0deje (approved):
Looks good but are there any specs?

titusfortner:
We don't have specs for any other method that uses element_call. This will all be much nicer once we don't have one catch all setting for everything. :)

PR #624, opened 2017-08-15
implement RadioSet

titusfortner:
This still needs all of documentation in the class itself, but I wanted to run the code by people before I started tidying it up.

Radio buttons with the same name are not independent because only one can be selected at a time. Ideally a user would be able to treat it just like a Select List and choose amongst the options available. I literally took the `select_list_spec.rb` file to create the `radio_set_spec.rb` file and tweaked it based on what makes sense for radio buttons vs options. There are probably a bunch more specs I'm missing since `RadioSet` doesn't inherit from anything.

This should probably include Enumerable for iteration... Yeah, this needs more work.

Also, this addresses the age old question of what `Radio#text` should return. :)

### DIFF
--- a/lib/watir/elements/input.rb
+++ b/lib/watir/elements/input.rb
@@ -1,0 +1,15 @@
+module Watir
+  class Input < HTMLElement
+
+    #
+    # Returns applicable label.
+    # Is not lazy loaded.
+    #
+
+    def label
+      el = parent(tag_name: 'form').label(for: id)
+      return el if el.exist?
+    end
+
+  end # Input
+end # Watir
--- a/lib/watir/elements/input.rb
+++ b/lib/watir/elements/input.rb
@@ -7,8 +7,7 @@ module Watir
     #
 
     def label
-      el = parent(tag_name: 'form').label(for: id)
-      return el if el.exist?
+      parent(tag_name: 'form').label(for: id)
     end
 
   end # Input
--- a/lib/watir/elements/input.rb
+++ b/lib/watir/elements/input.rb
@@ -2,8 +2,9 @@ module Watir
   class Input < HTMLElement
 
     #
-    # Returns applicable label.
-    # Is not lazy loaded.
+    # Returns label element associated with Input element.
+    #
+    # @return [Watir::Label]
     #
 
     def label
--- a/lib/watir/radio_set.rb
+++ b/lib/watir/radio_set.rb
@@ -47,18 +47,31 @@ module Watir
     # @return Watir::Radio
     #
 
-    def radio(opt)
-      name_opt = name.empty? ? {} : {name: name}
-      frame.radio(opt.merge name_opt)
+    def radio(opt = {})
+      n = name
+      if !n.empty? && (!opt[:name] || opt[:name] == n)
+        frame.radio(opt.merge name: n)
+      elsif n.empty?
+        return source
+      else
+        raise Watir::Exception::UnknownObjectException, "#{opt[:name]} does not match name of RadioSet: #{n}"
+      end
+
     end
 
     #
     # @return Watir::RadioCollection
     #
 
-    def radios
-      opt = name ? {name: name} : {}
-      element_call(:wait_for_present) { frame.radios(opt) }
+    def radios(opt = {})
+      n = name
+      if !n.empty? && (!opt[:name] || opt[:name] == n)
+        element_call(:wait_for_present) { frame.radios(opt.merge name: n) }
+      elsif n.empty?
+        Watir::RadioCollection.new(frame, element: source.wd)
+      else
+        raise Watir::Exception::UnknownObjectException, "#{opt[:name]} does not match name of RadioSet: #{n}"
+      end
     end
 
     #
--- a/lib/watir/radio_set.rb
+++ b/lib/watir/radio_set.rb
@@ -1,0 +1,218 @@
+module Watir
+  class RadioSet
+    extend Forwardable
+    include Watir::Exception
+    include Enumerable
+
+    delegate [:exists?, :present?, :visible?, :browser] => :source
+
+    attr_reader :source, :frame
+
+    def initialize(query_scope, selector)
+      unless selector.kind_of? Hash
+        raise ArgumentError, "invalid argument: #{selector.inspect}"
+      end
+
+      @source = Radio.new(query_scope, selector)
+      @frame = @source.parent(tag_name: :form)
+    end
+
+    #
+    # Yields each TableCell associated with this row.
+    #
+    # @example
+    #   radio_set = browser.radio_set
+    #   radio_set.each do |radio|
+    #     puts radio.text
+    #   end
+    #
+    # @yieldparam [Watir::RadioSet] element iterate through the radio buttons.
+    #
+
+    def each(&block)
+      radios.each(&block)
+    end
+
+    #
+    # Get the n'th radio button in this set
+    #
+    # @return Watir::Radio
+    #
+
+    def [](idx)
+      radios[idx]
+    end
+
+    #
+    # @return Watir::Radio
+    #
+
+    def radio(opt)
+      name_opt = name.empty? ? {} : {name: name}
+      frame.radio(opt.merge name_opt)
+    end
+
+    #
+    # @return Watir::RadioCollection
+    #
+
+    def radios
+      opt = name ? {name: name} : {}
+      element_call(:wait_for_present) { frame.radios(opt) }
+    end
+
+    #
+    # Returns true if any radio buttons in the set are enabled.
+    #
+    # @return [Boolean]
+    #
+
+    def enabled?
+      any?(&:enabled?)
+    end
+
+    #
+    # Returns true if all radio buttons in the set are disabled.
+    #
+    # @return [Boolean]
+    #
+
+    def disabled?
+      !enabled?
+    end
+
+    #
+    # Returns true if all radio buttons in the set are disabled.
+    #
+    # @return [String]
+    #
+
+    def name
+      @name ||= source.name
+    end
+
+    #
+    # If RadioSet exists, this always returns 'radio'.
+    #
+    # @return [String]
+    #
+
+    def type
+      source.send :assert_exists
+      'radio'
+    end
+
+    #
+    # Returns true if the radio set has one or more radio buttons where label matches the given value.
+    #
+    # @param [String, Regexp] str_or_rx
+    # @return [Boolean]
+    #
+
+    def include?(str_or_rx)
+      radio(label: str_or_rx).exist?
+    end
+
+    #
+    # Select the radio button whose value or label matches the given string.
+    #
+    # @param [String, Regexp] str_or_rx
+    # @raise [Watir::Exception::UnknownObjectException] if the Radio does not exist.
+    # @return [String] The value or text of the radio selected.
+    #
+
+    def select(str_or_rx)
+      found_by_value = radio(value: str_or_rx)
+      found_by_text = radio(label: str_or_rx)
+
+      if found_by_value.exist?
+        found_by_value.click unless found_by_value.selected?
+        return found_by_value.value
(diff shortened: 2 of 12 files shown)


titusfortner:
Ok, all of the documentation is updated, all the specs are passing.

This is going to make a big difference for page object implementations that won't have to treat each radio button separately.

jkotests:
I think this feature would be useful.

Would you ever expect someone to want a radio set for radios without a name attribute? The behaviour is a bit odd when I tested against the forms_with_input_elements.html.

~~~~~~~~ruby
set = browser.radio_set(id: /abs/) # just matches <input type="radio" id="new_user_newsletter_absolutely" value="absolutely">

set.exists?
#=> true

set.count
#=> 0
~~~~~~~~

There is no benefit to using RadioSet for radios without names. Maybe we should be throwing an exception when a name attribute cannot be found?

titusfortner:
So... the intention (for which I thought I had tests) is to have a RadioSet of count 1 if there is no name. The advantage of it is that you don't have to do anything different in client code if it is one or more than one. They can all be treated as RadioSet instances. Good catch on this, I'll figure out why it is doing that.

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,218 @@
...
+
+    #
+    # Select the radio button whose value or label matches the given string.
+    #
+    # @param [String, Regexp] str_or_rx
+    # @raise [Watir::Exception::UnknownObjectException] if the Radio does not exist.
+    # @return [String] The value or text of the radio selected.
+    #
+
+    def select(str_or_rx)
```
jkotests, line 137: This is different than `Select`, which has `#select` for matching text and `#select_value` for matching value. It would be nice to be consistent.

Though the solution might be to change `Select#select` to match value and text. Watir-Classic did this. Obviously if we were to do this, it would need to be a Watir 7.0 change.

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,218 @@
...
+
+    def [](idx)
+      radios[idx]
+    end
+
+    #
+    # @return Watir::Radio
+    #
+
+    def radio(opt)
```
jkotests: `opt` should be optional?

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,218 @@
...
+      name_opt = name.empty? ? {} : {name: name}
+      frame.radio(opt.merge name_opt)
+    end
+
+    #
+    # @return Watir::RadioCollection
+    #
+
+    def radios
```
jkotests: Is there a reason that we cannot specify `opt` like `#radio` does?

lib/watir/radio_set.rb
titusfortner: Oh? Watir Classic would match select options by either text or value? Yes, I'd like to see it changed in Select. Page Object gem actually recently implemented something to support that for select lists in their form filling method.

lib/watir/radio_set.rb
titusfortner: I was thinking of it like a private api/method and every place this code uses it needs to pass in the opt, but you're right that isn't what this is/should be. `my_radio_set.radio` should return the first radio in that set, the same as `my_radio_set.first` or `my_radio_set[0]`

lib/watir/radio_set.rb
titusfortner: yeah, again if this is public then we'd want to let users filter by class, etc.

lib/watir/radio_set.rb
titusfortner: @jkotests do you know off hand which took precedence for `Select#select` value or text? I would think we'd want value first (since it's presumably unique and more locale-safe), then text, but more people probably actually use text.

jkotests: [`Select#select`](https://github.com/watir/watir-classic/blob/master/lib/watir-classic/input_elements.rb#L45-L55) would have done by value, label and then text.

titusfortner:
@jkotests I just read your [radio/checkbox label blog post](https://jkotests.wordpress.com/2017/03/10/set-styled-radio-buttons-and-checkboxes-that-have-hidden-or-overlapped-the-input-field/).

I wanted to point out that the code in this PR would let you do:
`browser.checkbox(id: 'overlapped_checkbox').label.set`
Not sure how much cleaner it is, or if there are other things that this functionality would make easier.

jkotests:
@titusfortner, thanks for pointing that out. For that specific example, where the id is known, it is probably more efficient to locate the `label` directly - ie you know exactly what the `for` attribute is. In situations where you don't know the id, the `#label` method will definitely be a great addition.

titusfortner:
Not sure how crazy to get with it, but was considering the possibility of rescuing that error and auto retrying the click on its label.

lib/watir/elements/input.rb
```diff
@@ -0,0 +1,15 @@
...
+  class Input < HTMLElement
+
+    #
+    # Returns applicable label.
+    # Is not lazy loaded.
+    #
+
+    def label
+      el = parent(tag_name: 'form').label(for: id)
+      return el if el.exist?
```
jkotests: @titusfortner, can you explain why we don't want to lazy load the label element?

It seems weird to be returning `nil` when the label element doesn't exist. When there is no associated label, you will get a `NoMethodError`:

~~~~~~~~ruby
browser.radio.label.text
#=> undefined method `text' for nil:NilClass (NoMethodError)
~~~~~~~~

This isn't very descriptive of the problem. I would have expected the usual `UnknownObjectException`.

titusfortner:
@jkotests I think this addresses all of your concerns. :)
Let me know if I missed something.

lib/watir/elements/input.rb
```diff
@@ -0,0 +1,14 @@
+module Watir
+  class Input < HTMLElement
+
+    #
+    # Returns applicable label.
+    # Is not lazy loaded.
+    #
+
+    def label
+      parent(tag_name: 'form').label(for: id)
```
jkotests: Is there any benefit to going to the parent `form` rather than the `body`?

I didn't realize a `form` element was required, so when I created a test page without one, I was confused by the exception:

~~~~~~~~ruby
b.radio_set.enabled?
Watir::Exception::UnknownObjectException: unable to locate element: #<Watir::Form: located: false; {:tag_name=>"input", :type=>"radio"} --> {:tag_name=>:form, :adjacent=>:ancestor, :index=>0}>
        from E:/Temp/watir_test/watir-radio_group/lib/watir/elements/element.rb:690:in `assert_element_found'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/elements/element.rb:685:in `assert_exists'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/elements/element.rb:710:in `ensure_context'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/element_collection.rb:123:in `elements'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/element_collection.rb:69:in `to_a'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/element_collection.rb:28:in `each'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/radio_set.rb:33:in `each'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/radio_set.rb:84:in `any?'
        from E:/Temp/watir_test/watir-radio_group/lib/watir/radio_set.rb:84:in `enabled?'
        from (irb):38
        from C:/Ruby23/bin/irb.cmd:19:in `<main>'
~~~~~~~~

At least one [Stack Overflow question](https://stackoverflow.com/q/3294572/1200545) suggests that a `form` element is not required.

lib/watir/elements/input.rb
```diff
@@ -0,0 +1,14 @@
+module Watir
+  class Input < HTMLElement
+
+    #
+    # Returns applicable label.
+    # Is not lazy loaded.
```
jkotests: The "Is not lazy loaded." comment is no longer true.

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,231 @@
...
+      unless selector.kind_of? Hash
+        raise ArgumentError, "invalid argument: #{selector.inspect}"
+      end
+
+      @source = Radio.new(query_scope, selector)
+      @frame = @source.parent(tag_name: :form)
+    end
+
+    #
+    # Yields each TableCell associated with this row.
```
jkotests: Comment should be "Yields each Radio associated with this set"

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,231 @@
...
+    #
+    # @return [Boolean]
+    #
+
+    def disabled?
+      !enabled?
+    end
+
+    #
+    # Returns true if all radio buttons in the set are disabled.
```
jkotests: Should be something like "Returns the name attribute for the set"

lib/watir/radio_set.rb
```diff
@@ -0,0 +1,231 @@
...
+    # Returns true if two elements are equal.
+    #
+    # @example
+    #   browser.radio_set(id: 'new_user_newsletter_yes') == browser.radio_set(id: 'new_user_newsletter_no')
+    #   #=> true
+    #
+
+    def ==(other)
+      return false unless other.kind_of?(self.class)
+      frame == other.frame
```
jkotests: This is checking that both `RadioSet` belong to the same `form` element? This would be incorrect when a form has 2 sets of radio buttons. Given the page:

~~~~~~~~html
<html>
  <body>
    <form>
      <input type="radio" name="a" id="a1"><label for="a1">my a_1</label>  
      <input type="radio" name="a" id="a2"><label for="a2">my a_2</label>
      <input type="radio" name="a" id="a3"><label for="a3">my a_3</label>

      <input type="radio" name="b" id="b1" value="b1v"><label for="b1">my b_1</label>  
      <input type="radio" name="b" id="b2" value="b2v"><label for="b2">my b_2</label>
      <input type="radio" name="b" id="b3" value="b3v"><label for="b3">my b_3</label>
    </form>
  </body>
</html>
~~~~~~~~

Comparing the 2 groups gives `true`:

~~~~~~~~ruby
b.radio_set(name: 'a') == b.radio_set(name: 'b')
#=> true    #--> should be false
~~~~~~~~

I think you want to compare the `source` instead - ie the line should be:'

~~~~~~~~ruby
source == other.source
~~~~~~~~

lib/watir/elements/input.rb
titusfortner: Huh, interesting. I went to form because RadioSet doesn't make sense to me outside of a form context. These are independent of each other:
```html
<form name="sign_in" class="user">
    <input type="radio" name="opt">Opt In</input>
    <input type="radio" name="opt">Opt Out</input>
</form>
<form name="sign_up" class="user">
  <input type="radio" name="opt">Opt In</input>
  <input type="radio" name="opt">Opt Out</input>
</form>
```
If I went to `body` tag they would be considered the same `RadioSet` which they are not.

lib/watir/radio_set.rb
titusfortner: gah. copy/paste fail. Thanks.

lib/watir/elements/input.rb
jkotests: Hrmm... so ideally we would want the ancestor `form` if available, otherwise the `body`. Probably not worth addressing until someone runs into the issue.

titusfortner:
Ok, updated. Thanks for catching my sloppy errors. :)

Yeah, I think I'd like to avoid that extra conditional at this point on the form/body issue. Easy enough to add if we decide we want it.

equality needs to be comparing `#radios` since we can generate the same `RadioSet` from any of the child `Radio`, and we can't just verify that the names are the same, unfortunately.

jkotests:
Looks good to me.

titusfortner:
@p0deje any issues with this?

p0deje (approved):

titusfortner:
merged with rebase

PR #630, opened 2017-08-21
After hook updates

titusfortner:
Adds Afterhooks to places that might change DOM

Pulls out the specific references to webdriver in the default locator file. (I'll reference my reason for doing this soon).


p0deje (approved):
Have no idea why you need this, but the code looks good.

PR #631, opened 2017-08-21
fix bug with :text returning different values for String and Regex

titusfortner:
This is a stab at properly addressing #342 

The thought is to make `text: String` match the behavior of `text: Regexp` by ensuring the element is visible, and add `:html` locator to allow for the current behavior of `text: String`. 

This means that `html: Regexp` would then behave differently than `html: String`. 
But... and this is why I started working on this... it could be implemented specifically with [Watigiri gem](https://github.com/titusfortner/watigiri) or with Watizzle.

I think I need to think through the edge cases and add more specs, but I'm interested in big picture feedback on this. Because the number of elements that match on text are pretty low, I think this will tend to be only one extra wire call per text location.


p0deje:
Does it get slower now?

titusfortner:
No, it appears to be within the noise. ~65 minutes total for watirspec suite both master & this PR

What do you think of adding a new locator vs just deprecating current `:text` behavior (with change to be made in 7.0)?

What do you think about the locator being `:html` vs `:inner_text`. I'd also like to see the locator match a method that can be called on the element. So essentially #632 could be `#html` instead of (or aliased to) `#inner_text`.

Alternately, we could go the opposite way and evaluate `text: RegExp` using `inner_text` from #632 instead of `text` so it matches what `text: String` is doing so neither care about visibility. I don't think I'd prefer this, but its an option.

Thinking on this further, my current proposal is:

* `text: String` => Visible on Page (implemented with `#text` calls on found element(s))
* `text: RegExp` => Visible on Page (implemented with `#text` calls on found element(s))
* `html: String` => Exists in DOM (implemented with direct xpath location, no `#text` call)
* `html: RegExp` => Exists in DOM (implemented with `#inner_text` calls on found element(s))
* `alias_method :html, :inner_text` in `Element` (after #632)

Then Watigiri could just override the `html: RegExp` because overriding `text: RegExp` wouldn't make sense to save time since a wire call for visibility would need to be made on each element still. This would of course then require users to change a bunch of locator code if they want to take advantage of the speed enhancements, but that might actually be a good thing.

Thoughts? @jkotests? @tk8817?

titusfortner:
Ok, just added commit based on #632 that implements the proposal I outlined in the above comment.

jkotests:
My thoughts on the current proposal:
* Having `:text` always be based on the visible text would be good.
* `html: String` should also be based on `#inner_text`. Otherwise, based on the comments in #632, I think you will have problems with `String` and `Regexp` values not returning the same result.
* I would have the new locator be `:inner_text` and not aliased to `:html`. To me, a `:html` locator should actually check the element's outer HTML. At least that's what Watir-Classic does and I had actually used it this way once.

titusfortner:
@jkotests wait, didn't you say in the other PR that the method shouldn't be `#inner_text`? :)

`#dom_text`/`dom_text: 'foo'`?
`#html_text`/`html_text: 'foo'`?

Not sure `#all_text` makes sense to me.

Is there an example where XPath would find a string that `innerText` wouldn't or vice-versa? It should be the same parsing of the DOM, no? To me the question is whether innerText or textContent better matches what XPath locates.

jkotests:
Sorry, I meant to say that the new locator should be `:inner_text` or whatever the method in #632 ends up being named.

For an example of where the XPath and `innerText` is different, consider the page:

~~~~~~~~html
<html>
  <body>
    <div>
      <span style="display: none;">a</span> <span>b</span> 
    </div>
  </body>
</html>
~~~~~~~~

XPath will match the text "a b" where as the `innerText` is just "b". This results in different `String` vs `Regexp` results:

~~~~~~~~ruby
browser.div(html: 'a b').exists?
#=> true

browser.div(html: /a b/).exists?
#=> false
~~~~~~~~

titusfortner:
I updated this to use `text_content` as both the locator key & the method (instead of `html`).
Does that seem like the right name for matching what is in the DOM vs what is present on the page?

Any other issues with this?

jkotests:
I'm fine with `text_content`. It's a bit technical, but I don't have a better answer.

p0deje:
I'm not sure I agree on what's going on here, because having `:text` and `:text_content` seems to much for me and might be confusing. Still, I'm not sure what would be the best decision. On the one hand, there has been only 1 bug report related to text locator inconsistency, so maybe it's not a big deal to address. On the other hand, we should do our best to remove any inconsistencies.

However, I don't think that introducing a new locator is a real solution to the problem. It means: "Use `:text`, but whenever you don't understand what's happening, try using `:text_content`". Maybe the initial attempt to actually fix `:text` was more correct, even though it potentially makes locating slower. I'm not really going to use `:text_content` anytime, it just doesn't make sense for me.

In any event, it's up to you to merge it, but I don't think it's a very good idea. Sorry for taking it so long, but I barely can find the time to review all the PRs on Watir :)

titusfortner:
Firstly, definitely no worries on keeping up, I think I'm done with PRs for a while, at least major ones. I want to start focusing on third party gems now.

I agree that we don't want to make things too confusing. I also agree that `:text_content` is likely not a "good" way to locate things. 

My logic is basically:

1. If we want to make things consistent, then we need to throw deprecation warnings before changing what `:text` does.
2. If we deprecate the current behavior of `:text`, then we should provide an alternative for people to continue to get the current behavior

Another thought I had is that we could get `text: Regexp` to not care about visibility- same behavior as `text: String`, and then require people to specify `visible: true` if they care about visibility... But I think that should be the default, and `visible: false` is different from `visible: :ambivalent`. :)

It's possible that I'm getting overeager as I'm also trying to figure out a way to support faster text processing with nokogiri. I think I'll hold off on this PR for Watir for a bit longer and think it through again. Thanks for the feedback. 

@jkotests any additional thoughts? Maybe we discuss this & the 'giri implementations in person in a few weeks. :)

jkotests:
@titusfortner , I'm starting to lean towards the idea that the `:text` locator should not care about visibility. The locator has been like that for the longest time. As well, most uses of `text: Regexp` started to ignore visibility when we started converting them to XPath. 

It'll be a bit weird that `Element#text` will be different, but I don't think that's unreasonable.

titusfortner:
Let's discuss this at the [hackathon](https://www.eventbrite.com/e/ruby-open-source-testing-code-hackathon-tickets-36985700225). I'd really like for the method and the locator to be consistent in Watir 7, even if it isn't backward compatible for reasons I can elaborate on more later.

titusfortner:
It was decided at the hackathon that this is the opposite of the approach that should be taken.

PR #632, opened 2017-08-22
implement Element#inner_text

titusfortner:
So, this is a way to retrieve the text from the DOM without concerning if the text is visible on the page.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -302,8 +302,8 @@ module Watir
     # Returns outer (inner + element itself) HTML code of element.
     #
     # @example
-    #   browser.div(id: 'foo').outer_html
-    #   #=> "<div id=\"foo\"><a href=\"#\">hello</a></div>"
+    #   browser.div(id: 'shown').outer_html
+    #   #=> "<div id=\"shown\"><div id=\"hidden\" style=\"display: none;\">Not shown</div><div>Not hidden</div></div>"
     #
     # @return [String]
     #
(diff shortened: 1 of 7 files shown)


lib/watir/elements/element.rb
```diff
@@ -328,6 +328,20 @@ def inner_html
...
+    #
+    # @example
+    #   browser.div(id: 'hidden').inner_text
+    #   #=> "Not shown"
+    #
+    # @return [String]
+    #
+
+    def inner_text
+      element_call { execute_js(:getInnerHtml, @element) }.strip
```
jkotests: This should probably be calling the new JS snippet - ie `:getInnerText`?

titusfortner: Which is when you wonder how the specs are passing... Sigh. Thanks :-)

p0deje (approved):

jkotests:
When I was trying the `#inner_text` method, I was surprised that for:

~~~~~~~~html
<html>
  <body>
    <div>
      <span style="display: none;">a</span> <span>b</span> 
    </div>
  </body>
</html>
~~~~~~~~

I got:

~~~~~~~~ruby
browser.div.inner_text
#=> "b"   #--> I expected to get all text nodes - ie "a b"
~~~~~~~~

Apparently there are a number of things that `innerText` doesn't include (see http://perfectionkills.com/the-poor-misunderstood-innerText).

This got me wondering 2 things:

1. Is `innerText` actually what we want? Would `textContent` be more appropriate? (I'm not a JavaScript person, so I don't know if `textContent` introduces other issues. It at least returns "a b" for the above example.)
2. Is `#inner_text` the best method name? While it does match the JavaScript method, it might be confusing for Watir users. From the method names, it's not clear when to call `#text` vs `#inner_text`. I don't have a good proposal, but I feel it would be more accurate to have `#visible_text` and `#text` (all text nodes). However, I think we want people to normally check the visible text, so maybe we should do:
    * `#visible_text` with alias `#text`
    * `#all_text`

titusfortner:
Ok, fixed method and added the right tests for this.

After running the tests it is obvious that for the applications of the methods we want to use in #631 is `textContent` not `innerText`. My take on the naming of these at this point is to directly match them to their js equivalents. These aren't things Watir users have had access to before, so it makes sense to name them the same as what the documentation out in the world says they will do.

As for aliasing them to be more Watiry... I'm honestly not seeing much use for `#inner_text` at all now, since that is what Selenium's `#text` method is already approximating, though I don't see much downside to getting rid of it, either... :) And I actually think that `#text_content` is probably better than `#all_text`. And/or this is probably the right behavior to use for `#text!` if we want to use that method. :)

I'm going to merge this once Travis gives the thumbs up, make a PR for `#text_content` and then we can discuss any aliases we want as part of #631 before the 6.8 release.

PR #633, opened 2017-08-23
Text content

titusfortner:
Adds textContent JS support with `Element#text_content`
Moves all the js scripts into snippet files
pulls out all references to javascript execution into a separate module to be included by Elements class just to decrease the file size a bit since it is continuing to get out of hand. :-D


titusfortner:
@p0deje it looks like doctest doesn't work wit methods that are included in Modules? Or am I doing the doctest wrong? These passed before I moved it into another file.

p0deje:
They should work just fine. Maybe you need to tweak hooks in place?

PR #637, opened 2017-09-01
allow selector builder and locator class to interact with selector va…

titusfortner:
…lues independently

Is it ok if these are decoupled, or do you think changes should be reflected across classes?


p0deje (approved):

titusfortner:
merged

PR #638, opened 2017-09-05
add html for angular specs

titusfortner:
is there an easy way to put these in the subproject instead? I couldn't figure it out.


p0deje:
There is definitely a way to do that, but I guess you'll need to allow setting custom paths for Watirspecs. I can try to do that later today, but not sure when.

titusfortner:
It isn't urgent, but I'll eventually need to figure it out for the specs for: https://github.com/titusfortner/watir_angular

I had to hack into my `.rbenv` directory to test it locally. :-D

titusfortner:
I'm going to go ahead and just add these for now. If it makes sense to move them to a sub-project, I can do that later.

PR #639, opened 2017-09-05
get specs passing for chromedriver 2.32

titusfortner:
Is this actually desired behavior? "does not raise an error if the file does not exist"
w3c doesn't say anything about it one way or another.

Alternative is to specify that this spec is not compliant on Chrome instead of deleting it.


p0deje (approved):

titusfortner:
@p0deje - heh, I decided to make it `not_compliant_on` instead of deleting just before you approved it.

PR #641, opened 2017-09-05
Filter elements

titusfortner:
This should be a performance bump for anyone using `index: 0`, and the abstractions I need to be able to filter Nokogiri elements in a subclass for Watigiri.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -98,8 +98,7 @@ module Watir
             # could build xpath/css for selector
             if idx && idx != 0 || !visible.nil?
               elements = locate_elements(how, what)
-              elements = elements.select { |el| visible == el.displayed? } unless visible.nil?
-              elements[idx || 0] unless elements.nil?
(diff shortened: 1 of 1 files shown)


p0deje:
Seems good, but please tell me what is the point of passing index: 0 selector?

p0deje (approved):

titusfortner:
I don't know, but our specs do it everywhere. :)
Maybe our specs wanted to force using `#locate_elements` or `#wd_find_by_regexp_selector` with `:select`? But that seems silly (and means extra wire calls)

titusfortner:
merged

titusfortner:
Actually, the use case for this is to be able to define something like this:

```ruby
element(:foo) { |index| browser.div(class: 'foo', index: index) }
first = foo(0)
```

Instead of needing to interact with an element collection:
```ruby
elements(:foos) { browser.divs(class: 'foo') }
first = foos[0]
```

This has advantages in Page Object code that would allow for the extra method support that comes with assigning a keyword to an element directly.

PR #643, opened 2017-09-05
Select exact matches when using Select#select_by! with String

jkotests:
When using `#Select#select!` and `#Select#select_all!` with a `String` parameter, options are selected by partial matches. They should be selected by exact matches.

For example, in:
~~~~~~~~html
<select id="mySelect" size="4" multiple="multiple">
  <option value="a">Apple</option>
  <option value="p">Pear</option>
  <option value="b">Banana</option>
  <option value="o">Orange</option>
</select>
~~~~~~~~

Multiple get selected when only one matches the specified value:

~~~~~~~~ruby
b.select.select!('a')
#=> "Apple"

b.select.selected_options.map(&:text)
["Apple", "Pear"]
~~~~~~~~

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -162,7 +162,7 @@ module Watir
     def select_by!(str_or_rx, number)
       js_rx = case str_or_rx
               when String
-                str_or_rx
+                "^#{str_or_rx}$"
               when Regexp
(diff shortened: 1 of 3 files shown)


titusfortner:
part of me wants to just not support regex for `#select!`, but that seems like a big limitation.

PR #646, opened 2017-09-05
don't run after hooks when alert is present

titusfortner:
Geckodriver has changed the default Firefox implementation.
Calling `browser.window.present?` no longer closes an alert if present, which was the problem when we discussed error checkers back in #283 

If we are ok actually deprecating Legacy Firefox support with AfterHooks, this will make a lot of things significantly easier.


p0deje (approved):

titusfortner:
merged

PR #650, opened 2017-09-08
#visible? and #enabled? wait for element to exist

jkotests:
Wait for the element to exist before calling `#visible?` and `#enabled`.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -399,7 +399,7 @@ module Watir
     #
 
     def visible?
-      element_call(:assert_exists) { @element.displayed? }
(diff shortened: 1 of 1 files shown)


titusfortner:
Oh, I remember what is going on here. Predicate methods should not wait. They should immediately return their result, regardless.

So the right fix is to add it to the wait method, not the predicate method.

titusfortner:
I'm doing a before/after with stats turned on:
https://travis-ci.org/titusfortner/watir/builds/273495292
https://travis-ci.org/titusfortner/watir/builds/273495586

titusfortner:
6.8.4 should fix this. thanks for your help with it.

PR #656, opened 2017-09-22
Suggest looking in an iframe when element is not found and iframes ex…

bret:
…ist. Paired with @jkotests and @vcabrilo

Let's discuss as part of the Ruby Hackathon


titusfortner (approved):
I think this is a useful addition. I like it.

titusfortner:
I had the thought of also possibly checking inside the iframes for the actual element, but that gets complicated with recursion and such. I think this is probably the right approach.

titusfortner (changes requested):
I changed my mind because of conversation with @jkotests. This should get rescued in `Element#element_call` so that you don't have to worry about `#assert_exists` eating the error and making `@query_scope.iframes` calls unnecessarily.

PR #666, opened 2017-10-04
move locator_namespace to be Browser instance scoped

akostadinov:
Hello, I was looking at [titusfortner/watir_angular](https://github.com/titusfortner/watir_angular) project. It is one project
that adds additional selectors to Watir. The issue with it is that it
does it globally by using `Watir.locator_namespace`.

In my view, much better would be if user can configure browser per instance
instead of globally. There might be spawned different browsers for different
purposes within one test suite so having this ability would allow it.

One browser could be plain Watir, one can be WatirAngular and another might be
something else.

Eventually we could have element locators be more flexible like allowing search
within multiple namespaces. But I thought simply moving the namespace to be
instance specific instead of class global would be a nice first step and
probably all we need in the medium term.

This is btw needed for titusfortner/watir_angular#2 as a first user.

### DIFF
--- a/lib/watir/locators.rb
+++ b/lib/watir/locators.rb
@@ -57,11 +57,11 @@ module Watir
       end
 
       def build_locator
-        query_scope.send :ensure_context
+        @query_scope.send :ensure_context
 
         element_validator = element_validator_class.new
-        selector_builder = selector_builder_class.new(query_scope, @selector.dup, element_class.attribute_list)
-        locator_class.new(query_scope, @selector.dup, selector_builder, element_validator)
+        selector_builder = selector_builder_class.new(@query_scope, @selector.dup, element_class.attribute_list)
+        locator_class.new(@query_scope, @selector.dup, selector_builder, element_validator)
       end
     end
   end
--- a/lib/watir/locators.rb
+++ b/lib/watir/locators.rb
@@ -26,23 +26,23 @@ module Watir
   module Locators
     module ClassHelpers
       def locator_class
-        class_from_string("#{Watir.locator_namespace}::#{element_class_name}::Locator") ||
+        class_from_string("#{browser.locator_namespace}::#{element_class_name}::Locator") ||
             class_from_string("Watir::Locators::#{element_class_name}::Locator") ||
-            class_from_string("#{Watir.locator_namespace}::Element::Locator") ||
+            class_from_string("#{browser.locator_namespace}::Element::Locator") ||
             Watir::Locators::Element::Locator
       end
 
       def element_validator_class
-        class_from_string("#{Watir.locator_namespace}::#{element_class_name}::Validator") ||
+        class_from_string("#{browser.locator_namespace}::#{element_class_name}::Validator") ||
             class_from_string("Watir::Locators::#{element_class_name}::Validator") ||
-            class_from_string("#{Watir.locator_namespace}::Element::Validator") ||
+            class_from_string("#{browser.locator_namespace}::Element::Validator") ||
             Watir::Locators::Element::Validator
       end
 
       def selector_builder_class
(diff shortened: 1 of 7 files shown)


akostadinov:
Sorry, for some reason I didn't get notification that tests failed. I don't think errors in travis or appveyor are caused by my changes. Chrome spec in travis passed (as well for me locally). I can't tell why `firefox` failed. I see 
```
 invalid value for expiration date: "292277026596-12-04 15:22:30.000000000 +00:00"
```
But I can't see how possibly my changes could have caused this.

titusfortner:
@akostadinov does this PR matter any more now that WatirAngular doesn't actually use `#locator_namespace`? 

There are only 3 projects that use this method so far as I know: watizzle (which no one uses), watir_css (no one uses) and watigiri which uses default locators unless specifically called for (and also no one may use).

akostadinov:
Any global config is evil. This PR is not important to me as presently functionality is not used.

But if locator namespace is to be used by future extensions, then it makes sense to make it browser specific from project standpoint of view.

titusfortner:
@p0deje any strong feelings on this one either way? I kind of agree with minimizing globals, but not sure I see the actual advantage to making it an instance variable..

also @akostadinov is there a reason you wanted `@query_scope` to be private? Doesn't make sense to me to provide reader access to a private instance variable.

p0deje:
@titusfortner @akostadinov Happy to merge this as soon as conflicts are resolved. Great contribution!

akostadinov:
/retest

For me locally chrome and ff tests worked. I don't think AppVeyor tests failed because of my changes. 

```
                Errno::ECONNREFUSED:
                  Failed to open TCP connection to 127.0.0.1:4444 (No connection could be made because the target machine actively refused it. - connect(2) for "127.0.0.1" port 4444)
```
and 
```
Failed to open TCP connection to 127.0.0.1:4444 (No connection could be made because the target machine actively refused it. - connect(2) for "127.0.0.1" port 4444)
                # ./lib/watir/window.rb:225:in `assert_exists'
                # ./lib/watir/window.rb:108:in `exists?'
```

lib/watir/locators.rb
```diff
@@ -26,23 +26,23 @@ module Watir
   module Locators
     module ClassHelpers
       def locator_class
-        class_from_string("#{Watir.locator_namespace}::#{element_class_name}::Locator") ||
+        class_from_string("#{query_scope.locator_namespace}::#{element_class_name}::Locator") ||
```
titusfortner: @akostadinov we don't have `#locator_namespace` defined for `Watir::Element`, which is a valid query_scope instance. I think for this we want to hard code this to `browser`, and keep query_scope as an instance variable.

akostadinov:
@titusfortner , thank you for the excellent point! I updated the file to use `query_scope.browser.locator_namespace` because I don't see `#browser` defined for `lib/watir/element_collection.rb` which also includes `Locators::ClassHelpers`.

I think this PR needs a test case or two for this functionality before merging though. Please let me know if you have any further concerns about the implementation.

titusfortner:
Ah, yes, I think we should copy over `#browser` to `ElementCollection` then.
As for specs... I'll just make sure it works with `watir_css` before we release it. :)

akostadinov:
Added `#browser` method to ElementCollection. I prefer to keep `query_scope` as a method especially because it is used across module boundaries (by the `#build_locator` method). I generally prefer getters to direct instance variable use as they make adapting to change easier.

But if in watir project direct access is preferred, please tell me. I don't see any technical obstacles getting back to instance variables.

titusfortner:
I'm going to request we not change it because we don't manage any other private instance variables with a `private :foo` and it isn't like `query_scope` is ambiguous as to what it is across modules.

akostadinov:
@titusfortner , done.

akostadinov:
Thank you!

PR #668, opened 2017-10-05
Allow setting a cookie expire value with a String

lmtierney:
Selenium expects the expires value to be that of [`Time`, `DateTime`, or `Numeric`](https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/options.rb#L122). For `Cookies#load` to work, the expires value needs to be converted to one of those types.

I also added a guard for the a [bug I added for geckodriver](https://github.com/mozilla/geckodriver/issues/1000) as currently Session cookies are returned with a bogus number for expiry.

### DIFF
--- a/lib/watir/cookies.rb
+++ b/lib/watir/cookies.rb
@@ -59,7 +59,10 @@ module Watir
         value: value}
       cookie[:secure] = opts[:secure] if opts.key?(:secure)
       cookie[:path] = opts[:path] if opts.key?(:path)
-      cookie[:expires] = opts[:expires] if opts.key?(:expires)
+      expires = opts[:expires]
+      if expires
+        cookie[:expires] = ::Time.parse(expires) if expires.is_a?(String) else expires
+      end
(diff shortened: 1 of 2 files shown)


lib/watir/cookies.rb
```diff
@@ -59,7 +59,10 @@ def add(name, value, opts = {})
         value: value}
       cookie[:secure] = opts[:secure] if opts.key?(:secure)
       cookie[:path] = opts[:path] if opts.key?(:path)
-      cookie[:expires] = opts[:expires] if opts.key?(:expires)
+      expires = opts[:expires]
+      if expires
+        cookie[:expires] = ::Time.parse(expires) if expires.is_a?(String) else expires
```
titusfortner, line 64: Prefer the trinary operator for single line conditionals:
`cookie[:expires] = expires.is_a?(String) ? ::Time.parse(expires) : expires`

PR #669, opened 2017-10-05
Allow the headless option to work with Firefox

lmtierney:
Similar approach to what is already happening with Chrome


lmtierney:
Cookie failures on Travis are due to https://github.com/mozilla/geckodriver/issues/1000 addressed in #668

Issue #651, opened 2017-09-15
Newly created set! function is not accepting the multiple argument


Rajagopalan-M:
## Meta -
Watir Version:  6.8.4
Selenium Version: 3.5.2
Browser Version: FireFox 52

## Expected Behavior -
 It has to pass two arguments one by one 

## Actual Behavior -
It's throwing error

## Steps to reproduce -

b.text_field(name: 'q').set!('Auto',:tab)

This is the expected behavior because set function is accepting multiple argument. I have to pass tab key along with the given text. I understand that set! function passes the first and last key through selenium to make sure all the actions are being triggered. But what it actually has to do is, it needs to send the last character of last argument through selenium.

PR #670, opened 2017-10-12
Support multiple parameters in UserEditable#set!

jkotests:
Depending on your expectations, fixes or partially fixes, #651.

The changes:
* Allow `#set!` to accept multiple parameters.
* Raises an exception if one of the parameters is a `Symbol` (ie special key). While the example in #651 included `:tab`, I am not convinced that `#set!` should support it.

### DIFF
--- a/lib/watir/user_editable.rb
+++ b/lib/watir/user_editable.rb
@@ -22,7 +22,9 @@ module Watir
     # @param [String] input_value
     #
 
-    def set!(input_value)
+    def set!(*args)
+      raise ArgumentError, "#set! does not support special keys, use #set instead" if args.any? { |v| v.kind_of?(::Symbol) }
+      input_value = args.join
       set input_value[0]
       element_call { execute_js(:setValue, @element, input_value[0..-2]) }
       append(input_value[-1])
--- a/spec/watirspec/elements/text_field_spec.rb
+++ b/spec/watirspec/elements/text_field_spec.rb
@@ -297,6 +297,16 @@ describe "TextField" do
       expect(browser.text_field(name: "new_user_occupation").value).to eq "ĳĳ"
     end
 
+    it "sets the value to a concatenation of multiple arguments" do
+      browser.text_field(id: 'new_user_email').set('Bye', 'Cruel', 'World')
+      expect(browser.text_field(id: "new_user_email").value).to eq 'ByeCruelWorld'
+    end
+
+    it "sets the value to blank when no arguments are provided" do
+      browser.text_field(id: 'new_user_email').set
+      expect(browser.text_field(id: "new_user_email").value).to eq ''
+    end
+
     it "raises UnknownObjectException if the text field doesn't exist" do
       expect { browser.text_field(id: "no_such_id").set('secret') }.to raise_unknown_object_exception
     end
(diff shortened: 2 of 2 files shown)


Rajagopalan-M:
@jkotests 

why set should not support for :tab? Set is simply calling send_keys at the back,right? Send_keys job is to send whatever keys I sent,right?

jkotests:
@Rajagopalan-M, I assume you are asking about `#set!` rather than `#set`.

`#set!` uses `#send_keys` to simulate typing the first character, as well as actually typing the last character. The rest of the characters are inputted by executing JavaScript. The JavaScript execution is faster than `#send_keys`, but at the cost of not being realistic (ie might not trigger all events).

Supporting `:tab` (and other special keys) would mean creating some sort of JavaScript executor to map/simulate the key press. Alternatively, we could call `#send_keys` for the special keys and use the JavaScript execution for the rest. Then you'd probably need to add some special logic for handling which keys are the first/last to simulate.

Ultimately, it comes down to, I don't understand the use case for wanting to input special keys using `#set!` over `#set`. Without understanding the reason, I didn't want to add unnecessary code. If you can help me understand the use case, I'd be happy to update the changes.

Rajagopalan-M:
@jkotests 

`#set! uses #send_keys to simulate typing the first character, as well as actually typing the last character. The rest of the characters are inputted by executing JavaScript. The JavaScript execution is faster than #send_keys, but at the cost of not being realistic (ie might not trigger all events).`

Yes, I understood this well.


`Supporting :tab (and other special keys) would mean creating some sort of JavaScript executor to map/simulate the key press. Alternatively, we could call #send_keys for the special keys and use the JavaScript execution for the rest. Then you'd probably need to add some special logic for handling which keys are the first/last to simulate.`

yes, that's what my idea as well, send the last parameter through the send_keys would resolve the problem easily. 

for an example, 

If someone writes the following code,

b.text_field(id: 'something').set("raj",:tab)

then r and :tab has to be sent through send_keys and the rest of the characters we can pass through javascript. 

`Ultimately, it comes down to, I don't understand the use case for wanting to input special keys using #set! over #set. Without understanding the reason, I didn't want to add unnecessary code. If you can help me understand the use case, I'd be happy to update the changes.`

My point is, WATIR is paving the way to write the code through selenium and Javascript, So why not the same code which has been written in Javascript works like Selenium equivalent? So WATIR maintains the stability for both selenium and Javascript, someone who writes purely from javascript can't have this stability as he is writing from WATIR. 

I wanted to state my another problem, it's select list, 

when you write the following code

b.select_list(id: 'something').select "Ruby"

It is actually clicks the option by using this selenium equivalent
`
driver.find_element(xpath: "//normalize-space()='Ruby']").click # this is not a complete code`

but when we use Javascript select, 
`b.select_list(id: 'something').select! "Ruby"`

It's actually setting the option, So it's not triggering all the actions associated with click, So it's not giving the selenium equivalent of selecting the option. So can we stimulate the click through Javascript anyway? It would lead us to write WATIR code completely in Javascript !

titusfortner (approved):

titusfortner:
@jkotests this looks great, you have merging powers, right?

@Rajagopalan-M submit a pull request with the functionality you want and we'll discuss it.

jkotests:
@titusfortner, sorry, I don't have merging powers.

@Rajagopalan-M, while I agree we could do that, my question is still what problem would we be solving? At the moment, I do not see why you would call `.set!("raj", :tab)` over `.set("raj", :tab)`. If we add support, we'd have to decide what keys are simulated by `browser.text_field.set!(:tab, 'a', :numpad0, 'bc', :tab)`. Again, I would rather see a specific use case before writing code that may never be used.

titusfortner:
Ah, I thought you did. I won't be back in the code until later this week. Sorry for the delay. Make sure you get enough PRs in to get your t-shirt, though. :)

Rajagopalan-M:
@jkotests Why set! function is introduced in WATIR while set function is available? Just to speed up and give the support through Javascript,right? If so, why can't we bring to the level of set function? Set function does certain task but set! function is not doing it, So I am asking you to bring set! function to the level of set. 

I said last parameter last word we could pass through selenium, and the rest we could pass through Javascript. For an example, in the given example,

 browser.text_field.set!(:tab, 'a', :numpad0, 'bc', :tab)

We need to pass first :tab and last :tab via selenium send_keys and rest has to go via JavaScript  .

titusfortner:
@Rajagopalan-M feel free to make a PR if this does not satisfy what you need.

Issue #678, opened 2017-10-31
WATIR API is not completely allowing to use Selenium Locators


Rajagopalan-M:
To click a link WATIR API gives the following code,

`b.link(text: 'something').click`

but this is not utilizing Selenium 'link' locator, it is using 'xpath' locator of selenium at the back, the above code is getting converted in selenium

`driver.find_element(xpath: "//a[normalize-space()='something']").click`

So WATIR API is not giving any way to utilize the Selenium locator 'link', infact I can write the below code to invoke Selenium locator link

`b.element(link: 'something').click`

this will allow WATIR to write the following code in selenium

`driver.find_element(link: 'something').click`

but Titus Fortner feels that I am not using WATIR API in this place, Said that, WATIR can do following the arrangement,

`b.link(link: 'something').click`

but this code seems to be meaningless to me because 'b.link' is explicitely telling me that that is the link, So I am suggesting the following way 

we have two types of code to access link

```
b.link()
b.a()
```

but these two lines of code are doing the same Job, So we could use one of them for 'link' locator and another one for 'xpath'

when I write 

`b.link(text: 'something').click`

It needs to be converted into

`driver.find_element(link: 'something').click`

and when I write 

`b.a(text: 'something').click`

it needs to be converted into

`driver.find_element(xpath: "//a[normalize-space()='something']").click`

What's on your thoughts @jkotests ? @titusfortner ?

jkotests:
@Rajagopalan-M, I disagree with supporting the `:link` locator for `#link` or `#a`. I actually disagree that `:link` should be supported in `#element`.

It's just not a locator that makes sense to me:
* Why is there a `:link` locator, but there isn't a similar `:button` locator?
* It also breaks the concept that passing in a `String` value does exact matches while a `Regexp` value does partial matching. Last I checked, using a `Regexp` will never return anything unless the element has a matching `link` attribute. Instead you have to use the `:partial_link_text` locator.

That said, I would be willing to support the underlying use case - ie matching based on the visible text. The support should be for all element types, not just links.

Note that we had talked about doing this in #631, but ended up deferring such work until a later time (ie someone requested the feature).

titusfortner:
So the issue is that we've decided to have text locator find things that aren't displayed, where link text locator will only find displayed. We don't want to break backward compatibility. We could allow link_text or partial_link_text as locators for a link element in addition to a generic element, but that also gets weird with how regular expressions would work in this context.
[...]

PR #685, opened 2017-11-18
Selenium's link_text and partial_link_text locators

titusfortner:
Based on the conversation in #678 

This allows `link_text` and `partial_link_text` to be used in conjunction with other locators just like all other selenium location methods. `String` and `RegExp` values are supported. It works for `#element` and `#link` types only, and otherwise will throw a `TypeError`.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -36,7 +36,7 @@ module Watir
         end
 
         def locate
-          e = by_id and return e # short-circuit if :id is given
+          e = by_se_locator and return e # short-circuit if selenium locator is given
 
           element = if @selector.size == 1
                       find_first_by_one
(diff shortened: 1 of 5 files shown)


titusfortner:
I would have sworn I had this working before, but now it isn't...

I can get it working if `:link_text` or `:partial_link_text` values are `RegExp`, or if they are the only locator. I can't seem to get it to work if they are `String` and part of multiple locators

Any ideas on how to approach this better?

If we can't get this to work, part of me thinks we should get rid of supporting those two locator strategies entirely in Watir since it doesn't fit in to our overall locator strategy.

jkotests:
I would still prefer to have a generic visible text locator - see #688 for my attempt.

I find it weird that we'd have `link_text: /text/` and `partial_link_text: 'text'`, both of which would be expected to return the same thing.

jkotests:
I was able to get the changes working by doing the following:
* Have Element::SelectorBuilder#build_wd_selector bail if the selector includes :link_text or :partial_link_text. Currently, these locators get include as XPath attributes when using String values.
* Have Element::Locator#delete_regexps_from include :link_text and :partial_link_text in the rx_selector.
* Have Element::Locator#matches_selector? convert String values for :partial_link_text to a Regexp.

p0deje (approved):

titusfortner:
@jkotests, Let's go ahead and add this implementation onto #688, rather than accepting this as-is.
As part of it, I want to deprecate these two locators and tell users to use `:visible_text` instead. We'll continue to support it until Watir 7, but I want to put in the deprecation notice.

PR #686, opened 2017-11-18
make text location equivalent between String and RegExp values

titusfortner:
At the Hackathon in Toronto we (@bret, @tk8817, @jkotests & others) discussed at length the proper way of addressing #342.  This was not the solution I proposed, but I was persuaded that it makes the most sense for consistency and backward compatibility based on the unique considerations of text nodes.

I'm not sure this is the best implementation of our idea, so any feedback on it would be appreciated.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -154,7 +154,7 @@ module Watir
         def fetch_value(element, how)
           case how
           when :text
-            element.text
+            Watir::Element.new(@query_scope, element: element).send(:execute_js, :getTextContent, element).strip
           when :tag_name
             element.tag_name.downcase
           when :href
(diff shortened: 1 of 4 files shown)


lib/watir/locators/element/locator.rb
```diff
@@ -154,7 +154,7 @@ def wd_find_all_by(how, what)
         def fetch_value(element, how)
           case how
           when :text
-            element.text
+            Watir::Element.new(@query_scope, element: element).text_content
```
jkotests: Would it be bad to do:

    Watir::Element.new(@query_scope, element: element).send(:execute_js, :getTextContent, element).strip

It adds a bit of code duplication, but it would save an extra wire call (ie waiting for the element to exist even though we just retrieved the element).

p0deje (approved):

lib/watir/locators/element/locator.rb
titusfortner: I agree this is better. I'll add it.

titusfortner:
merged 37a1b22

Issue #681, opened 2017-11-08
Some attributes are considered 'invalid'


enkessler:
Although it is nice to finally be able to locate an element based on whether or not it has an attribute, Watir seems to have a limited view of what is an attribute. If I can inspect an element and see an attribute on it in the DOM, then I expect to be able to use that attribute, as is, to locate that element. Hyphenated attributes, in particular, seem to be a problem for Watir.

## Meta -
Watir Version: 6.7.3
Selenium Version: 3.7.0
Browser Version: Chrome 62
Browser Driver Version: chromedriver 2.33
OS Version: OSX

## Expected Behavior -

I can locate elements based on arbitrary element attributes.

e.g.

`page.div_elements('sidebar-close' => true)`

would successfully locate the element

`<div id="foo" class="stuff" sidebar-close="main"">`

## Actual Behavior -

`Watir::Exception::MissingWayOfFindingObjectException: invalid attribute: "sidebar-close"`

titusfortner:
Watir only supports valid html attributes based on the specification. You can add custom attributes by prepending `data-` and referencing it with `data_custom_attribute: 'foo'`

enkessler:
Could you elaborate on that?

None of these work to locate that element.

`page.div_elements('data-sidebar-close' => true)  # Returns invalid attribute error`
`page.div_elements('data_sidebar-close' => true)  # Returns no elements`
`page.div_elements(data_sidebar_close:  true)       # Returns no elements`

titusfortner:
Can you show the html?

titusfortner:
Also, it would be useful to not include page object gem and use watir code directly.

jkotests:
Adding "data-" would mean needing to modify the HTML to:

~~~~~~~~html
<div id="foo" class="stuff" data-sidebar-close="main"">
~~~~~~~~

If you don't have the option to modify the HTML, you can also [add to the list of valid attributes](https://jkotests.wordpress.com/2014/12/15/adding-custom-attributes-to-the-list-of-valid-locators/). Note that you will need to use underscores instead of dashes.

~~~~~~~~ruby
# adding to the attribute list
Watir::HTMLElement.attributes << :sidebar_close

# will allow you to do
browser.div(sidebar_close: 'main').exists?
~~~~~~~~

enkessler:
I'm trying to locate the following div (slightly sanitized approximation).

`<div class="stuff" sidebar-close="main">`

[...]

PR #687, opened 2017-11-18
allow location of elements using custom attributes

titusfortner:
This is an example of how #681 can be addressed

Ideally people use HTML 5 compliant attributes, but I'm not sure there's a good reason to enforce that. This code would just record the presence of non compliant attributes and make note of it in the error message if it isn't found.

Creating methods is different. If they want to get the value of a non-compliant attribute, then they can do: `my_element.attribute_value('custom_attribute')` without providing support for `my_element.custom_attribute`.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -587,9 +587,9 @@ module Watir
 
       element_validator = element_validator_class.new
       selector_builder = selector_builder_class.new(@query_scope, @selector.dup, self.class.attribute_list)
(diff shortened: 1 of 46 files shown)


jkotests (changes requested):
`MissingWayOfFindingObjectException` no longer appears to be used. We should remove it from exceptions.rb.

p0deje (approved):

titusfortner:
Good call. Removed.

jkotests (approved):

PR #688, opened 2017-11-19
Locate elements by visible text

jkotests:
Based on the conversation in #678.

This adds a generic :visible_text locator for the purpose of supporting the use case provided by Selenium's built-in :link_text and :partial_link_text locators. This adds support across all element types - `#element`, `#link`, `#div`, etc. String and RegExp values are supported. 

Performance will be better for links if the :visible_text locator directly uses Selenium's :link_text and :partial_link_text. I haven't added the short-circuit yet. I figured we should agree on the generic approach before adding performance optimizations.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -91,22 +91,23 @@ module Watir
 
           idx = selector.delete(:index) unless selector[:adjacent]
           visible = selector.delete(:visible)
+          visible_text = selector.delete(:visible_text)
 
           how, what = selector_builder.build(selector)
 
           if how
             # could build xpath/css for selector
-            if idx && idx != 0 || !visible.nil?
+            if idx && idx != 0 || !visible.nil? || !visible_text.nil?
               elements = locate_elements(how, what)
-              filter_elements elements, visible, idx, :single
+              filter_elements elements, visible, visible_text, idx, :single
             else
               locate_element(how, what)
             end
           else
             # can't use xpath, probably a regexp in there
-            if idx && idx != 0 || !visible.nil?
+            if idx && idx != 0 || !visible.nil? || !visible_text.nil?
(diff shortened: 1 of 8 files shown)


lib/watir/locators/element/locator.rb
```diff
@@ -155,6 +157,8 @@ def fetch_value(element, how)
           case how
           when :text
             element.text
+          when :visible_text
```
titusfortner, line 160: This can be done with `when :text, :visible_text`

spec/watirspec/html/non_control_elements.html
```diff
@@ -131,5 +131,10 @@ <h6 id="empty_header"></h6>
       <del class="footer" onclick="this.innerHTML = 'This is a del with text set by Javascript.'">This is a del.</del>
     </div>
 
+		<div id="visible_text">
+			<a id="all_visible_text">all visible</a>
+			<a id="some_visible_text">some visible<span style="display:none;"> some hidden</span></a>
```
titusfortner, line 136: It might be more clear if we did:
```ruby
<a id="some_visible_text">some <span style="display:none;">some hidden</span>visible</a>
```

titusfortner (changes requested):
Yes, I like this better than supporting `:link_text` and `:partial_link_text`.

lib/watir/locators/element/locator.rb
jkotests: Won't :text and :visible_text be different once we merge with #686?

titusfortner: lol. Yes. Disregard.

titusfortner:
I'm going to just rebase and merge this as is so I can get another release out today. :)

titusfortner:
Merged: https://github.com/watir/watir/commit/60b9d10393eb372849f2557f10a87fa6fb46de7b

PR #689, opened 2017-11-22
Se locators

titusfortner:
I find this code overall much more straightforward than what was there, but let me know what doesn't make sense in this re-write.

1. gets rid of `#find_*_by_one` methods since those were only useful for selenium selectors and it made for weird naming conventions, as well as duplicating some code with `#by_id`

2. Removes code duplication between `#find_first_by_multiple` and `#find_all_by_multiple`

2. It treats all selenium selectors like we were treating `:id` with `#by_id`

3. Supports using selenium selectors for `#locate_all` as well `#locate`

4. Finding a single element with `:tag_name` + `:css` or `:xpath` no longer just validates the tag name of the first element found, it checks if the first found matches the tag name in `#using_selenium`, but then will go into `#using_watir` to filter `find_elements` with the `:tag_name` to pull out what matches.

5.  `Element::Locator` subclasses have different things to override. `#wd_find_first_by` and `#by_id` were often just duplicating code from `Element::SelectorBuilder` subclasses, now it is just necessary to force `#using_watir` and it all just works

6. `:link_text`, `:partial_link_text` and `:link` do not fit in with the rest of the Watir locating system. You can't currently mix them with other locators and you can't pass in regular expressions. So I want to deprecate them in favor of the code that @jkotests just wrote with `:visible_text` which does support all of those things. I say we support them until Watir 7 when we have **the great deprecation purge of 2018**.

### DIFF
--- a/lib/watir/locators/button/locator.rb
+++ b/lib/watir/locators/button/locator.rb
@@ -2,19 +2,12 @@ module Watir
   module Locators
     class Button
       class Locator < Element::Locator
-        def locate_all
-          find_all_by_multiple
-        end
+
 
         private
 
-        def wd_find_first_by(how, what)
-          if how == :tag_name
-            how  = :xpath
-            what = ".//button | .//input[#{selector_builder.xpath_builder.attribute_expression(:input, type: Watir::Button::VALID_TYPES)}]"
-          end
-
-          super
+        def using_selenium(*)
+          # force watir usage
         end
 
         def can_convert_regexp_to_contains?
--- a/lib/watir/locators/button/validator.rb
+++ b/lib/watir/locators/button/validator.rb
@@ -2,7 +2,7 @@ module Watir
   module Locators
     class Button
       class Validator < Element::Validator
-        def validate(element, selector)
+        def validate(element, _selector)
           return unless %w[input button].include?(element.tag_name.downcase)
           # TODO - Verify this is desired behavior based on https://bugzilla.mozilla.org/show_bug.cgi?id=1290963
           return if element.tag_name.downcase == "input" && !Watir::Button::VALID_TYPES.include?(element.attribute(:type).downcase)
--- a/lib/watir/locators/cell/locator.rb
+++ b/lib/watir/locators/cell/locator.rb
@@ -2,15 +2,13 @@ module Watir
   module Locators
     class Cell
       class Locator < Element::Locator
-        def locate_all
-          find_all_by_multiple
-        end
 
         private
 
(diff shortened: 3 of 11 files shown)


p0deje (approved):

titusfortner:
@jkotests this doesn't implement the shortcut for `:visible_text` in `Anchor`, not sure how important you think that is for the next release.

jkotests:
I don't think the optimizations for :visible_text are needed right away. They are just a nice to have for those who currently use :link_text and :partial_link_text, which we hope isn't that many people. Should definitely do it before we remove their support.

jkotests:
A couple of questions:
* Is this change necessary for 6.10 to be released?
* Would this be a breaking change for the gems that create custom locators (eg Watizzle, Watigiri)?

titusfortner:
I just refactored `watir_angular` to not care about any of the locator extensibility.
The update to `watigiri` is [here](https://github.com/titusfortner/watigiri/commit/0bc33ac53eb67ac1d906ba67b01b9b7156e3ab4e). 
`watir_css` won't be affected because it only has selector builder code
`watizzle` will need updates, but no one is using it in production, and @p0deje says we don't need to support it before pushing this change.

Yes, I'd like for Watir 6.10 to be all the locators commits, so 6.11 can be all the capabilities updates that are needed. :)

I made a stab at using the selenium locators in the Anchor class when `:visible_text` is used, and I couldn't get it to work based on how things are getting filtered out where. I'll have to try again at it later.

titusfortner:
@jkotests did you want to look over this code more? Trying to figure out when I can get the next release out.

jkotests:
@titusfortner, there's quite a lot of code to unpack here. I've only just started understanding how the old locator code worked and am having a hard time working through all the new code paths. The changes are definitely an improvement, but I still think we could do more simplifications.

At any rate, I don't see anything obviously wrong, so I'm fine if you want to push this.

jkotests (approved):

titusfortner:
What more simplifications are you thinking of? I'm fresh off of going through all this code, so now would be a good time to discuss your thoughts. :-D

My process was
1. Create an elegant solution
2. Hack on ugly add-ons until all the tests pass
3. Consolidate ugly hacks into fewer ugly hacks
4. Hack on more ugly add-ons until all the tests pass

Hit me up on Slack if you want to discuss this more.

titusfortner:
This was merged: cf57f11261eb7c47225c7a037552d08c2288fa2c

Issue #660, opened 2017-09-28
element.flash hold the background color red in Firefox instead of blinking


sjieg:
## Meta -
Watir Version: 6.8.4
Selenium Version: 3.6.0
Browser Version: Firefox 55.0.1 (64bit)
Browser Driver Version: geckodriver 0.18.0
OS Version: Windows 7

## Expected Behavior -

The element background to flash multiple times from red to original color

## Actual Behavior -

The element background color stays red

## Steps to reproduce -

Open `irb`
```
require 'watir'
b = Watir::Browser.new :firefox
b.goto 'google.com'
elm = b.div(:id => 'lga')
elm.flash
```

## Note

The behaviour is different for some elements, for example when flashing the input element, it only blinks red once.

```
require 'watir'
b = Watir::Browser.new :firefox
b.goto 'google.com'
elm = b.input(:name=> 'q')
elm.flash
```

When using chrome, the element works just fine:
```
require 'watir'
b = Watir::Browser.new :chrome
b.goto 'google.com'
elm = b.div(:id => 'lga')
elm.flash
```

titusfortner:
Fixed: https://github.com/watir/watir/commit/80d29457f20d0e610afdaa939bd3f35d37ffe503
It'll be part of the next release

PR #695, opened 2017-12-06
Set flash default timeout to 0.01

sjieg:
Reduced the default flash time from 2 seconds to 0.1 second.
Because `.flash` is mostly used to present the automation product to a potential client, 2 seconds delay to flash an element will make the solution look slow. Tested on Firefox 57.

### DIFF
--- a/lib/watir/js_execution.rb
+++ b/lib/watir/js_execution.rb
@@ -45,8 +45,6 @@ module Watir
 
     def flash(color: 'red', flashes: 5, delay: 0.1)
       background_color = style("backgroundColor")
-      background_color = 'white' if background_color.empty?
-      element_color = element_call { execute_js(:backgroundColor, @element) }.strip
 
       (flashes * 2).times do |n|
         nextcolor = n.even? ? color : background_color
@@ -54,7 +52,7 @@ module Watir
         sleep(delay)
       end
 
-      element_call { execute_js(:backgroundColor, @element, element_color) }
+      element_call { execute_js(:backgroundColor, @element, background_color) }
 
       self
     end
--- a/lib/watir/js_execution.rb
+++ b/lib/watir/js_execution.rb
@@ -43,11 +43,10 @@ module Watir
     # @return [Watir::Element]
     #
 
-    def flash(color: 'red', flashes: 5, delay: 0.01)
+    def flash(color: 'red', flashes: 5, delay: 0.1)
       background_color = style("backgroundColor")
       background_color = 'white' if background_color.empty?
       element_color = element_call { execute_js(:backgroundColor, @element) }.strip
-      element_color = 'white' if element_color.empty?
 
       (flashes * 2).times do |n|
         nextcolor = n.even? ? color : background_color
--- a/lib/watir/js_execution.rb
+++ b/lib/watir/js_execution.rb
@@ -43,7 +43,7 @@ module Watir
     # @return [Watir::Element]
     #
 
-    def flash(color: 'red', flashes: 5, delay: 0.2)
+    def flash(color: 'red', flashes: 5, delay: 0.01)
       background_color = style("backgroundColor")
       background_color = 'white' if background_color.empty?
       element_color = element_call { execute_js(:backgroundColor, @element) }.strip


titusfortner:
I just tested this code and it flashes so rapidly 5 times that you can't even see it.

The default settings are exactly how I would expect to see it.

This works if you need it faster
```ruby
my_element.flash(delay: 0.01)
```

@p0deje or @jkotests or @Paul3816547290 do you have any preference for this? If we increase the flashes we'd at least need to up the default number of times it flashes for humans to comprehend it. :)

titusfortner:
I guess I should specify that my use case is to flash the element while I'm scanning the page to see what is flashing when I'm not certain what the element is. My use case is not pointing to an element in a demo to show that it is flashing.

p0deje:
It's not 2s, it's 200ms, ain't it?

The default looks good for me, don't see why it should be changed.

titusfortner:
Yes, the suggestion is to change from `0.2` to `0.01`. The previous default was:
```ruby
flash(color: 'red', flashes: 10, delay: 0)
```
I changed it because even 10 flashes at that speed was too quick for me to see it, and seemed too busy.

jkotests:
I don't use `#flash`, so I don't have a strong preference. Though 0.01 seems too short.

I'm guessing the delay is one of those things that you'll never get universal agreement with. Maybe the better solution is to allow users to set the default.

titusfortner:
The point of having parameters to pass in is to allow them to pass them in. I'd rather not add configuration specifically to tweak defaults on any method. If it were up to me alone, I'd remove the `Watir.default_timeout` configuration. :-)

sjieg:
The flash function was changed because of a bug in firefox: https://github.com/watir/watir/issues/660#event-1348565277
Since watir-webdriver flash has always been an 'animation' of less than a quarter of a second.
Could we aggree on `0.05`?

titusfortner:
That still seems too fast for me to process when I don't know where on the page I'm supposed to be looking. I don't love, but maybe we can do: `flashes: 10, delay: 0.1`?

titusfortner:
Does `color.empty?` not always equal white? If color is empty when it starts, it won't change the page back to the correct color at the end, depending on what is input for color.

sjieg:
Hmm, after some research, it seems that the using `backgroundColor.js` will always return `""`. And `style("backgroundColor")` will always return an RGBA value, e.g `"rgba(34, 34, 34, 1)"`. So:
```
background_color = 'white' if background_color.empty?
```
Will never be true and
```
element_color = 'white' if element_color.empty?
```
Will always be true. Let me try something else...

titusfortner:
After thinking about this more... The two use cases for this are very different. 
1. part of the automation for either demonstrations in real time or to highlight things in video recordings. The goal is to be quick so as not to slow down the automation itself.
2. part of the manual effort when creating automation. The goal is to be slow enough to check to see if what you are looking at is what you want.

Maybe it would make sense to have
`my_element.slow_flash`
or
`my_element.flash :slow`

sjieg:
I was thinking the same, but I thought it might clutter the code a bit being such a minor functionality. If a user repeatedly uses `.flash` with specific parameters, they should just create a custom function in their project.

Anyway, if you're okay with adding a function for this purpose, what do you think of names like `blink` or `pulse`:
- Make `element.flash` default to `flashes: 1, delay: 0.2`
- Create new method `element.pulse` which calls `element.flash(flashes:10, delay: 0.2)`

titusfortner:
@sjieg sorry for leaving this one hanging. 

I took a look at watir-classic for original behavior, looks like it was 10 flashes at .05 seconds
https://github.com/watir/watir-classic/blob/master/lib/watir-classic/element.rb#L192

So, I'm ok with changing to values in this PR, but you also deleted the things that were added with https://github.com/watir/watir/commit/80d29457f20d0e610afdaa939bd3f35d37ffe503, which I thought we needed for a Firefox bug.

sjieg:
I'm unable to push new code, but I deleted some things from 80d2945 because the element color now *always* ends up white. The current PR doesn't fix that, but this should:
```
  def flash(color: 'red', flashes: 10, delay: 0.05)
      background_color = original_color = style("background-color")
      background_color = 'white' if background_color.empty?

      (flashes * 2).times do |n|
        nextcolor = n.even? ? color : background_color
        element_call { execute_js(:backgroundColor, @element, nextcolor) }
        sleep(delay)
      end

      element_call { execute_js(:backgroundColor, @element, original_color) }

      self
    end
```
This fixes both, tested in chrome & firefox:
```
require 'watir'
b = Watir::Browser.new :firefox
b.goto 'watir.com'
b.figure(class: 'highlight').flash
```
- Before PR: flashes slow , background of element ends white
- After PR: Flashes faster, background of element ends the same as it starts (grey)

titusfortner:
Great, thanks for testing it. What do you mean can't push new code?

sjieg:
Quite litteral, when I hover over the edit-icon [here](https://github.com/watir/watir/pull/695/files), it says "You must be signed in and have push access to the patch-1 branch to make changes.", it seems, somehow the branch of my PR is moved to a private area??? I don't know.. Let me know if you want me to make a new PR.

titusfortner:
Oh, weird, yeah, whatever the branch this was getting merged from must have been deleted. Go ahead and create a new PR.

sjieg:
Created new PR here https://github.com/watir/watir/pull/747
Also added some specs for the bg-color changing after flashing.

PR #704, opened 2018-01-21
implement date_field and date_time_field

titusfortner:
These are currently implemented as text_field, but `#set` does not work, since these display as read_only, which means it will error out on `#clear` call.

Per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) on `type="date"`

> One thing to note is that the displayed date format differs from the actual value — the displayed date format will be chosen based on the set locale of the user's browser, whereas the date value is always formatted yyyy-mm-dd.

(The same is true for `type="datetime-local"`)

Since there is no way to figure out the displayed date format, there is no way to ensure that typing text will work for any locale. We can ensure that this can be set by using the javascript value. 

I think this should apply to `#set` & `#value=` in addition to `#set!`, but I can see a case being made for only supporting the bang method since it is a javascript only solution. I just don't see an alternative for `#set` unless we override the read-only check.

Did I miss anything in this implementation?

Also, I just realized I didn't make the tests for collections, I'll make that happen before it gets merged.


titusfortner:
Oh, forgot to mention:

`DateField` only supports `Date`
and
`DateTimeField` supports `DateTime` and `Time`

lib/watir/elements/date_field.rb
```diff
@@ -0,0 +1,39 @@
+module Watir
+  class DateField < Input
+
+    #
+    # Selects this radio button.
```
p0deje: Copy paste?

p0deje (approved):

lib/watir/elements/date_field.rb
```diff
@@ -0,0 +1,39 @@
...
+    #
+
+    def set!(date)
+      message = "DateField##{__method__} only accepts instances of Date"
```
jkotests: Could we also support:
* Strings in the format "%Y-%m-%d" and
* Time objects?

In my application's date control, depending on what I am doing, I will use either the String format or a Date object - ie sometimes it's more convenient to directly specify the `String`.  I also use `active_support/time` to do things like `10.days.ago`. As this creates a `Time` object, it will be rejected by this validation.

lib/watir/elements/date_time_field.rb
```diff
@@ -0,0 +1,39 @@
+module Watir
+  class DateTimeField < Input
+
+    #
+    # Selects this radio button.
```
jkotests: Same copy/paste error as mentioned for DateField.

lib/watir/elements/date_time_field.rb
```diff
@@ -0,0 +1,39 @@
...
+    #
+
+    def set!(date)
+      message = "DateTimeField##{__method__} only accepts instances of DateTime or Time"
```
jkotests: Same as DateField, it would be nice to be able to support `String`.

spec/watirspec/elements/date_time_field_spec.rb
```diff
@@ -0,0 +1,194 @@
+require "watirspec_helper"
+
+describe "DateField" do
```
jkotests: Should be describing "DateTimeField".

spec/watirspec/elements/date_time_field_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    it "raises UnknownObjectException if the date-time field doesn't exist" do
+      expect { browser.browser.date_time_field(id: 'no_such_id').enabled? }.to raise_unknown_object_exception
+    end
+  end
+
+  # Manipulation methods
+  describe "#value= " do
+    it "sets the value of the element" do
+      browser.time
```
jkotests: `browser.time` isn't needed.

jkotests (changes requested):

lib/watir/elements/date_field.rb
titusfortner: fixed

lib/watir/elements/date_field.rb
titusfortner: I just added code to parse Date or DateTime if it is a String. Huh, I didn't realize Rails did their days using Time class... We can support it the way that DateTmeField is supported, or the user can `#strftime` or `#to_date`

titusfortner: Yeah, ok, I added support for Time class to both then.

titusfortner:
@jkotests I think that addresses everything. Let me know if there is anything else

lib/watir/elements/date_field.rb
```diff
@@ -0,0 +1,41 @@
...
+  class DateField < Input
+
+    #
+    # Enter the provided value.
+    #
+
+    def set!(date)
+      date = Date.parse date if date.is_a?(String)
+
+      message = "DateTimeField##{__callee__} only accepts instances of Date or Time"
```
jkotests: `DateTimeField` should just be `DateField`.

jkotests (approved):
Looks good to me except for one typo.

lib/watir/elements/date_field.rb
titusfortner: I'm winning with the copy/paste today

PR #708, opened 2018-03-07
Reference constant correctly

kevinjcoleman:
@titusfortner @jarib we were encountering a `NoMethodError: undefined method `KEY' for Selenium::WebDriver::Firefox::Options:Class` when using remote firefox like so. This fixes that error.
```
Watir::Browser.new :firefox, {headless: true, timeout: 120, url: Rails.application.secrets[:chargeback][:url]}
```


p0deje:
Thank you!

kevinjcoleman:
No problem!

PR #717, opened 2018-04-18
changes label_from_text to use :visible_text instead of :text

sjieg:
Using `b.text_field(label: /some text/)` threw a deprecation warning about using `:text` instead of `:visible_text`.  Updating the `label_from_text` function to use `:visible_text` fixes this issue.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -186,7 +186,7 @@ module Watir
         def label_from_text(label_exp)
           # TODO: this won't work correctly if @wd is a sub-element
           locate_elements(:tag_name, 'label').find do |el|
-            matches_selector?(el, text: label_exp)
+            matches_selector?(el, visible_text: label_exp)
           end
         end
 
--- a/spec/watirspec/elements/text_field_spec.rb
+++ b/spec/watirspec/elements/text_field_spec.rb
@@ -25,6 +25,8 @@ describe "TextField" do
       expect(browser.text_field(label: /(Last|First) name/)).to exist
       expect(browser.text_field(label: 'Without for')).to exist
       expect(browser.text_field(label: /Without for/)).to exist
+      expect(browser.text_field(label: 'With hidden text')).to exist
+      expect(browser.text_field(label: /With hidden text/)).not_to exist
     end
 
     it "returns the first text field if given no args" do
--- a/spec/watirspec/html/forms_with_input_elements.html
+++ b/spec/watirspec/html/forms_with_input_elements.html
@@ -39,6 +39,7 @@
             <label for="new_user_occupation">Occupation</label>
             <input type="text" name="new_user_occupation" id="new_user_occupation" value="Developer" onfocus="document.getElementById('onfocus_test').innerHTML = 'changed by onfocus event'"/> <br />
             <label>Without for <input /></label>
+            <label>With<span style="display:none;"> hidden</span> text<input /></label>
             <label for="new_user_species">Species</label>
             <input type="text" name="new_user_species" id="new_user_species" value="Homo sapiens sapiens" disabled="disabled" /> <br />
             <label for="new_user_code">Personal code</label>


p0deje:
We should only do that in case of regular expression. Can you confirm that we still use `text:` selector when we pass string to `label:` selector?

jkotests:
@sjieg, what does the label element look like? Is there actually a mix of visible and non-visible text?

I think the deprecation warning is being raised in situations where it shouldn't be. Consider the element:

~~~~~~~~
<div id="same">visible<span style="display:none;">hidden</span>visible</div>
~~~~~~~~

If you use a regular expression that only matches the visible portion, the deprecation warning still occurs. However, as the behaviour won't change when we switch to matching the text content, I wouldn't expect the warning.

~~~~~~~~
browser.div(text: /visible/).exists?
#=> 2018-04-18 17:14:03 WARN Watir [DEPRECATION] :text locator with RegExp values to find elements based on only visible text is deprecated. Use :visible_text instead.
~~~~~~~~

If this is the same scenario that you have (or where all of the element text is visible), I think the fix should be where we decide whether or not to output the deprecation warning.

sjieg:
@p0deje the `label_from_text` function is not being called when using `label: ''something'`. So, `:visible_text` is _only_ used the `:label` is called with a regular expression.

@jkotests I would agree that the deprecation warning itself might not be as expected. The difficulty being that the deprecated function received new, non-deprecation, functionality. However, I don't think that's relevant for this case, where you want to select an input element by the text of a label.

Then again, if you want to drag this distinction over to the `:label` selector, there should be different selectors, like `:label` and `:visible_label`. But once you start doing that, you can continue with `:label_title` and `:label_custom_attribute`. I don't think that's the goal of this selector. All in all, I think it's okay to say, "When using a regular expression on the `:label` selector, you look for :visible_text on that label". But it's your call, I just wanted to get rid of the deprecation spam :-)

p0deje:
The changes seem reasonable to me, but since we're slightly changing the behavior, do you mind adding a test for it? Something around [these expectations](https://github.com/watir/watir/blob/d99e5c83d600431e6cef6b2f9ee96f50f96a681d/spec/watirspec/elements/text_field_spec.rb#L24-L27) and [this HTML file](https://github.com/watir/watir/blob/master/spec/watirspec/html/forms_with_input_elements.html).

sjieg:
Added specs for the changed behaviour.

titusfortner:
So, I agree that it is the wrong error message, but the reason there is a deprecation notice there is because we're looking to change the behavior. Presumably `browser.text_field(label: /With text/)` passes now, and will fail when we update the code.

So, either it needs a different deprecation warning, or we need to make a change to how we are locating elements with labels.

titusfortner:
I think the right answer is to create a `:visible_label` key, even though that might be a pain.

I so want to spend some time simplifying this location code in general, this is all kind of... inelegant. :-/
@jkotests how close did you get with what you were working on earlier this year?

jkotests:
@titusfortner, I got pretty close. Still a bit inelegant with lots of areas for improvement, but there are less branches in the code.

What blocked me from pushing was a failing spec for an element going stale during lookup ([element_spec.rb](https://github.com/watir/watir/blob/17d8cceed21308e038794e64b1e327b7d3c0804c/spec/element_spec.rb#L22-L30)). Looking at the failing test again, it looks like the existing code doesn't completely handle the scenario either - see Issue #718.

I can revert my changes that minimized the usages of the Selenium locators, which should fix this failing spec. We won't get as much performance gains, but at least the simplified code could be made available.

titusfortner:
I have some other ideas to improve performance, we just really need to reduce the conditionals and branching. :)

titusfortner:
Ok, so here's the full scoop. With the html code provided in this PR, the following specs pass:
```
      expect(browser.text_field(label: 'With hidden text')).to exist
      expect(browser.text_field(label: 'With text')).not_to exist

      expect(browser.text_field(label: /With text/)).to exist
      expect(browser.text_field(label: /With hidden text/)).not_to exist
```

The problem is that when we update the deprecated code for Watir 7, the  second two will fail.

@sjieg would you need a `:visible_label` locator to make your tests work here? I'd prefer not to add an additional, obscure, difficult to understand locator, but that's the only way I'd know how to protect against non-backward-compatible behavior. (assuming that the deprecation updates that @jkotests made with d6e789bb8c98b2ab4d3cb535f81cb5d53efd6049 don't silence your issue)

p0deje (commented):
I'm not quite sure what to do about this, so I'll leave that to @titusfortner to decide 😛

jkotests:
In general, I expect people will be using `:text` over `:visible_text` for locating any element. `:text` can be more performant and it is easier to write. Therefore, I think `:label` should also default to `:text` (ie non-visible text).

To allow for backwards compatibility, what if we allow `:label` to accept a Hash argument (just like any other element locating)?

~~~~~~~~
browser.text_field(label: /some text/)                  #=> matches against label's non-visible text
browser.text_field(label: {visible_text: /some_text})   #=> matches against label's visible text
~~~~~~~~

This would avoid needing to add extra locators. It would also allow locating the label by any supported locator.

titusfortner:
Hmm, I'm not sure how I feel about nested hashes here. Let me post an alternative PR and we can see if it makes sense to maintain.

sjieg:
Closing because PR https://github.com/watir/watir/pull/746 fixes the issue in a different (more streamlined manner)

Issue #718, opened 2018-04-28
Infinite loop occurs when element goes stale during lookup and using_watir locator branch


jkotests:
## Meta -
Watir Version: 6.10.3
Selenium Version: 3.11.0
Browser Version: Chrome 66.0.3359.139
Browser Driver Version: ChromeDriver 2.38
OS Version: Windows 10

## Expected Behavior -

`Element#present?` should return false when an element goes stale during lookup - see the [element_spec.rb](https://github.com/watir/watir/blob/17d8cceed21308e038794e64b1e327b7d3c0804c/spec/element_spec.rb#L22-L30)

## Actual Behavior -

Watir returns the expected result when located using `Element::Locator#using_selenium` branch. However, Watir will get stuck in an infinite loop if the `Element::Locator#using_watir` branch is used.

## Steps to reproduce -

Given the HTML file:

~~~~~~~~html
<html>
  <body>
    <div id="foo" data-id="foo">test</div>
  </body>
</html>
~~~~~~~~

Run the following spec (which is a modification of the existing spec - [element_spec.rb](https://github.com/watir/watir/blob/17d8cceed21308e038794e64b1e327b7d3c0804c/spec/element_spec.rb#L22-L30)):

~~~~~~~~ruby
require 'watir'
require 'rspec/autorun'

describe Watir::Element do
  describe '#present?' do
    it "returns false if the element is stale and located by Watir" do
      browser = Watir::Browser.new
      browser.goto File.join(__dir__, 'test.htm')
      
      wd_element = browser.div(data_id: "foo").wd

      # simulate element going stale during lookup
      allow(browser.driver).to receive(:find_element).with(:xpath, ".//div[@data-id='foo']") { wd_element }
      browser.refresh

      expect(browser.div(:data_id, 'foo')).to_not be_present
      browser.close
    end
  end
end
~~~~~~~~

Notice that the spec runs forever.

titusfortner:
@jkotests let me know if this one isn't resolved

jkotests:
@titusfortner , this is resolved now. Thanks.

PR #719, opened 2018-04-29
Reduce `enabled` Wire Calls

titusfortner:
I believe this addresses #718.

There is a fundamental difference between locating elements from the context of `#element_call`, i.e. to *do something with the element, and locating from the context of a predicate call (e.g. `#exists?` or `#enabled?`, etc). Behaviors applicable for one do not always apply to the other. This PR attempts to do a better job distinguishing them and using the correct logic in the correct locations.

We have been using `#enabled?` to represent a generic wire call to ensure that an element is not stale before using it. The obvious extension of what I did in Watir 6.4 (the big performance release last summer) is to not make this check every time we go to use an element, but to follow the same pattern of trying to use it and dealing with the unknown object error.

Our Element related watir specs see about a 60% drop in usage of this wire call (approx 3700 to 1350) with this PR code. This could be pretty significant when running remotely.

This code will also return `false` for `#stale?` when the element is still in the DOM but the driver is pointing to a different browsing context (nested iframe, etc). This isn't a backward compatible change, but I think it has been the intended behavior and just overlooked. I don't expect people are using a staleness check to ensure they've switched to a different frame.

Also, we don't have to contort ourselves to emulate staleness in the specs any more. We have a predicate `#stale?` that does not change regardless of how often it is called, and every place that can result in a staleness exception is now being rescued and handled, so there is no longer any need  need to fake it out with mocked Selenium calls.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -47,6 +47,7 @@ module Watir
     #
 
     def exists?
+      return false if @element && stale?
       assert_exists
       true
     rescue UnknownObjectException, UnknownFrameException
@@ -401,7 +402,10 @@ module Watir
     #
 
     def visible?
-      element_call(:assert_exists) { @element.displayed? }
+      assert_exists
+      @element.displayed?
+    rescue Selenium::WebDriver::Error::StaleElementReferenceError
+      raise unknown_exception
     end
 
     #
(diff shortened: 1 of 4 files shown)


titusfortner:
Performance wise, by far the most significant wire call that Watir makes at this point is `#tag_name` (like almost 50% of the wire calls in my testbed). I think we are using it for validation much more often than we actually need to, but I don't want to wade into this part of the code until we can get it more simplified. Hopefully this PR doesn't significantly impact the work @jkotests did, and it'll be easy enough to move to a more straightforward approach.

jkotests (approved):

p0deje (approved):
Can't wait to try this!

titusfortner:
I'm getting a weird intermittent error when i add this to @jkotest commit. His looked fine by itself, so I merged that one and will keep investigating for this one. It is the submit method, Firefox only, Travis only, and intermittent. :)

PR #720, opened 2018-04-29
Simplify element locator

jkotests:
@titusfortner, these are the changes I have for simplifying the locator code.

The idea is:

1. Separate locators that can be included in the XPath from those that require filtering (inspecting the element).
2. Build the XPath
3. Apply filters

To avoid changing too much, the changes are isolated to the Element::Locator class. At some point, some of the logic could be moved to the selector builder (eg adding of Regexps predicates). 

Some random notes:
* For the specs, the get_element_tag_name wire calls dropped by 76% (from 11,656 to 2,772). 
* When locating an element by index and filtering is required, lazy evaluation is used to avoid inspecting elements that are not needed. This should help address the [performance issue raised on Watir General](https://groups.google.com/d/msg/watir-general/vKqwbPyxgTc/JjOs4MuXCAAJ).

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -59,7 +59,8 @@ module Watir
             if filter == :all
               found = locate_elements(sel, value)
               return found if sel == :tag_name
-              return filter_elements_by_locator(found, tag_name: tag_name, filter: filter).compact
+              filter_selector = tag_name ? {tag_name: tag_name} : {}
+              return filter_elements(found, filter_selector, filter: filter).compact
             else
               found = locate_element(sel, value)
               return sel != :tag_name && tag_name && !validate([found], tag_name) ? nil : found
@@ -69,38 +70,37 @@ module Watir
         end
 
         def using_watir(filter = :first)
+          query_scope = ensure_scope_context
           selector = selector_builder.normalized_selector
-          visible = selector.delete(:visible)
-          visible_text = selector.delete(:visible_text)
-          tag_name = selector[:tag_name].is_a?(::Symbol) ? selector[:tag_name].to_s : selector[:tag_name]
-          validation_required = (selector.key?(:css) || selector.key?(:xpath)) && tag_name
+
+          if selector[:label]
+            query_scope = convert_label_to_scope_or_selector(query_scope, selector)
+            return unless query_scope # stop, label not found
+          end
 
           if selector.key?(:index) && filter == :all
             raise ArgumentError, "can't locate all elements by :index"
           end
-          idx = selector.delete(:index) unless selector[:adjacent]
+
+          filter_selector = delete_filters_from(selector)
 
           how, what = selector_builder.build(selector)
+          unless how
+            raise Error, "internal error: unable to build Selenium selector from #{selector.inspect}"
+          end
+          what = add_regexp_predicates(what, filter_selector) if how == :xpath
 
-          needs_filtering = idx && idx != 0 || !visible.nil? || !visible_text.nil? || validation_required || filter == :all
-
+          needs_filtering = filter == :all || !filter_selector.empty?
           if needs_filtering
-            matching = matching_elements(how, what, selector)
-            return filter_elements_by_locator(matching, visible, visible_text, idx, tag_name: tag_name, filter: filter)
-          elsif how
-            locate_element(how, what)
+            elements = locate_elements(how, what, query_scope) || []
+            filter_elements(elements, filter_selector, filter: filter)
           else
-            wd_find_by_regexp_selector(selector, :first)
+            locate_element(how, what, query_scope)
           end
         end
 
         def validate(elements, tag_name)
           elements.compact.all? { |element| element_validator.validate(element, {tag_name: tag_name}) }
-        end
-
-        def matching_elements(how, what, selector = nil)
-          found = how ? locate_elements(how, what) : wd_find_by_regexp_selector(selector, :all)
-          found || []
         end
 
         def fetch_value(element, how)
@@ -112,6 +112,8 @@ module Watir
               Watir.logger.deprecate(':text locator with RegExp values to find elements based on only visible text', ":visible_text")
             end
             vis
+          when :visible
+            element.displayed?
           when :visible_text
             element.text
           when :tag_name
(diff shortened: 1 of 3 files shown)


titusfortner:
Just looking from my phone this looks great. Fewer changes than I was expecting, plus I'm guessing this should be a pretty straightforward addition to #719.

Doesn't look like Fixnum#negative? supports 2.2.x.
@p0deje are you ok deprecating < 2.3 since previous versions aren't maintained any longer?
We should update Travis to be running 2.5.1 as well.

Looking forward to running this with selenium_statistics.gem when I get home

titusfortner:
Yeah, so, this code decreases *total wire calls in the suite by 34%, and when combined with the code I wrote, it decreases its up to 41%!!!

I still need to step through the code, but this could be a bigger performance improvement than we saw last summer!

lib/watir/locators/element/locator.rb
```diff
@@ -70,39 +71,58 @@ def using_selenium(filter = :first)
...
           end
-          idx = selector.delete(:index) unless selector[:adjacent]
+
+          filter_selector = delete_filters_from(selector)
+          filter_selector[:tag_name] = tag_name if validation_required
+
+          query_scope = ensure_scope_context
+
+          if filter_selector.key?(:label) && selector_builder.should_use_label_element?
+            label = label_from_text(filter_selector.delete(:label)) || return
```
titusfortner: Does `|| return` still belong here now that it isn't in `#wd_find_by_regexp_selector`?

Also, what do we think about breaking up this method more?
Can we pull these lines out into something like:
```
query_scope = set_query_scope(selector, filter_selector)
```

lib/watir/locators/element/locator.rb
```diff
@@ -70,39 +71,58 @@ def using_selenium(filter = :first)
...
+            end
+          end
 
           how, what = selector_builder.build(selector)
+          unless how
+            raise Error, "internal error: unable to build Selenium selector from #{selector.inspect}"
+          end
 
-          needs_filtering = idx && idx != 0 || !visible.nil? || !visible_text.nil? || validation_required || filter == :all
+          if how == :xpath && can_convert_regexp_to_contains?
```
titusfortner: Can we pull out this conditional into its own method like:
```
what = #{what}#{add_predicates(filter_selector)} if how ==:xpath
```

lib/watir/locators/element/locator.rb
```diff
@@ -123,64 +145,44 @@ def fetch_value(element, how)
...
+            matches.take(idx + 1).to_a[idx]
+          else
+            elements.select { |el| matches_selector?(el, selector) }
           end
-
-          elements = locate_elements(how, what, query_scope)
-          filter_elements_by_regex(elements, rx_selector, filter)
-        end
-
-        def filter_elements_by_locator(elements, visible = nil, visible_text = nil, idx = nil, tag_name: nil, filter: :first)
```
titusfortner: Getting rid of these methods means I need to update `watigiri` at the very least. We'll need to ensure `watizzle` & `watir_css` also work, even though I don't think anyone uses those?

lib/watir/locators/element/locator.rb
```diff
@@ -123,64 +145,44 @@ def fetch_value(element, how)
...
-          if how == :xpath && can_convert_regexp_to_contains?
-            rx_selector.each do |key, value|
-              next if key == :tag_name || key == :text
 
-              predicates = regexp_selector_to_predicates(key, value)
-              what = "(#{what})[#{predicates.join(' and ')}]" unless predicates.empty?
-            end
+            # Lazy evaluation to avoid fetching values for elements that will be discarded
+            matches = elements.lazy.select { |el| matches_selector?(el, selector) }
+            matches.take(idx + 1).to_a[idx]
```
titusfortner: TIL `#lazy` and `#take` and will attempt to find ways to use them more.

lib/watir/locators/element/locator.rb
```diff
@@ -70,39 +71,58 @@ def using_selenium(filter = :first)
 
         def using_watir(filter = :first)
           selector = selector_builder.normalized_selector
-          visible = selector.delete(:visible)
-          visible_text = selector.delete(:visible_text)
           tag_name = selector[:tag_name].is_a?(::Symbol) ? selector[:tag_name].to_s : selector[:tag_name]
           validation_required = (selector.key?(:css) || selector.key?(:xpath)) && tag_name
 
           if selector.key?(:index) && filter == :all
             raise ArgumentError, "can't locate all elements by :index"
```
titusfortner: What do you think about moving this to line 160 to avoid the extra conditionals here?

titusfortner (commented):
This looks great. There's just a lot of code in `#using_watir` now.

I'm thinking we release v6.11.beta1 and encourage a few Watir using companies to check it out to make sure it works.
Also @lmtierney did you want to gander at this since you're presumably up on our location code with `nerodia`?

lib/watir/locators/element/locator.rb
jkotests: `|| return` is still required. If the label cannot be found, there is no point in continuing - ie return element not found.

I had debated breaking up the code. I still think a lot of this should be the selector builders responsibility. However, given I don't know when I'll get around to tackling that, I'll at least look at tidying up the current code.

titusfortner:
Also, can you squash these commits when you rebase it next?

lib/watir/locators/element/locator.rb
titusfortner: Ah, I see it now. Thanks

titusfortner: How do you see the responsibilities needing to be split? The only thing that seems obvious to me to figure out how to push it into the SelectorBuilder class is the predicate generation. Everything else is some kind of filtering or validation.

jkotests: I feel like deciding what should be in the XPath vs a filter would be part of the selector builder. The selector builder would:
* Be given all of the properties (selectors plus filters)
* Take which ever properties it can handle and build the XPath
* Return the XPath along with a list of properties that it didn't know what to do with (ie would eventually be the filters)

I guess I see the Locator class as managing the workflow. It shouldn't be deciding what the selector builder can or cannot do.

lib/watir/locators/element/locator.rb
jkotests: I assume you are okay with the performance hit - ie we'd go grab elements from the page, only to later say we cannot do anything?

Given this should only happen in very rare occurrences, it seems like a reasonable change. Though sadly it doesn't seem to just be a matter of moving the line of code. I got a failed spec when trying, so I'll have to take another stab at it.

titusfortner: It'll only be a performance hit if the test is going to fail, which I don't care about optimizing. If it isn't trivial though don't worry about it

p0deje (approved):
I don't quite follow all the changes because I need to deep dive into the code to do that, but as long as specs pass, I approve the changes!

jkotests:
@titusfortner, I was able to split out most of the code you suggested, though not exactly as suggested. I changed the #negative? check so that this will hopefully work on Ruby 2.2. Commits were also squashed.

Releasing a beta version would be great. I am definitely concerned that the specs don't cover everything.

PR #724, opened 2018-05-10
Locate more elements using Watir's Selector Builder 

titusfortner:
This is my stab at optimizing and reorganizing. I haven't run my stats package on it, yet.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -5,14 +5,10 @@ module Watir
         attr_reader :selector_builder
         attr_reader :element_validator
 
-        WD_FINDERS = [
-          :class,
-          :class_name,
+        W3C_FINDERS = [
           :css,
-          :id,
           :link,
           :link_text,
-          :name,
           :partial_link_text,
           :tag_name,
           :xpath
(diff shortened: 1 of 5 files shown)


titusfortner:
Well, pretty much this decreases tag_name calls in our suites by about 25%. So overall numbers it isn't a lot, but it also likely represents improvements for more typical usage.

p0deje (approved):
As long as specs pass, 👍

lib/watir/locators/element/locator.rb
```diff
@@ -49,53 +45,31 @@ def locate_all
...
-              return found if sel == :tag_name
-              filter_selector = tag_name ? {tag_name: tag_name} : {}
-              return filter_elements(found, filter_selector, filter: filter).compact
-            else
-              found = locate_element(sel, value)
-              return sel != :tag_name && tag_name && !validate([found], tag_name) ? nil : found
-            end
-          end
-          nil
+          return unless @selector.size == 1
```
jkotests: This doesn't work for `:link`, `:link_text` and `:partial_link_text - eg:

~~~~~~~~ruby
browser.link(link_text: 'valid_text_value').text
~~~~~~~~

The `@selector` will be `{:link_text=>'valid_text_value', :tag_name=>'a'}`. It'll get pushed to `using_watir`, which treats this as an attribute instead of Seleniums special link logic.

lib/watir/locators/element/locator.rb
```diff
@@ -141,57 +116,77 @@ def filter_elements(elements, selector, filter: :first)
...
-        def delete_filters_from(selector)
-          filter_selector = {}
+        def create_normalized_selector(filter)
+          return @normalized_selector if @normalized_selector
+          @driver_scope = ensure_scope_context
+
+          @normalized_selector = selector_builder.normalized_selector
+
+          process_label if @normalized_selector[:label]
+          return if @normalized_selector.nil?
```
jkotests: Would this only apply if `#process_label` did not find a label? To make it clear it's special logic for handling the labels, maybe we could do:

~~~~~~~~ruby
if @normalized_selector[:label]
  process_label 
  return if @normalized_selector.nil?
end
~~~~~~~~

jkotests (changes requested):

lib/watir/locators/element/locator.rb
titusfortner: Ah, right, well, this will be more simple when we just stop supporting them. :-D

jkotests (approved):

titusfortner:
And I forgot a question mark. I'll update when I get to a laptop. Sounds like there is also some legacy page object issue that needs to be verified before we do another beta release

PR #728, opened 2018-05-10
separate out unit tests that don't require a browser

titusfortner:
This puts all the browser specs inside `watirspecs` and moves all the unit tests into its own folder.

The need for this is more obvious in my next PR, but wanted feedback on this as I don't understand how Watirspecs work as well as I could. :)


p0deje (approved):

titusfortner:
merged 8e95051cc7cbb8c6e388ae3dd155dd3122b98018

Issue #730, opened 2018-05-16
No longer able to wait for re-created element to become present


jkotests:
## Meta -
Watir Version: 6.11.0.beta1, 6.11.0.beta2 (problem does not occur in 6.10.3)
Selenium Version: 3.12.0
Browser Version: 66.0.3359.139
Browser Driver Version: 2.38
OS Version: Windows 10

## Expected Behavior -

I have an element that:
* Initially exists in the DOM, but is hidden
* Is re-created before being displayed

I would expect to be able to wait for the element to become present by using `#wait_until_present`.

## Actual Behavior -

Watir behaved as expected in 6.10.3. However, as of 6.11.0.beta, this no longer works. A timeout will now occur.

## Steps to reproduce -

Given the HTML file (modified from wait.html):

~~~~~~~~html
<!DOCTYPE HTML>
<html>
  <head>
    <title>wait test</title>
    <script type="text/javascript" charset="utf-8">

      function setTimeoutAddDisplay(id, current_id, timeout) {
          setTimeoutAdd(id, "none", id, current_id, timeout);
          setTimeoutDisplay(id, "block", timeout*2)
      }

      function setTimeoutDisplay(id, display, timeout) {
          setTimeout(function() {
              document.getElementById(id).style.display = display;
          }, timeout);
      }

      function setTimeoutAdd(id, display, text, existing_id, timeout) {
          var newDiv = document.createElement("div");
          newDiv.setAttribute("id", id);
          newDiv.style.display = display;
          newDiv.appendChild(document.createTextNode(text));

          var currentDiv = document.getElementById(existing_id);

          setTimeout(function() {
              document.body.insertBefore(newDiv, currentDiv);
          }, timeout);
      }

      function setTimeoutRemove(id, timeout) {
          setTimeout(function() {
              var e = document.getElementById(id);
              e.parentNode.removeChild(e);
          }, timeout);
      }
    </script>
  </head>

  <body>
    <div id="foo" style="display:block;">foo</div>
    <div id="bar" style="display:none;" onclick='this.innerHTML = "changed"'>bar</div>
    <a id="readd_bar" href="#" onclick="setTimeoutRemove('bar', 1000); setTimeoutAddDisplay('bar', 'foo', 2000);">re-add bar</a>
  </body>
</html>
~~~~~~~~

The following code will now timeout:

~~~~~~~~ruby
browser = Watir::Browser.new :chrome
browser.goto File.join(__dir__, 'test.htm')
[...]

PR #732, opened 2018-05-17
Reset for stale elements

titusfortner:
I figured out #730 came from not properly replicating this line in the new code:
https://github.com/watir/watir/commit/ca377c3d0041edae59bd8ce7bb7390571ac4a471#diff-a0982f69261cbd95246962ebecdabf3cL571

What I don't like about this code is that it means that calling `#present?` twice returns different results. So, I'd like to deprecate that behavior. `#stale?` will always return result of Staleness no matter how many times it is called. But `#present?` will treat an element like a Watir element, not a Selenium element and say that if an element is at that address it is present, even if it isn't the same Selenium element, and will return true or false based on current status.

This will also deprecate `#visible?` which we've discussed previously.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -405,6 +405,7 @@ module Watir
       assert_exists
       @element.displayed?
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
+      reset!
       raise unknown_exception
     end
 
--- a/spec/watirspec/html/wait.html
+++ b/spec/watirspec/html/wait.html
@@ -54,6 +54,7 @@
     <a id="hide_foo" href="#" onclick="setTimeoutDisplay('foo', 'none', 500);">hide foo</a>
     <a id="remove_foo" href="#" onclick="setTimeoutRemove('foo', 1000);">remove foo</a>
     <a id="add_foobar" href="#" onclick="setTimeoutAddDisplay('foobar', 'bar', 1000);">add foobar</a>
+    <a id="readd_bar" href="#" onclick="setTimeoutRemove('bar', 500); setTimeoutAddDisplay('bar', 'foo', 1000);">re-add bar</a>
     <div id="buttons">
       <button id="btn" type="button" onclick="setDisabled('btn', true, 0)" disabled>Click To Disable!</button>
       <a id="enable_btn" href="#" onclick="setDisabled('btn', false, 500);">enable btn</a>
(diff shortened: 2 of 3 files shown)


p0deje:
What about `visible: true` selector? How will it work if visible is deprecated? Should we maybe just alias `visible?` to `present?`?

titusfortner:
I can see two ways to handle `#visible?` deprecation, let's save that conversation for next release. I've removed it from this PR.

Are we ok with changing the result of `#present?` for a stale element?

p0deje:
Sounds reasonable, I don't have anything against.

jkotests:
#730 seems like a rare scenario, so I don't mind the deprecation. However, once the behaviour in `#present?` is deprecated, what would I write for the example in #730?

Would I have to declare my link twice or is there a better way to force the re-locating:

~~~~~~~~
browser.link(id: 'readd_bar').click.wait_until(&:stale?)
browser.link(id: 'readd_bar').wait_until(&:present?)
~~~~~~~~

titusfortner:
#730 will continue to work after deprecation.
In fixing #730 I realized that I made assumptions that aren't true, but that I think should be true.

1. We shouldn't use `#present?` to check for staleness.
2. When the DOM doesn't change, making the same method call shouldn't return different results on subsequent executions.

jkotests:
I'm just a bit confused. If it's going to continue to work after deprecation, how come I get a deprecation warning?

titusfortner:
You get the deprecation warning for it only because it is a weird use case.
This might be another instance of being overly aggressive at throwing deprecation warnings, but I'm not sure a better way to do it.

I'm concerned about the following scenario. Right now it works, but I don't think it should:
```
      element = browser.div(id: 'foo').tap(&:exists?)
      browser.refresh
      expect { element.wait_while(&:present?) }.to_not raise_exception
```

I want to change `#present?` to retry instead of returning false when it is Stale.
```
    def present?
      assert_exists
      @element.displayed?
    rescue UnknownObjectException, UnknownFrameException
      false
      reset!
      retry
    end
```

This would break the above scenario, and would require people to change from: `element.wait_while(&:present?)` to `element.wait_until(&:stale?)`.

titusfortner:
Actually, I'm going to remove this deprecation message as well... I want to get this gem released with the relocations, so I'm going to make sure the specs are passing with the updated `#reset!` code, then next week we can do a point release with updated deprecations that we hash out. :)

PR #733, opened 2018-05-18
Only deprecate text and regexp when text content will change behaviour

jkotests:
The deprecation warning for :text with Regexp is too aggressive. A warning is *always* logged. We still want people to prefer :text over :visible_text since :text is faster (can be combined into XPath).

The deprecation warning should only be logged when switching to text content will change the :text with Regexp behaviour.

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -83,12 +83,7 @@ module Watir
         def fetch_value(element, how)
           case how
           when :text
-            vis = element.text
-            all = Watir::Element.new(@query_scope, element: element).send(:execute_js, :getTextContent, element).strip
-            unless all == vis.strip
-              Watir.logger.deprecate(':text locator with RegExp values to find elements based on only visible text', ":visible_text")
-            end
-            vis
+            element.text
           when :visible
             element.displayed?
           when :visible_text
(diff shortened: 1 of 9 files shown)


lib/watir/locators/element/locator.rb
```diff
@@ -196,13 +191,23 @@ def label_from_text
...
             else
               what === fetch_value(element, how)
             end
           end
+
+          if selector[:text]
+            text_content = Watir::Element.new(@query_scope, element: element).send(:execute_js, :getTextContent, element).strip
+            text_content_matches = selector[:text] === text_content
+            unless matches == text_content_matches
+              Watir.logger.deprecate(':text locator with RegExp values to find elements based on only visible text', ":visible_text")
```
titusfortner: What if we change this to:
```
Watir.logger.deprecate(":text locator with RegExp: #{@selector[:text].inspect} matched an element with hidden text", ":visible_text")
```

jkotests: That seems reasonable. I can update the pull request tonight if you haven't done so by then.

titusfortner (approved):
This is good to merge

lib/watir/locators/element/locator.rb
jkotests: @titusfortner , message has been updated.

titusfortner: You have merge powers, right? Shows you're an owner, just making sure.

PR #734, opened 2018-05-18
deprecate `#visible?`

titusfortner:
We have a few options with this.

1. This PR, which is straightforward deprecation so there is no confusion. No one uses visible as a method any longer. Normally I like aliasing as part of the "principle of least surprise," but when we are changing the behavior, we are changing what is most likely to be surprising about the method.

2. Alias to `#present?`. This has the problem of trying to figure out how to give a good deprecation notice that is sufficiently focused before we change behavior. I'm especially concerned with how Page Object gems which have their own logic for what "visible" means might be using this.

3. Leave as-is and just create a better guide to `#exists?` vs `#present?` vs `#displayed?` vs `#visible?`. 

I personally don't see a problem with deprecating `#visible?` while still having a `:visible` locator.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -402,6 +402,7 @@ module Watir
     #
 
     def visible?
+      Watir.logger.deprecate "#visible?", "#present?"
       assert_exists
       @element.displayed?
     rescue Selenium::WebDriver::Error::StaleElementReferenceError


p0deje:
What's the reason to deprecate? I think I still use `#visible?` 😄

titusfortner:
The difference between it and #present? is both small and confusing. It doesn't provide any additional information, and there's no use case for needing an exception over a false return

jkotests:
I would guess that there are more people using `#visible?` with elements that exist than those that have rescue blocks for elements that might not exist. I think it would be better to alias the two methods or do nothing.

p0deje:
Well, alias looks good to me. We might want to filter element collection by visibility:

```ruby
text_fields.select(&:visible?)
```

Visible makes sense here, while the code selecting for `&:present?` is not expressive - we don't need to check if elements exist in a collection.

titusfortner:
I'm not following this logic. You can add visible to the selector if that is what is desired. Especially now that we're lazy loading most collections now. Use the selector visible: true instead of filtering after.

p0deje:
You might or might not want to use a selector. Consider the following code:

```ruby
class Page
  divs :all_we_need, class: 'foo'

  def all
    {
       visible: all_we_need_divs.select(&:visible?)
       hidden: all_we_need_divs.reject(&:visible?)
    }
  end
end
```

Of course we can rework it into selectors, but why do we force that? Watir always allowed to use both selectors and custom methods interchangeably (e.g. `class` selector and `class_name` method, `data_foo` selector and same-named method, etc.). I see no reason to have `visible` selector but _not_ have `visible?` method.

titusfortner:
I'm still not following what this means for code. Sounds like we're not in favor of straight deprecating like I was hoping, so either filtering by present or visible would work.

titusfortner:
I'm not seeing how alias affects this

p0deje:
Ok, let me re-phrase it: if you can use `visible` as selector, why can't you use it as a method?

The other question is: if during debugging you want to inspect why element is not present, how would you make check that element exists but not visible?

titusfortner:
exists? && !visible? :)

titusfortner:
I don't want to encourage people to rescue exceptions in their code

p0deje:
> exists? && !visible? :)

But visible is deprecated.

titusfortner:
Heh. What I get for replying from my phone at an airport. `exists? && !present?` would give the desired information.

p0deje:
Yeah, but that's not self-explanatory code. It's a bit strange to see element that exists but not present? You need to know exactly how it's implemented to reason about that code.

Another question is if we don't want people to use `visible?` method, why do we want them to use `:visible` selector? Why not renaming it to `:present` too?

```ruby
browser.div(class: 'foo', present: true)
```

titusfortner:
That makes sense. I'll close this and re-open a new one with code for aliasing.

PR #735, opened 2018-05-18
deprecate using `#present?` to determine if an element is stale

titusfortner:
As discussed in #732:

I'm concerned about the following scenario. Right now it works, but I don't think it should:
```
element = browser.div(id: 'foo').tap(&:exists?)
browser.refresh
expect { element.wait_while(&:present?) }.to_not raise_exception
```
I want to change `#present?` to `retry` instead of returning `false` when the element is stale.
```
def present?
  assert_exists
  @element.displayed?
rescue UnknownObjectException, UnknownFrameException
  false
  reset!
  retry
end
```
This would require the scenario above to change to:
```
element = browser.div(id: 'foo').tap(&:exists?)
browser.refresh
expect { element.wait_until(&:stale?) }.to_not raise_exception
```

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -405,6 +405,7 @@ module Watir
       assert_exists
       @element.displayed?
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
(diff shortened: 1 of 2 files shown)


p0deje (approved):
Let's fix TODOs, but otherwise seems good.

jkotests (approved):

PR #736, opened 2018-05-18
Show example of how to fix specific code in deprecation notice

titusfortner:
Does this make sense as a more helpful deprecation notice?


lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -49,7 +49,8 @@ def attribute_expression(building, selectors)
           def equal_pair(building, key, value)
             if key == :class
               if value.strip.include?(' ')
-                Watir.logger.deprecate "Using the :class locator to locate multiple classes with a String value", "Array"
+                Watir.logger.deprecate "Using the :class locator to locate multiple classes with a String value (i.e. #{value})",
+                                       "Array (e.g. [#{value.gsub(' ', ', ')}])"
```
jkotests: Like the idea, but I don't think the Array comes out right. A value of "a b" will create the message "Array (e.g. [a, b])". We should include the quotes around each value. 

I think you want `"Array (e.g. #{value.split})"` to give "Array (e.g. ["a", "b"])".

jkotests (changes requested):

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: this was a dumb mistake, thanks

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -49,7 +49,8 @@ def attribute_expression(building, selectors)
           def equal_pair(building, key, value)
             if key == :class
               if value.strip.include?(' ')
-                Watir.logger.deprecate "Using the :class locator to locate multiple classes with a String value", "Array"
+                Watir.logger.deprecate "Using the :class locator to locate multiple classes with a String value (i.e. #{value})",
+                                       "Array (e.g. [#{value.split}])"
```
jkotests: This still doesn't quite format correctly. A value of "a b" will result in:
* "Using the :class locator to locate multiple classes with a String value (i.e. a b)" - should include quotes around value?
* "Array (e.g. [[\"a\", \"b\"]])" - duplicates "[" and "]"

jkotests (changes requested):

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: oops, yes, I left in the brackets, should just be:
```
"Array (e.g. #{value.split})"
```

jkotests: Did you want to quote the deprecated String value - ie `(i.e. "a b")` instead of `(i.e. a b)`?

titusfortner: yes, and I'll even fix the associated spec while I'm at it

jkotests (approved):

PR #737, opened 2018-05-21
Angular attribute support

jfitisoff:
HTML for Angular apps can have boolean attributes like "_ngcontent-c24" and "_nghost-pmm-5"

These currently can't be used as custom attributes because all underscores in the attribute keys get converted to dashes. This change will skip a leading underscore when converting the attribute key so that these attributes get processed successfully.

There's a performance hit but it's very tiny, about 0.1 seconds per 100k key conversions.

Adding support for attributes with leading underscores would allow you to do some useful things for angular apps. One example would be the angular components here:

https://material.angular.io/components/select/overview

When one of these select controls is clicked, the custom element that contains the select options shows up outside of the element for the select control. But the ngcontent attribute from the select control could be used to directly locate the container element for the select options as well the options themselves.

I spent some time looking at how custom attributes are getting implemented. There may be a few other cases where a leading underscore is used (other than angular.) I couldn't find any case where a leading dash was used (apparently that's _way_ too nutty for anyone out there.) But this change would support both of those cases. Because of how they'd have to specify these sorts of attributes it seems unlikely that anyone would do it by accident.

@titusfortner

### DIFF
--- a/lib/watir/locators/element/selector_builder/xpath.rb
+++ b/lib/watir/locators/element/selector_builder/xpath.rb
@@ -66,6 +66,8 @@ module Watir
             case key
             when :text, 'text'
               'normalize-space()'
+            when String
+              "@#{key}"
             when :href
               # TODO: change this behaviour?
               'normalize-space(@href)'
--- a/spec/element_locator_spec.rb
+++ b/spec/element_locator_spec.rb
@@ -123,6 +123,12 @@ describe Watir::Locators::Element::Locator do
                    text: "foo"
       end
 
+      it "handles 'text' key when it's a string" do
+        expect_one :xpath, ".//div[normalize-space()='foo']"
+        locate_one tag_name: "div",
+                   "text" => "foo"
+      end
+
       it "translates :caption to :text" do
         expect_one :xpath, ".//div[normalize-space()='foo']"
 
@@ -142,6 +148,13 @@ describe Watir::Locators::Element::Locator do
 
         locate_one tag_name: "div",
                    aria_label: "foo"
+      end
+
+      it "doesn't modify attribute name when the attribute key is a string" do
+        expect_one :xpath, ".//div[@_ngcontent-c24]"
+
+        locate_one tag_name: "div",
+                   "_ngcontent-c24" => true
       end
 
       it "normalizes space for the :href attribute" do


titusfortner:
My first thought was that we should add it to [WatirAngular](https://github.com/titusfortner/watir_angular/) instead of here, but this is a very minimal change, so that seems overkill. @jkotests / @p0deje any objections?

p0deje:
I don't see [anything in the spec](https://html.spec.whatwg.org/#syntax-attribute-name) that forbids to start attribute names from underscore, so let's merge.

jkotests:
I think that adding this makes sense. There have been cases where the conversion from underscores to dashes was problematic. For example, one person had a attribute of [`data-card_id`](https://stackoverflow.com/q/24778434/1200545). Adding this would be a step towards addressing the larger issue.

However, I'm not convinced that the regexp, in `gsub(/(?<=[a-z,0-9])_/, '-')`, is correct. A key of "data_tesT_a" will be converted to "data-tesT_a" when before it would be "data-tesT-a".

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -74,7 +74,7 @@ def lhs_for(_building, key)
               # https://github.com/watir/watir/issues/72
               XpathSupport.downcase('@type')
             else
```
titusfortner: What if we do:
```
when String
  "@#{key}"
```
`Symbol` always converts `_` to `-`, but `String` won't do any conversion

jfitisoff:
Yeah, I'd actually prefer that (just pass through without modifying if it's a string.) I can make that change if everyone's happy with that.

titusfortner:
Until 11/2017 we forced attributes to be valid, so this idea would just be a minor addition to it. 

@p0deje I can't load that page on the wifi I'm on (keeps crashing my browser), but I thought that custom attributes per spec needed to be prepended with `data-`, so all of these things would technically be outside the spec.

p0deje:
@titusfortner You are actually right and the only custom attributes allowed are `data-*`, though [there are discussions about changing that](https://github.com/whatwg/html/issues/2271). Keeping this in mind, I think we should just pass strings unmodified so that anyone can use any kind of custom attributes without changing Watir code.

@jfitisoff Can you fix indentation (it should be 2 spaces)? Once it's done, you can either squash commits or we'll do that when merging PR.

jfitisoff:
@p0deje yeah, sorry about that, was cutting and pasting between two editors and missed that part. indentation is hard. :-)

titusfortner:
@p0deje that conversation was way too long to follow. If I get to go to TPAC in October maybe I'll check in on it. :)

Also, I'm working on a fix for the specs so we don't have everything showing up as failing. I think I have the right fix, expect it soon.

Issue #738, opened 2018-05-29
Watir 6.11 collections no longer compatible with enumerable.


Donavan:
In commit 7c4e28861b1c12a69b2527f8a567046b5fcdf56e, the implementation of [] for element collection was changed in a way that breaks compatibility with enumerable.

[] has to accept more than just a single index.  By removing the to_a[idx] you've removed the ability to use ranges and force us to take one at a time.  For example: divs[1..4] should work and did work in 6.10

titusfortner:
Technically, this has nothing to do with Enumerable, since that module is intended to provide a means for iterating, and only requires the implementation of an `#each` method. Additionally, the current documentation specifies that the `#[]` method only supports an `Integer`
>     # @param [Integer] idx Index of wanted element, 0-indexed

With the current code, you can use: `divs.to_a[1..4]`

That being said, adding explicit support for a Range input makes sense to me, so I made the PR referenced above.

PR #740, opened 2018-05-29
add explicit support for Range inputs to `ElementCollection#[]`

titusfortner:
Addresses #738

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -35,20 +35,24 @@ module Watir
     alias_method :empty?, :none?
 
     #
-    # Get the element at the given index.
+    # Get the element at the given index or range.
     #
     # Any call to an ElementCollection including an adjacent selector
     # can not be lazy loaded because it must store correct type
     #
-    # @param [Integer] idx Index of wanted element, 0-indexed
-    # @return [Watir::Element] Returns an instance of a Watir::Element subclass
+    # Ranges can not be lazy loaded
+    #
+    # @param [Integer, Range] value Index (0-based) or Range of desired element(s)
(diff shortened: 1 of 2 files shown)


lib/watir/element_collection.rb
```diff
@@ -35,20 +35,22 @@ def each(&blk)
...
+    # @param [Integer, Range] value Index (0-based) or Range of desired element(s)
     # @return [Watir::Element] Returns an instance of a Watir::Element subclass
     #
 
-    def [](idx)
-      if @selector.key? :adjacent
-        to_a[idx] || element_class.new(@query_scope, {invalid_locator: true})
+    def [](value)
+      if value.is_a? Range
+        value.map { |i| self[i] }
```
jkotests: This won't work with negatives - eg `browser.divs[1..-1]`. I have used this type of Range before, so it would be nice to able to fix.

jkotests (changes requested):

lib/watir/element_collection.rb
titusfortner: I'm not following, we support [negative indexes](https://github.com/watir/watir/blob/master/spec/watirspec/elements/element_spec.rb#L309). Also, using `to_a` (e.g. , `divs.to_a[2..5]`) works with current code, we're just looking for a way to make it lazy loaded.

titusfortner: Ah, ok, I see, the `Range` instance can't evaluate the negative number without knowing the length ahead of time. This means that it doesn't make sense for it to be lazy loaded.

titusfortner:
@jkotests this is now just sending it along to `#to_a` so that user's don't have to change their code. No lazy loading is allowed when using Ranges.

jkotests (approved):

p0deje (approved):

Issue #739, opened 2018-05-29
Watir 6.11 collections broken for elements Watir doesn't have support for. 


Donavan:
In commit: 7ebfc838a14958b424525540fcf99ec922fb4819 to_a was modified to use  Watir.tag_to_class[tag_name].new which will explode if the tag name is not one Watir has full support support for.

For example:  @browser.elements(class: 'mat-spinner').count will blow up because the tag name of the element is actually mat-spinner.  Prior to this change I could treat them as a generic element and everything was fine.

titusfortner:
Yup, there's a better method to use there, just pushed the fix.

Also, would it be useful to add the custom Angular element classes to: https://github.com/titusfortner/watir_angular

Donavan:
In order to do that properly a small change needs made to watir.  Since the tag to class map was frozen, I needed a way to open it back up, add my stuff, then freeze it again. I also had to change the element_class_for so that it could handle element names with dashes in them.

This is what I'm using in my extension:

```ruby
module Watir
  class << self
    def open_tag_to_class
      @tag_to_class = tag_to_class.dup
    end

    def element_class_for(tag_name)
      tag_to_class[tag_name.to_s.tr('-', '_').to_sym] || HTMLElement
    end
  end
end
```

titusfortner:
One of our goals is to remove the need for users to monkey patch Watir to do the things they need to do. If you find yourself needing to do workarounds, let us know and we'll see if we can help.

titusfortner:
Let's consider this issue closed and we'll track in the PR & The Watir Angular issue

PR #741, opened 2018-05-29
freezing prevents extending

titusfortner:
I'm not sure I see the harm in not freezing this one. Removing it allows extensions to add custom element classes that Watir can use. (https://github.com/watir/watir/issues/739#issuecomment-392923066)


jkotests (approved):
I don't know the history of this one, but the change seems reasonable to me.

p0deje (approved):

Issue #731, opened 2018-05-17
Pass browser into Screenshot


titusfortner:
Based on this project: https://github.com/samnissen/watir-screenshot-stitch
It makes sense to pass in the browser variable to the `Screenshot` class instead of just the driver variable, since that would simplify the project code. (@samnissen)

PR #742, opened 2018-05-29
initialize Screenshot class with Browser instance not a Driver instance

titusfortner:
This is for #731
Which was created to improve the API for https://github.com/samnissen/watir-screenshot-stitch

I added the deprecation notice since `Watir::Screenshot` is a public class, so theoretically people could be initializing it directly (even though the likelihood is low).

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -156,5 +156,3 @@ require 'watir/elements/input'
 require 'watir/radio_set'
 
 require 'watir/aliases'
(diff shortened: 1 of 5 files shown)


jkotests (approved):

p0deje (approved):
All the watir-screenshot-stitch can be implemented with pure Selenium, but I see no harm in the proposed change.

PR #743, opened 2018-05-30
keep RSpec from running everything twice

titusfortner:
By Default RSpec runs everything using `#exist?` and `#exists?` and makes sure they both match. There should be no need for this extra work.


p0deje (approved):
The day I learned something new!

Issue #626, opened 2017-08-17
Element Existentialism


titusfortner:
Similar and related ideas:

* Selenium: `#find_element` & `#displayed?`
* Watir: `#exist?`, `#visible?`, `#present?`
* W3C: `"interactable"`, 

Many of these are non-intuitive for their actual behavior.
"exist" probably makes the most sense.
"present" is typically a synonym for "exist," but it also has the connotation of "being in view" so it can also be synonymous with "visible", but "displayed" would seem to be more correct/precise. In practice this is what we use for "interactable"
"visible" has the added complication of indicating to a user that they can see it on the page they are looking at without scrolling

As far as current behaviors go, I can't think of any reasons to favor #visible?, over #present?
I've been considering after looking through `watir-scroll`, is the idea of "inside viewport". 

How big of a deal would it be to deprecate `#visible?` in favor of `present?` in Watir 6, and then re-introduce it in Watir 7 as "exists, present, and entirely contained in the current viewport"? Actually, I kind of want to deprecate `#visible?` regardless because i think it is just unnecessary confusion that it is almost exactly the same but slightly and non-intuitively different.

I'm not completely sold that this is a good enough idea to break backward compatibility, but I thought I'd throw it out there to see what others think.

jkotests:
+1 for deprecating `#visible?`

I have always recommended the usage of `#present?` first. There are some places where `exists?` is important. I haven't had a reason to recommend the usage of `#visible?`. I'm not convinced that re-implementing it as "in the current viewport" is needed. At least I cannot think of a situation where the differentiation between requiring scrolling or not mattered.

titusfortner:
I have had people ask about it, or at least be surprised that it was saying visible when they couldn't see it. The Page Object gem's test for scrolling actually uses that method assuming it will be false if it isn't scrolled there. :)

That being said, I'm not sure it is actually real world useful to most people, either, especially if it causes any confusion to change it.

akostadinov:
In my tests I never needed this. But consider somebody wants to perform responsiveness testing. They may want to make sure some elements are visible/non-visible depending on browser window size. I don't mind if nother method name might be chosen for it though. Just saying, I don't have a horse.

husieo:
In my tests, I sometimes experience the need for `#accessible?`.
Example scenario:
1. Modal popup window is opened
2. Parent popup window calls nested popup window.
3. Interaction with nested popup window is completed, nested window is closed.
[...]

PR #744, opened 2018-05-30
Add deprecation notice for when we alias visible? and present?

titusfortner:
This is based on the discussion in the PR where I tried to deprecate `#visible?` (https://github.com/watir/watir/pull/734)

Which was based on this issue that I forgot about: https://github.com/watir/watir/issues/626 (See @jkotests  & @p0deje we did talk about actually deprecating it last fall).

Regardless, here's the code for prepping to alias `#visible?` to `#present?`

`#visible?` gets the deprecation notices it needs to inform users who are calling that method directly and then rescuing the method in their own code that the behavior will be changing.

`#present?` can't call `#visible?` for the deprecation notices to make sense

`#wait_for_present` can't call `#visible`, and now that `assert_exists` doesn't re-locate an element by itself, there is no performance hit to calling `wait_for_exists`, just the annoying `Watir.relaxed_locate?` conditionals still. :)

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -403,10 +403,12 @@ module Watir
     #
 
     def visible?
+      Watir.logger.warn "#visible? behavior will be changing slightly, consider switching to #present? (more details: http://watir.com/element-existentialism/)",
+                             ids: [:visible_element]
       assert_exists
       @element.displayed?
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
-      Watir.logger.deprecate "Checking `#visible?` or `#present? == false` to determine a stale element", "`#stale? == true`",
+      Watir.logger.deprecate "Checking `#visible? == false` to determine a stale element", "`#stale? == true`",
                              ids: [:stale_visible]
       reset!
       raise unknown_exception
@@ -432,7 +434,13 @@ module Watir
     #
 
     def present?
-      visible?
+      assert_exists
+      @element.displayed?
+    rescue Selenium::WebDriver::Error::StaleElementReferenceError
+      Watir.logger.deprecate "Checking `#present? == false` to determine a stale element", "`#stale? == true`",
+                             ids: [:stale_present]
+      reset!
+      false
     rescue UnknownObjectException, UnknownFrameException
       false
     end
@@ -537,8 +545,8 @@ module Watir
     end
 
     def wait_for_present
-      return visible? unless Watir.relaxed_locate?
-      return if present?
+      p = present?
+      return p if !Watir.relaxed_locate? || p
 
       begin
         @query_scope.wait_for_present unless @query_scope.is_a? Browser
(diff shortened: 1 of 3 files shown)


p0deje (approved):
> Which was based on this issue that I forgot about: #626 (See @jkotests & @p0deje we did talk about actually deprecating it last fall).

Apparently I wasn't involved in that discussion 😄

lib/watir/elements/element.rb
```diff
@@ -406,7 +406,11 @@ def visible?
       @element.displayed?
     rescue Selenium::WebDriver::Error::StaleElementReferenceError
       reset!
+      Watir.logger.deprecate "Getting an exception from #visible? when an Element does not exist", "#present?"
       raise unknown_exception
+    rescue UnknownObjectException, UnknownFrameException
+      Watir.logger.deprecate "Getting an exception from #visible? when an Element does not exist", "#present?"
```
jkotests: If users just switch `#visible?` to `#present?`, won't that be a change in behaviour? Shouldn't we provide a solution that allows users to maintain their current behaviour - eg `e.exists? ? e.visible? : raise(Watir::Exception::UnknownObjectException, "unable to locate element: #{e.inspect}")`

jkotests (changes requested):

lib/watir/elements/element.rb
titusfortner: Right, but one of the reasons for changing it is to change the behavior away from something that likely isn't actually desired, so I don't want them to just blindly replace `#visible?` with that ternary.

The other big concern I have with this approach is that some users might have tests that fail when an exception is rescued. So a user can run this code, all the tests pass and they don't get a deprecation notice. Then we update to alias `#visible?` with `#present?` and the System breaks, but the tests keep passing because it no longer throws the exception.

Really the only way to actually avoid that is to deprecate `#visible?`. I really want to deprecate it. Or at the very least throw a warning that "You are using `#visible?` when you probably want to use `#present?`". Maybe we can add a feature to the logger to make it so we can turn off specific warnings for people to silence specific warnings.

We also need to see how this affects PO gem before merging.

jkotests: I agree that deprecation would be easier (I had voted such last year). However, I think there are enough code bases and examples that warrants keeping `#visible?`. I'm guessing few of those usages are relying on raising the exception. 

Given it's hard to predict how people are using the exception, what if we just warn users that the behaviour is going to change? Don't provide a solution. Alternatively, provide a wiki or blog post link, which provides a longer discussion/examples.

titusfortner: Yeah, I think a general warning with a blog post link is a good idea.

titusfortner:
@jkotests how does this look for when I get the blog post written?

jkotests (approved):

jkotests:
@titusfortner , I think that looks good.

PR #746, opened 2018-06-01
Create visible_label locator to maintain backward compatibility

titusfortner:
This is an alternative to #717 

I don't love this locator, but I think I like this locator more than I like the alternatives (breaking backward compatibility and nested hashes).

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -172,7 +172,8 @@ module Watir
         end
 
         def process_label(label_key)
-          return unless @normalized_selector[label_key].kind_of?(Regexp) && selector_builder.should_use_label_element?
+          regexp = @normalized_selector[label_key].kind_of?(Regexp)
+          return unless (regexp || label_key == :visible_label)  && selector_builder.should_use_label_element?
 
           label = label_from_text(label_key)
           unless label # label not found, stop looking for element
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -120,8 +120,14 @@ module Watir
 
           @normalized_selector = selector_builder.normalized_selector
 
-          if @normalized_selector[:label]
-            process_label
+          if @normalized_selector.key?(:label)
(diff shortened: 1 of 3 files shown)


p0deje (approved):
I also don't quite like that solution, but I guess it's the only way.

sjieg:
Yes! This makes much more sense:
```
expect(browser.text_field(label: 'With hidden text')).to exist
expect(browser.text_field(label: 'With text')).not_to exist
expect(browser.text_field(visible_label: /With text/)).to exist
expect(browser.text_field(visible_label: /With hidden text/)).not_to exist
```
I've closed my PR.

lib/watir/locators/element/locator.rb
```diff
@@ -165,10 +171,10 @@ def create_filter_selector
           @filter_selector
         end
 
-        def process_label
-          return unless @normalized_selector[:label].kind_of?(Regexp) && selector_builder.should_use_label_element?
+        def process_label(label_key)
+          return unless @normalized_selector[label_key].kind_of?(Regexp) && selector_builder.should_use_label_element?
```
jkotests: This check will need to be different based on the label key:
* :label is only processed if a Regexp (no change)
* :visible_label is always processed - ie regardless of whether String or Regexp

Right now, `browser.text_field(visible_label: 'text')` will go into the XPath builder and be treated like a custom attribute.

jkotests (changes requested):
I'm still not a fan. I agree a nicer looking API than nested Hashes, but a nuisance in the code base.

lib/watir/locators/element/locator.rb
titusfortner: ugh, you're right. sigh.

titusfortner:
@jkotests  the fix was simpler than I feared. I'm thinking we don't even advertise this locator, and only mention it in the deprecation notice for the handful of people it might apply to.

jkotests:
@titusfortner , that sounds reasonable to me.

jkotests (approved):

PR #747, opened 2018-06-01
increase .flash speed and keep original bg-color of element

sjieg:
PR https://github.com/watir/watir/pull/695 could not be updated anymore, so hereby the new PR.

I've also added specs that check if the bg-color is the same after flashing. Tested in chrome and ff.

### DIFF
--- a/lib/watir/js_execution.rb
+++ b/lib/watir/js_execution.rb
@@ -43,11 +43,9 @@ module Watir
     # @return [Watir::Element]
     #
 
-    def flash(color: 'red', flashes: 5, delay: 0.2)
-      background_color = style("backgroundColor")
+    def flash(color: 'red', flashes: 10, delay: 0.05)
+      background_color = original_color = style("background-color")
       background_color = 'white' if background_color.empty?
-      element_color = element_call { execute_js(:backgroundColor, @element) }.strip
-      element_color = 'white' if element_color.empty?
 
       (flashes * 2).times do |n|
         nextcolor = n.even? ? color : background_color
@@ -55,7 +53,7 @@ module Watir
         sleep(delay)
       end
 
-      element_call { execute_js(:backgroundColor, @element, element_color) }
+      element_call { execute_js(:backgroundColor, @element, original_color) }
 
       self
     end
--- a/lib/watir/js_execution.rb
+++ b/lib/watir/js_execution.rb
@@ -33,22 +33,38 @@ module Watir
     #   browser.li(id: 'non_link_1').flash
     #   browser.li(id: 'non_link_1').flash(color: "green", flashes: 3, delay: 0.05)
     #   browser.li(id: 'non_link_1').flash(color: "yellow")
+    #   browser.li(id: 'non_link_1').flash(color: ["yellow", "green"])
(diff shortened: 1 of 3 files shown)


titusfortner (commented):
This flashes too quickly for me, but this was what the original watir-classic used, so I'm less against it.

What if we set it to 0.1 by default, and implemented: `element.flash :fast` (0.04sec) & `element.flash :slow` (0.2sec)?

jkotests (approved):
I don't notice a lot of difference between the delay times, so I'm fine with whatever you two agree on.

sjieg:
- Updated default `dalay: 0.1`
- Added possibility to allow multiple colors in 1 flash
- Added preset parameter
```
presets = {
          fast: { delay: 0.04 },
          slow: { delay: 0.2 },
          long: { flashes: 5, delay: 0.5 },
          rainbow: { flashes: 5, color: %w(red orange yellow green blue indigo violet) }
      }
return self.flash(presets[preset]) unless presets[preset].nil?
```
- Reworked how the flash loops to allow multiple colors.

I'm pretty sure this is not the best way to add presets, but this is my solution with my limited skills. Feedback is appreciated.
Let me know if the `:rainbow` preset is too much, I thought it was a nice _fun_ addition for TA'ers to talk about.

titusfortner:
This implementation looks solid to me. @jkotests can you double take on the new code?

lib/watir/js_execution.rb
```diff
@@ -33,29 +33,43 @@ def fire_event(event_name)
...
-      element_color = element_call { execute_js(:backgroundColor, @element) }.strip
-      element_color = 'white' if element_color.empty?
+      color = [color] if color.is_a? String
+      color.push(background_color)
 
-      (flashes * 2).times do |n|
-        nextcolor = n.even? ? color : background_color
+      (flashes * color.length).times do |n|
+        modulo = n % color.length
+        nextcolor = color[modulo]
```
jkotests: While I think this works, it'd be easier to read as:

~~~~~~~~
(color * flashes).each do |nextcolor|
  element_call { execute_js(:backgroundColor, @element, nextcolor) }
  sleep(delay)
end
~~~~~~~~

lib/watir/js_execution.rb
```diff
@@ -33,29 +33,43 @@ def fire_event(event_name)
...
+          rainbow: { flashes: 5, color: %w(red orange yellow green blue indigo violet) }
+      }
+      return self.flash(presets[preset]) unless presets[preset].nil?
+
+      background_color = original_color = style("background-color")
       background_color = 'white' if background_color.empty?
-      element_color = element_call { execute_js(:backgroundColor, @element) }.strip
-      element_color = 'white' if element_color.empty?
+      color = [color] if color.is_a? String
+      color.push(background_color)
```
jkotests: You can avoid the conditional by doing:

~~~~~~~~
color = Array(color).push(background_color)
~~~~~~~~

jkotests (changes requested):
Looks good, but I have some suggestions to make it easier to read.

sjieg:
Thanks for the feedback @jkotests! Always happy to learn more efficient notations.

jkotests (approved):

PR #748, opened 2018-06-02
Ignore warnings

titusfortner:
The idea here is to allow people to granularly ignore specific warnings if they don't think they apply to them.

I also added the ability to override the `progname` so that other projects can use this code...  Honestly, I *really* want to pull this code out of the Watir project and create a new gem called `WatirLogged`. Yes, mostly because I think it is an excellent pun, but I'm finding myself wanting to use this functionality in all of my gems.

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -46,11 +46,12 @@ module Watir
     end
 
     def always_locate_message
-      Watir.logger.warn <<-EOS
+      msg = <<-EOS.gsub("\n", " ")
 Watir#always_locate is deprecated; elements are always cached and will always
 be re-located if they go stale before use.
 Use Element#stale? or Element#wait_until(&:stale?) if needed for flow control.
       EOS
+      Watir.logger.warn msg, ids: [:always_locate, :deprecations]
     end
 
     #
@@ -63,11 +64,12 @@ Use Element#stale? or Element#wait_until(&:stale?) if needed for flow control.
     end
 
     def prefer_css_message
-      Watir.logger.warn <<-EOS
+      msg = <<-EOS.gsub("\n", " ")
 Watir#prefer_css is deprecated; all elements that can not be passed directly
 as Selenium locators will be translated to XPath. To continue using CSS Selectors
 require the watir_css gem - https://github.com/watir/watir_css
       EOS
+      Watir.logger.warn msg, ids: [:prefer_css, :deprecations]
     end
 
     #
--- a/lib/watir/logger.rb
+++ b/lib/watir/logger.rb
@@ -33,8 +33,9 @@ module Watir
       @ignored = []
     end
 
-    def ignore(id)
-      @ignored << id.to_s
+    def ignore(ids)
+      ids = [ids] unless ids.is_a? Array
+      @ignored.concat ids.map(&:to_s)
     end
 
     def output=(io)
--- a/lib/watir/logger.rb
+++ b/lib/watir/logger.rb
@@ -22,26 +22,36 @@ module Watir
 
     def_delegators :@logger, :debug, :debug?,
                    :info, :info?,
-                   :warn, :warn?,
+                   :warn?,
                    :error, :error?,
                    :fatal, :fatal?,
                    :level
 
-    def initialize
+    def initialize(progname="Watir")
       @logger = create_logger($stdout)
+      @logger.progname = progname
+      @ignored = []
+    end
+
+    def ignore(id)
+      @ignored << id.to_s
     end
 
     def output=(io)
-      # `Logger#reopen` was added in Ruby 2.3
-      if @logger.respond_to?(:reopen)
-        @logger.reopen(io)
-      else
-        @logger = create_logger(io)
-      end
+      @logger.reopen(io)
     end
 
     #
-    # For Ruby < 2.3 compatibility
+    # Only log a warn message if it is not set to be ignored.
+    #
+    def warn(message, ids: [], &block)
+      msg = ids.empty? ? "" : "[#{ids.map!(&:to_s).map(&:inspect).join(", ")}] "
+      msg += message
+      @logger.warn(msg, &block) unless (@ignored & ids).any?
+    end
+
+    #
+    # For Ruby < 2.4 compatibility
     # Based on https://github.com/ruby/ruby/blob/ruby_2_3/lib/logger.rb#L250
     #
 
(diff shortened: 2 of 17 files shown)


kapoorlakshya:
I would love to user `WatirLogged` in our internal gems as well. Currently following your approach with the `Logger` from `webdrivers` gem.

titusfortner:
And the code in the webdrivers gem is a copy of the Watir code which is a copy of the Selenium code, which @p0deje did an awesome job putting together last year. :)

Rajagopalan-M:
You can't say the code in WATIR is the copy of selenium, you can say WATIR
is now writing selenium script in WATIR-CLASSIC syntax.

titusfortner:
@Rajagopalan-M specifically referring to the Logger class.

Rajagopalan-M:
Oh okay

p0deje (approved):

titusfortner:
@p0deje do you have any qualms about my pulling the logging code out into a separate gem?

p0deje:
@titusfortner Why would you want to do that - it's just a single class implementing logger?

titusfortner:
1. There's nothing in this logging implementation that is specific to Watir/Selenium.

2. Because I'm finding myself just copy/pasting the code to get the same functionality in other projects (https://github.com/titusfortner/webdrivers/blob/master/lib/webdrivers/logger.rb). This PR potentially alleviates this issue, because I can now pass in a `progname` when initializing an instance of `Watir::Logger`, rather than copying it wholesale, but `webdrivers` for instance doesn't rely on Watir, and shouldn't force the inclusion of the gem in order to use this feature.

3. Seriously, `WatirLogged` is an amazing name for a logging gem :)

p0deje:
Nothing stops you from open-sourcing it, isn't it?

What is the pun? I don't get it.

titusfortner:
I started to write a new gem, then made this PR instead. Wouldn't make sense to have the code in a separate gem if you & Justin were opposed to pulling it out of Watir proper, and/or don't enjoy my pun as much as I do.

The pun is just a play on the term http://www.dictionary.com/browse/waterlogged
We have all of these other project names that play off the name of Watir as Water without having any additional semantic meaning, but in this case "Logged" actually means what the gem is doing ("Logging"), so it's extra amusing to me.

lib/watir.rb
```diff
@@ -63,11 +64,12 @@ def prefer_css?
     end
 
     def prefer_css_message
-      Watir.logger.warn <<-EOS
+      msg <<-EOS
```
jkotests: Same comment as above for #always_locate_message.

lib/watir.rb
```diff
@@ -46,11 +46,12 @@ def always_locate?
     end
 
     def always_locate_message
-      Watir.logger.warn <<-EOS
+      msg <<-EOS
```
jkotests: This will throw an "undefined method `msg' for Watir:Module" exception due to the missing assignment operator. It will also format a bit funny due to the return carriages:

~~~~~~~~
2018-06-06 23:03:10 WARN Watir [:always_locate, :deprecations] Watir#always_locate is deprecated; elements are always cached and will always
be re-located if they go stale before use.
Use Element#stale? or Element#wait_until(&:stale?) if needed for flow control.
~~~~~~~~

Would suggest:

~~~~~~~~
def always_locate_message
  msg = <<-EOS.gsub("\n", " ")
Watir#always_locate is deprecated; elements are always cached and will always
be re-located if they go stale before use.
Use Element#stale? or Element#wait_until(&:stale?) if needed for flow control.
  EOS
  Watir.logger.warn msg, ids: [:always_locate, :deprecations]
end
~~~~~~~~

lib/watir/logger.rb
```diff
@@ -22,26 +22,36 @@ class Logger
...
+      @logger.reopen(io)
+    end
+
+    #
+    # Only log a warn message if it is not set to be ignored.
+    #
+    def warn(message, ids: [], &block)
+      msg = ids.empty? ? "" : "[#{ids.map(&:inspect).join(", ")}] "
+      msg += message
+      @logger.warn(msg, &block) unless @ignored.include?(:id)
```
jkotests: Probably don't want this to be based on `:id`. Should be `(@ignored & ids).any?` ?

lib/watir/logger.rb
```diff
@@ -22,26 +22,36 @@ class Logger
...
                    :level
 
-    def initialize
+    def initialize(progname="Watir")
       @logger = create_logger($stdout)
+      @logger.progname = progname
+      @ignored = []
+    end
+
+    def ignore(id)
```
jkotests: I would suggest supporting multiple arguments. Right now, to ignore multiples, you would have to do:

~~~~~~~~
Watir.logger.ignore(:class_array)
Watir.logger.ignore(:always_locate)
~~~~~~~~

It would be nice to be able to do:

~~~~~~~~
Watir.logger.ignore(:class_array, :always_locate)
~~~~~~~~

spec/unit/logger_spec.rb
```diff
@@ -42,5 +42,21 @@ module Watir
       message = /WARN Watir \[DEPRECATION\] #old is deprecated\. Use #new instead\./
       expect { Watir.logger.deprecate('#old', '#new') }.to output(message).to_stdout
     end
+
+    it 'allows to selectively ignore warnings with Strings' do
+      Watir.logger.ignore("old deprecated")
+      expect { Watir.logger.deprecate('#old', '#new', ids: ['old deprecated']) }.to_not output.to_stdout
+    end
```
jkotests: I'm not sure I understand the intent here. There don't appear to be any String ids at the moment. Is this for a future usage? Wouldn't it be weird to mix String and Symbol ids?

lib/watir.rb
```diff
@@ -63,11 +64,12 @@ def prefer_css?
     end
 
     def prefer_css_message
-      Watir.logger.warn <<-EOS
+      msg <<-EOS
 Watir#prefer_css is deprecated; all elements that can not be passed directly
 as Selenium locators will be translated to XPath. To continue using CSS Selectors
 require the watir_css gem - https://github.com/watir/watir_css
       EOS
+      Watir.logger.warn msg, ids: [:prefer_css]
```
jkotests: Should include `:deprecations`.

jkotests (changes requested):

jkotests:
I have no preference on whether this is included here vs another gem. Though calling the gem "WatirLogged" suggests it is related to Watir. Might be bad marketing if you want an uptake in other projects.

spec/unit/logger_spec.rb
titusfortner: The idea is to allow other implementations to use String or Symbol instances. Not certain it's the best idea, the updated commits make my intention more clear for my current desired behavior.

lib/watir/logger.rb
titusfortner: I implemented as an array instead of args

titusfortner:
@jkotests 
> Might be bad marketing

Well, I was thinking about it like: "A Logging gem maintained by Watir project," so more of a brand stamp than an indication of functionality. While it is in the Watir project itself, it *has to be used with Watir. :) So any separate gem would be an improvement in flexibility regardless of the name.

Made the rest of the changes. Let me know what you think.

jkotests (approved):

jkotests:
@titusfortner, the changes look good to me.

PR #749, opened 2018-06-11
deal with elements going stale during filtering

titusfortner:
This addresses [an issue](https://gist.github.com/jff05/c4b6bc0d91707009db077bd97f50bdf7) that was raised on Slack.

I don't love this code, does anyone have any suggestions for improving it?

### DIFF
--- a/lib/watir/locators/element/locator.rb
+++ b/lib/watir/locators/element/locator.rb
@@ -69,8 +69,17 @@ module Watir
           what = add_regexp_predicates(what) if how == :xpath
 
           if filter == :all || !@filter_selector.empty?
-            elements = locate_elements(how, what, @driver_scope) || []
-            filter_elements(elements, filter: filter)
+            retries = 0
+            begin
+              elements = locate_elements(how, what, @driver_scope) || []
+              filter_elements(elements, filter: filter)
+            rescue Selenium::WebDriver::Error::StaleElementReferenceError
+              retries += 1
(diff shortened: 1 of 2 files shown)


p0deje:
Can we have a test for that issue first? It's hard to improve this code without being able to test it.

jff05:
In case it helps you build your test, I updated my [issue ](https://gist.github.com/jff05/c4b6bc0d91707009db077bd97f50bdf7)to provide you with spec, objects and HTML samples of what I was using.

jkotests:
We also do a rescue for `Selenium::WebDriver::Error::StaleElementReferenceError` in `#locate`:

~~~~~~~~ruby
def locate
  using_selenium(:first) || using_watir(:first)
rescue Selenium::WebDriver::Error::NoSuchElementError, Selenium::WebDriver::Error::StaleElementReferenceError
  nil
end
~~~~~~~~

Would the new rescue in `#using_watir` change this? I'm not exactly sure what situations we had been trying to rescue before.

titusfortner:
@jkotests we have it in `#locate` but not in `#locate_all` which is what is triggering this.
Additionally, we retry Stale Element exceptions in other locations, it seems like we shouldn't just return nil if it happens in the middle of the filtering process.

@p0deje I added a spec as requested. :)

p0deje (approved):

jkotests (approved):

PR #752, opened 2018-06-26
Adds getters  for query scope and selector to element collections.

jfitisoff:
Individual elements have getters for :query_scope and :selector but collections don't. Seemed like a low risk thing so I added the getters. I've actually been using them and this change would allow me to get rid of some logic to get the vars differently when the object is a collection. 

All specs are passing except "Browser #new with parameters accepts switches argument," which I've seen fail before this change -- think that's probably related to my laptop's screen resolution.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -5,6 +5,8 @@ module Watir
   #
 
   class ElementCollection
+    attr_reader :query_scope, :selector
+
     include Enumerable
     include Locators::ClassHelpers
 


titusfortner:
@query_scope for Element is not currently publicly readable, and the only reason for @selector to be readable is because I wanted to improve the performance of a couple weird Table specific methods. Do you have a use case in mind for this?

titusfortner:
I'm going to close this one for now; let me know if you have any additional use case for this.

jfitisoff:
Sorry Titus, totally missed the comments. I'm writing wrapper classes around the watir element classes (for a PO library) so it's not a common use case. I'll just keep handling the collections the way that I currently am.  Thanks for following up.

titusfortner:
@jfitisoff no worries. If you have a good use case we can do this, just wanted to make sure there was a good reason to open it up.

#allthepageobjects lol, everyone seems to write their own.

jfitisoff:
yup. :-)

PR #755, opened 2018-07-06
deprecate locating windows by index

titusfortner:
Per w3c, drivers do not need to return window handles in any order. Having `:index` as a locator option gives the incorrect impression that windows are ordered (which likely they were with original Watir).

Perhaps we can provide an alternate way to implement the intention "switch to the other window"

### DIFF
--- a/lib/watir/has_window.rb
+++ b/lib/watir/has_window.rb
@@ -52,6 +52,25 @@ module Watir
       @original_window ||= window
     end
 
+    #
+    # Waits for and returns second window if present
+    # See Window#use
+    #
+    # @example
+    #   browser.switch_window
+    #
+    # @return [Window]
+    #
+
+    def switch_window
+      win = window
+      wins = windows
+      wait_until { (wins = windows) && wins.size > 1 } if wins.size == 1
+      raise StandardError, "Unable to determine which Window to Switch to" if wins.size > 2
+      wins.find { |w| w != win }.use
+      window
+    end
+
     private
 
     def filter_windows(selector, windows)
--- a/lib/watir/window.rb
+++ b/lib/watir/window.rb
@@ -215,6 +215,8 @@ module Watir
       if @selector.empty?
         nil
       elsif @selector.key?(:index)
+        Watir.logger.deprecate "Using :index as a selector for Window", ":title or :url",
+                               ids: [:window_index]
         @driver.window_handles[Integer(@selector[:index])]
       else
         @driver.window_handles.find { |wh| matches?(wh) }
--- a/spec/watirspec/html/window_switching.html
+++ b/spec/watirspec/html/window_switching.html
@@ -2,11 +2,21 @@
 <html>
   <head>
     <title>window switching</title>
+    <script>
+        function windowOpenDelayed(timeout) {
+            setTimeout(function() {
+                window.open("closeable.html");
+            }, timeout);
+        }
+    </script>
   </head>
 
   <body>
     <p>
       Click <a id="open" href="#" onclick='window.open("closeable.html")'>here</a> to open a new window.
+    </p>
+    <p>
+      Click <a id="delayed" href="#" onclick=windowOpenDelayed(1000)>here</a> to open a new window after a delay.
     </p>
   </body>
 </html>
--- a/spec/watirspec/window_switching_spec.rb
+++ b/spec/watirspec/window_switching_spec.rb
@@ -2,6 +2,7 @@ require "watirspec_helper"
 
 describe "Browser" do
   before do
+    @msg = /Using :index as a selector for Window is deprecated/
     url = WatirSpec.url_for("window_switching.html")
     browser.goto url
     browser.a(id: "open").click
@@ -45,8 +46,8 @@ describe "Browser" do
     end
 
     it "finds window by :index" do
-      w = browser.window(index: 1).use
-      expect(w).to be_kind_of(Watir::Window)
+      expect { @w = browser.window(index: 1).use }.to output(@msg).to_stdout_from_any_process
+      expect(@w).to be_kind_of(Watir::Window)
     end
 
     it "should not find incorrect handle" do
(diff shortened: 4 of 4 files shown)


abotalov:
As a note, Capybara's window API is good. In particular, [`window_opened_by`](https://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Session#window_opened_by-instance_method) method.

titusfortner:
Oh interesting. I like that usage of blocks. That being said, if 99.9% of users are only ever going to interact  with one additional window, I'm kind of leaning toward a straight up `Browser#switch_window`method that will wait until the window is opened and raise an exception if it is not.

abotalov:
I think `window_opened_by` isn't more complex than `switch_window` and could work in more circumstances.
Behavior of `switch_window` that switches window to a next one returned by `window_handles` would be surprising and not useful if more than two windows are used.

titusfortner:
I don't like unnecessary usage of blocks by end users. This example isn't a bad usage at least (I hate when blocks are used exclusively for controlling order of operations). But I think it is also more than what is needed for the vast majority of use cases.

I"ll write code for the method I'm thinking.

Also, I should look into stealing more things from @twalpole in general. :)

Rajagopalan-M:
I thought of learning capybara, It seems to have many feature . What is
your opinion on this?

titusfortner:
Heh. My opinions are many and varied. I know you've seen the other issue that went into more detail on the various distinctions between Watir and Capybara.

Rajagopalan-M:
Okay thanks

lib/watir/has_window.rb
```diff
@@ -52,6 +52,25 @@ def original_window
...
+
+    def switch_window
+      win = window
+      wins = windows
+      wait_until { (wins = windows) && wins.size > 1 } if wins.size == 1
+      raise StandardError, "Unable to determine which Window to Switch to" if wins.size > 2
+      wins.find { |w| w != win }.use
+      window
+    end
+
```
titusfortner, line 73: This looks inefficient, but it minimizes wire calls

titusfortner:
@abotalov I updated this PR with the `#switch_window` implementation I was considering. It only works for 2 windows, it will wait for the second window to open, and it figures out what the other window is regardless if it's the first or the last in the handles Array.

I think `#window_opened_by` is a cool feature though, and I'd be up for including it if you want to add it. Need to see what Alex & Justin think.

p0deje (approved):

titusfortner:
@jkotests any opinions on this?

jkotests (approved):
I'm on the fence about this.

"drivers do not need to return window handles in any order" isn't new. However, I haven't run into a situation where the order wasn't consistent. To my knowledge, the Watir specs haven't had issues using `:index`. If we remove `browser.window(index: 1)`, people would switch to `browser.windows[1]`. I don't feel like we gain much by the deprecation.

That said, I don't have any usages of `:index` in my code base, so I won't personally miss it.

titusfortner:
@jkotests apparently it is not consistent in IEDriver, which is why Jim Evans pushed to make sure that wording was in the spec.

You're right, people are more likely to use `browser.windows.last` or the like. Now you're making me want to create a `WindowCollection` class to protect against this. :-P (just kidding). (mostly).

jkotests:
I still prefer Watir-Classic over IEDriver, so don't have much experience with it. Though in a quick test, the window handles seemed to be returned in a consistent order. I'm sure Jim's right, but it isn't an obvious problem to me.

I'm tempted to just leave things as-is. If `:index` is working for people, let them use it. If it doesn't work, hopefully someone one will give an example and we can come up with a good solution or get rid of it. Though like I said before, I'm not strongly opposed to deprecating now.

titusfortner:
I guess I don't want to encourage a situation where a user thinks they are getting a particular order when they are not... But maybe I'll check with Jim Evans first to see what the issue is in IE.

Maybe I'll skip that for now. `#switch_window` feel like a good add regardless?

Rajagopalan-M:
Watir-classic over IE, yes only in the case of IE watir-classic performs an
excellent Job, I don't know why it's deprecated, I don't find any problem
with Watir-classic except writing xpath. By the way are you still using
WATIR-classic?

jkotests:
@titusfortner , `#switch_window` could be useful, though I don't personally have a use case for it. I've always preferred `#use` with a block since it automatically returns back to the original window.

@Rajagopalan-M , Watir-Classic is deprecated because no one wanted to maintain it. It'll also stop working once IE11 hits end-of-life. For the test suite I use, I keep it Watir-Classic and Watir compatible. Watir-Classic's `#attach` method is invaluable to my manual testing efforts, but I've made it Watir compatible for future-proofing.

titusfortner:
I saw @jimevans post that the window order is alphabetical in IE, which means that when opening a new window, the order will be correct 50% of the time. :)

jkotests:
Another common approach to finding windows is `b.windows.last.use` (eg https://stackoverflow.com/a/51365776/1200545). The deprecation of `:index`, wouldn't catch this.

titusfortner:
Right, so if I were serious about this I would have to make `browser.windows.class # => WindowCollection`, and then throw a deprecation notice that using any of the indexing methods on that class might not work consistently. :)

jkotests:
Doesn't feel like a top priority. I'd kick the can down the road.

kapoorlakshya:
I have a test case which deals with four browser windows, and two of them have dynamically generated titles. We use an approach without `index`, but `index` could be useful in this case.

Our approach is to immediately switch to the new window with `b.windows.last.use` and then use `b.windows.last.close` with `b.windows.last.use` to switch to the previous tab/window. A better solution could be to have  `previous_window` and `next_window` methods to switch between windows relative to the current `window`. What do you think @titusfortner?

@jkotests What do you mean by "catch this" in this context?

> Another common approach to finding windows is b.windows.last.use (eg https://stackoverflow.com/a/51365776/1200545). The deprecation of :index, wouldn't catch this.

You mean it won't affect it?

titusfortner:
@kapoorlakshya the issue is that `b.windows.last` might not be the last window opened. In Internet Explorer the windows are stored in an array "alphabetically" whatever that means. So "next_window" and "previous_window" won't make sense unless Watir is storing this information separately from Selenium....

Which actually might be a better solution if we can do that.

I'm going to punt on this PR for this next release and think through some alternatives.

PR #756, opened 2018-07-06
deprecate locating elements by ordered parameters

titusfortner:
Using ordered pair locators (how, what) makes it more difficult to add additional selectors if necessary. I'd like to see Watir standardize on always requiring Hashes. (e.g. `:id, 'foo'` be replaced in user code with `id: 'foo'`)

### DIFF
--- a/lib/watir/container.rb
+++ b/lib/watir/container.rb
@@ -36,6 +36,9 @@ module Watir
     def extract_selector(selectors)
       case selectors.size
       when 2
+        Watir.logger.deprecate "Using ordered parameters to locate elements (:#{selectors.first}, #{selectors.last.inspect})",
+                               "{#{selectors.first}: #{selectors.last.inspect}}",
+                               ids: [:selector_parameters]
         return { selectors[0] => selectors[1] }
       when 1
         obj = selectors.first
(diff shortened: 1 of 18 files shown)


p0deje (approved):
👍

lib/watir/container.rb
```diff
@@ -36,6 +36,7 @@ def elements(*args)
     def extract_selector(selectors)
       case selectors.size
       when 2
+        Watir.logger.deprecate "Using ordered parameters to locate elements", "a Hash"
```
jkotests: To make it easier to find places that need corrections, can we include the selector in the message - eg `"Using ordered parameters, #{selectors}, to locate elements"` ?

jkotests (approved):

lib/watir/container.rb
titusfortner: Ah yes, we can do one better:
```
"Using ordered parameters to locate elements (:#{selectors.first}, #{selectors.last.inspect})", 
"{#{selectors.first}: #{selectors.last.inspect}}"
```

jkotests: Looks great.

PR #757, opened 2018-07-07
allow messages to be procs to minimize unnecessary wire calls

titusfortner:
Gives more flexibility & doesn't call inspect every single time when not needed.

### DIFF
--- a/lib/watir/wait.rb
+++ b/lib/watir/wait.rb
@@ -46,7 +46,7 @@ module Watir
           result = yield(object)
           return result if result
         end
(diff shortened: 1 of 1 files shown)


p0deje (approved):

p0deje:
Actually, can you explain how this doesn't call inspect every time? From what i can see, you still evaluate block which calls inspect.

titusfortner (commented):
#message_for only gets called when the timeout exception is raised. When you put the #inspection call in a String, it gets called every time a wait is used, even if it never times out.

PR #760, opened 2018-08-02
Contenteditable elements

titusfortner:
When I was looking at our WYSIWYG guide I realized that Watir should be able to directly support using our editing methods with elements that have [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) attribute set.

Ideally we would automatically extend these, but I'm not sure we want to add the overhead this would require to text_fields, etc. This seems like a good compromise.


p0deje (approved):
I'd say it'd be better to automatically extend if contenteditable is set, but 👍

lib/watir/elements/element.rb
```diff
@@ -529,6 +529,16 @@ def reset!
...
     end
 
+    #
+    # Allows this instance to respond to editing methods
+    #
+    # @see Watir::UserEditable
+    #
+
+    def make_editable
+      self.extend UserEditable
```
jkotests: Does `#append` work for you? If the element isn't already in edit mode, the value appends to the beginning for me (was using Chrome).

Do you care that `#set!` won't work in this case?

spec/watirspec/elements/element_spec.rb
```diff
@@ -704,4 +704,13 @@
       expect(browser.p.attribute_value(:data_type)).to eq "ruby-library"
     end
   end
+
+  describe '#make_editable' do
+    it "allows text to be set" do
+      editing = browser.div(id: 'contenteditable')
+      editing.make_editable
+      editing.set("foo")
```
jkotests: Given that most calls to `#make_editable` would be followed by `#set`, should we make this easier to chain - ie `editing.make_editable.set("foo")`? Or do we tell people to use `editing.tap(&:make_editable).set("foo")`?

titusfortner:
@p0deje figured out how to use the method_missing to make the extra method unnecessary
@jkotests append now throws an exception

Amazing what you can do when you write all the specs you want implemented first

PR #761, opened 2018-08-03
don't publish documentation for watirspecs

titusfortner:
I think the watirspecs are unnecessary to include in the documentation. What do you think?


p0deje (approved):
Sounds right.

PR #763, opened 2018-08-07
allow wait methods to wait for values of any attribute

titusfortner:
When debugging the question on watir-general it got me thinking about why we can't make the wait_while & wait_until methods more powerful.

Get a Hash of everything that doesn't match one of the predefined keywords and then wait for them to have the value provided. This code supports RegExp & String, not sure the use cases for other types, so let me know if you think it should be opened up,


jkotests (approved):
I like the idea.

p0deje (approved):

PR #764, opened 2018-08-08
Rubocop Updates

titusfortner:
This was created with support for Ruby 2.3
I added the exceptions that made sense to me in .rubocop.yml

1. Let me know if we'd prefer to not make any of these changes
2. Did we say we wanted to wait for Watir 7 to support 2.3, or can we pull the trigger on that now?
3. I can go either way on merging this now or having this be the start of a Watir 7 branch.


jkotests:
We said that we could go to Ruby 2.3 in a minor release. We just had to get it added to the gemspec.

titusfortner:
Ok, then it was Selenium itself that we said we were waiting for major version bump. Thanks.

jkotests (approved):

p0deje (approved):

titusfortner:
Updated & Merged

PR #770, opened 2018-09-01
add message argument to wait_until-while_present methods

kubiksamek:


lib/watir/wait.rb
```diff
@@ -172,26 +172,28 @@ def wait_while(deprecated_timeout = nil, deprecated_message = nil, timeout: nil,
     #   browser.text_field(name: "new_user_first_name").wait_until_present
     #
     # @param [Integer] timeout seconds to wait before timing out
+    # @param [Float] interval seconds to wait before each try
```
titusfortner, line 175: These are no longer parameters, they are keywords, so they don't get documented like this.

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+### 6.12.1 (2018-08-31)
```
titusfortner: Maintainers will update the changes file prior to a release, no need to do it in a PR.

lib/watir/wait.rb
```diff
@@ -172,26 +172,28 @@ def wait_while(deprecated_timeout = nil, deprecated_message = nil, timeout: nil,
...
 
-    def wait_until_present(deprecated_timeout = nil, timeout: nil, interval: nil)
+    def wait_until_present(deprecated_timeout = nil, timeout: nil, interval: nil, message: nil)
       if deprecated_timeout
         Watir.logger.deprecate "Using arguments for #wait_until_present", "keywords", ids: [:timeout_arguments]
         timeout = deprecated_timeout
       end
       if self.is_a? Watir::Element
-        wait_until(timeout: timeout, interval: interval) do
+        wait_until(timeout: timeout, interval: interval, message: message) do
```
titusfortner: let's add a default message here. The default message in `#wait_until` is "waiting for true condition" (because the condition is in the block). In this case, though, we know what the actual condition is, so we can specify it. Also, this needs to be a Proc instance similar to: https://github.com/watir/watir/pull/770/files#diff-0ac8030a39db381bf5dd122f0bfdd337R126

lib/watir/wait.rb
```diff
@@ -203,26 +205,28 @@ def wait_until_present(deprecated_timeout = nil, timeout: nil, interval: nil)
     #   browser.text_field(name: "abrakadbra").wait_while_present
     #
     # @param [Integer] timeout seconds to wait before timing out
+    # @param [Float] interval seconds to wait before each try
```
titusfortner, line 210: see above

lib/watir/wait.rb
```diff
@@ -203,26 +205,28 @@ def wait_until_present(deprecated_timeout = nil, timeout: nil, interval: nil)
...
 
-    def wait_while_present(deprecated_timeout = nil, timeout: nil, interval: nil)
+    def wait_while_present(deprecated_timeout = nil, timeout: nil, interval: nil, message: nil)
       if deprecated_timeout
         Watir.logger.deprecate "Using arguments for #wait_while_present", "keywords", ids: [:timeout_arguments]
         timeout = deprecated_timeout
       end
       if self.is_a? Watir::Element
-        wait_while(timeout: timeout, interval: interval) do
+        wait_while(timeout: timeout, interval: interval, message: message) do
```
titusfortner: see above about default message

titusfortner (changes requested):
Excellent, thanks for the first submission. Here are a couple things, let me know if they don't make sense.

lib/watir/wait.rb
kubiksamek: Ok, How should I document these?

CHANGES.md
kubiksamek: OK

lib/watir/wait.rb
titusfortner: you don't need to. :)

kubiksamek: Oh, ok then.
Just to know, why keywords should not be documented? Ruby style? Thanks.

kubiksamek: And just one more Q, should I remove `timeout` arg document line? (was there already)

titusfortner: Ok, I'm wrong, it looks like it should be using param: 
https://www.rubydoc.info/gems/yard/file/docs/Tags.md#option
But, looks like we should be specifying the name of it before the type:
https://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md

titusfortner: timeout as an ordered parameter is still available, just deprecated

lib/watir/wait.rb
kubiksamek: Why should I add default message?
Default is nil, so If no message is provided, `nil` is passed to `wait_until` method and then `message ||= Proc.new { |obj| "waiting for true condition on #{obj.inspect}" }` will be executed or am I wrong?

titusfortner: because we know the condition ("present" or "not present"), so there is no reason to leave it a vague "true condition" or "false condition".

kubiksamek: Added

titusfortner:
Rebased and merged: a70ce65ed1faa312c3245febc586f5918544aa1f
(I updated the default messages with the added info).

Thanks for the commit!

PR #771, opened 2018-09-01
Fix Context Methods

### LINKED ISSUE #759: Waiting for element scope doesn't happen
## Meta -
I use nerodia but it mostly has the same implementation as Watir so I think bug is also reproducible for Watir.

## Expected Behavior -
I think Watir should do one (or both) of these things:
* Wait for element scope to be on the page (and raise TimeoutError if it's not present)
* Return empty list when iterating through `elements` if element scope is not on the page (so nothing will be raised in example below)

## Actual Behavior -
If query scope isn't present on the page, UnknownObjectException is raised.

## Steps to reproduce -
```
from nerodia import browser

br = browser.Browser()
br.goto("http://example.com")
els = br.element(id="a").elements(class_name="b")
for el in els:
  print el.text
```
This raises `UnknownObjectException: unable to locate element: #<HTMLElement: located: False; {'id': 'a'}>`

### LINKED ISSUE #769: Iterating a collection fails if elements are removed during iteration
I use Nerodia but Watir has the same issue

## Expected Behavior -
If an element is removed during iteration, iteration fails.
I think if some element turns out to be stale, iteration should start from the beginning.

## Actual Behavior -
If some element is removed during iteration, an exception is raised.

## Steps to reproduce -
```
import nerodia

nerodia.default_timeout = 3
br = nerodia.browser.Browser()
br.goto("https://fiddle.jshell.net/SMZNH/177/show")
iframe = br.iframe()
for i in xrange(0, 100):
  iframe.button(text="add elements!").click()
  iframe.button(text="remove elements!").click()
  print len(iframe.elements(tag_name="p"))
```
fails with `UnknownObjectException: timed out after 3 seconds, waiting for #<HTMLElement: located: False; {'tag_name': 'iframe'} --> {'index': 0, 'tag_name': 'p'}> to be located`. In this situation I would expect Watir to start from the beginning, see that there are no elements anymore, and return 0.

titusfortner:
1. Locators class shouldn't need to know anything about scope
2. Ensuring Context shouldn't also locate; it only needs to locate the parent
3. This also required fixing weirdness with Frames and FramedDriver. This code allows frames to use more of Element super class methods and logic. The only weirdness is in having to treat `#find_element` and `#find_elements` differently.

This should let us fix #759 though I wanted to get this refactor passing all the tests first

### DIFF
--- a/lib/watir/browser.rb
+++ b/lib/watir/browser.rb
@@ -311,13 +311,17 @@ module Watir
     #
 
     def assert_exists
-      ensure_context
+      locate
       return if window.present?
       raise Exception::NoMatchingWindowFoundException, "browser window was closed"
     end
 
-    def ensure_context
+    def locate
       raise Exception::Error, "browser was closed" if @closed
+      ensure_context
+    end
+
+    def ensure_context
       driver.switch_to.default_content unless @default_context
       @default_context = true
     end
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -168,8 +168,7 @@ module Watir
     end
 
     def locate_all
-      @locator ||= build_locator
-      @elements ||= @locator.locate_all
+      build_locator.locate_all
     end
 
     def element_class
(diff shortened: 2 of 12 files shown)


titusfortner:
Added the commit to fix #759

p0deje (approved):

titusfortner:
Closing in favor of #773

PR #772, opened 2018-09-02
Html5 2

titusfortner:
I hate dealing with webidl, since I pretty much have no idea what I'm doing. :)

Alex, for some reason I'm having an issue with extended attributes. This monkey patch I created "works," but I have no idea why, and I have no idea if it is the right fix. If it is the right fix, we probably need to submit it to the webidl gem.

Also, I have no idea how to get SVG2 to parse. I ran the code we have against the version of SVG spec that I'm pretty sure we used last time, without success. I think we need to reference the exact spec we'r running against instead of the working draft so at least we can (theoretically) reproduce previous runs.

Any ideas on what isn't working with SVG? Is their doc not formatted correctly any longer?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,7 +17,7 @@ namespace :spec do
 end
 
 {
-  html: 'https://www.w3.org/TR/html/single-page.html',
+  html: 'https://www.w3.org/TR/html52/single-page.html',
   svg: 'http://www.w3.org/TR/SVG2/single-page.html'
 }.each do |type, spec_uri|
   namespace type do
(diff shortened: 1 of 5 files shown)


p0deje:
There was a bug in WebIDL gem which is fixed in 
https://github.com/jarib/webidl/commit/60cbc7d80ba529abd434af08d62753bde9fd3d79.

After the fix, specification updates just fine so I already pushed in e0afc13981f9b21c0d446befbf83d44d59a6b6b5.

In regards, to SVG - it has always been flooded with incorrect IDLs, so I had to manually fix it to be IDL-compliant. I'll have to do that again :)

p0deje:
Actually, SVG just needed a proper `includes` statement support in WebIDL so it should "just work" since ff3a3d4fdd392d673f60696dfdb0cd819c4d55c4.

titusfortner:
Thanks for fixing this, @p0deje

PR #775, opened 2018-09-11
Feature - Add Watir::Element#attribute_values and #attribute_list which return attributes from an element.

kapoorlakshya:
Tagging @jkotests and @titusfortner.

**Question**: Should the attributes (keys) be returned as `Symbol` or `String` (from JS)? I prefer `Symbol`, but maybe `String` is more "pure" ¯\_(ツ)_/¯. Currently returns `String`.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -275,6 +275,43 @@ module Watir
     alias attribute attribute_value
 
     #
+    # Returns all attribute values. Attributes with special characters are returned as String,
+    # rest are returned as a Symbol.
+    #
+    # @return [Hash]
+    #
+    # @example
+    #   browser.pre(id: 'rspec').attribute_values
+    #   #=> {class:'ruby', id: 'rspec' }
+    #
+
+    def attribute_values
+      result = element_call { execute_js(:attributeValues, @element) }
+      result.keys.each do |key|
+        next unless key == key[/[a-zA-Z\-]*/]
+
+        result[key.tr('-', '_').to_sym] = result.delete(key)
+      end
+      result
+    end
+    alias attributes attribute_values
+
+    #
+    # Returns list of all attributes. Attributes with special characters are returned as String,
+    # rest are returned as a Symbol.
+    #
+    # @return [Array]
+    #
+    # @example
+    #   browser.pre(id: 'rspec').attribute_list
+    #   #=> [:class, :id]
+    #
+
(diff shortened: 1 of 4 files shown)


jkotests:
Using Symbol keys might be nicer as it would match the appearance of locators. However, we'd have to decide what to do with attributes that mix dashes and underscores - eg:
* `id` becomes `:id`
* `data-type` becomes `:data_type`
* `data-two_part` becomes what?

We started supporting the mixed dash/underscore via Symbols vs Strings. Would we do the same here?

Given:

~~~~~~~~
<div id="a" data-type="b" data-two_part="c"></div>
~~~~~~~~

Would we get:

~~~~~~~~
browser.div.attribute_values
#=> {:id => "a", :data_type => "b", "data-two_part" => "c"}
~~~~~~~~

Seems weird to me, but at least consistent with the rest of the code base.

titusfortner:
yeah, not that I see a pressing use case, but it would make sense to be able to use the same value as a locator and a method and a key for this Hash.
Something like this?

```ruby
      result = element_call { execute_js(:attributeValues, @element) }
      result.keys.each do |key|
        next if key.include?('_')
        result[key.tr('-','_').to_sym] = result.delete(key)
      end
      result
```

abotalov:
As a note, [HTML spec says](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2) attribute name can contain mostly any characters:
> Attribute names must consist of one or more characters other than controls, U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and noncharacters.

https://stackoverflow.com/a/926136

Probably it makes sense to convert strings that contain only characters from `[a-zA-Z_\-]`.

titusfortner:
Ah, that's interesting. Maybe we want to keep the key as a string unless it has "special" characters in it. Something like this:

```ruby
  def attribute_values
      result = element_call { execute_js(:attributeValues, @element) }
      result.keys.each do |key|
        next unless key == key[/[a-zA-Z\-]*/]
        result[key.tr('-','_').to_sym] = result.delete(key)
      end
      result
  end
```

kapoorlakshya:
I have updated the method definition as per the discussions in this PR, added unit tests + updated HTML to test for special characters in an attribute, verified rubocop and doctest pass, and squashed the commits.

titusfortner:
Nice work!

PR #777, opened 2018-09-13
fix #776 by building XPath using local-name when tag name is specified

### LINKED ISSUE #776: Can't locate an element by class with Watir 6.12
## Meta -
Watir Version: 6.12
Selenium Version: 3.14.0
Browser Version: Chrome 69
Browser Driver Version: 2.35
OS Version: OSX 10.13 (High Sierra)

## Expected Behavior - Following script should work with Watir 6.12 as it works fine in Watir 6.10.1
```
require 'watir'
browser = Watir::Browser.new :chrome
browser.goto 'https://advisors.massmutual.com/'
fail 'Missing state plot on map' unless browser.path(class: 'massachusetts').present?
browser.close 
```

## Actual Behavior - Above script fails when I am using class to locate the element after upgrading to Watir 6.12 but if use css with same class name it works like shown below.
```
require 'watir'
browser = Watir::Browser.new :chrome
browser.goto 'https://advisors.massmutual.com/'
fail 'Missing state plot on map' unless browser.path(css: '.massachusetts').present?
browser.close
```
## Steps to reproduce -
Above script should be enough to reproduce the issue.

titusfortner:
Let's just use local-name for everything.

@twalpole thanks for pointing to the root cause of this!

### DIFF
--- a/lib/watir/locators/element/selector_builder/xpath.rb
+++ b/lib/watir/locators/element/selector_builder/xpath.rb
@@ -9,9 +9,10 @@ module Watir
 
           def build(selectors)
             adjacent = selectors.delete :adjacent
-            xpath = adjacent ? process_adjacent(adjacent) : './/'
+            xpath = adjacent ? process_adjacent(adjacent) : './/*'
 
-            xpath << (selectors.delete(:tag_name) || '*').to_s
+            tag_name = selectors.delete(:tag_name).to_s
+            xpath << "[local-name()='#{tag_name}']" unless tag_name.empty?
 
             index = selectors.delete(:index)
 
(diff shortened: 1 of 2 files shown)


twalpole:
The downside to using local-name is if multiple namespaces are used with the same tag names it doesn't differentiate -- probably not a real issue since one can just write specific XPaths  if necessary -- but just wanted to point it out.

titusfortner:
@twalpole can you give an example of something I would be trying to locate that would have a problem with this?

twalpole:
@titusfortner It would only be if someone used a custom XML namespace that contained element names that collided with existing element names (whole purpose of XML namespaces in the first place) -- it would be a very rare case in a web page though (more common in general XML documents)

titusfortner:
`<custom><div>Foo</div></custom><div>Bar</div>` Could match Bar instead of Foo?

Regardless, it fixes a known problem, I guess I'll wait for someone else to have an issue with it to address that one. :)

twalpole:
Correct - something like

    <blah xmlns="..."><div>abc</div></blah><div>def</div>

would now match both `div` elements - whereas previously it would only match the second one -  really should not be common in web pages though.

PR #778, opened 2018-09-14
update Exceptions and Watir namespacing

titusfortner:
* Removes unnecessary namepsacing; including modules being referenced 
* Creates `LocatorException` for exceptions related to locating an Element
* Creates `WatirException` as a general exception instead of having to use `Error::Exception`
* Removes unused exceptions: `UnknownCellException` & `UnknownRowException`

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -1,13 +1,11 @@
 module Watir
   class Select < HTMLElement
-    include Watir::Exception
-
     #
     # Clears all selected options.
     #
 
     def clear
-      raise Error, 'you can only clear multi-selects' unless multiple?
+      raise Exception::Error, 'you can only clear multi-selects' unless multiple?
 
       selected_options.each(&:click)
     end
@@ -189,7 +187,7 @@ module Watir
         next unless what.is_a?(String) ? value == what : value =~ what
         return true if opt.enabled?
 
-        raise Watir::Exception::ObjectDisabledException, "option matching #{what} by #{how} on #{inspect} is disabled"
+        raise ObjectDisabledException, "option matching #{what} by #{how} on #{inspect} is disabled"
       end
       false
     end
@@ -236,7 +234,7 @@ module Watir
       when :value
         element.value =~ exp
       else
-        raise Error, "unknown how: #{how.inspect}"
+        raise Exception::Error, "unknown how: #{how.inspect}"
       end
     end
   end # Select
--- a/lib/watir/elements/table.rb
+++ b/lib/watir/elements/table.rb
@@ -27,7 +27,7 @@ module Watir
 
     def hashes
       all_rows = rows.locate
-      header_row = all_rows.first || raise(Exception::Error, 'no rows in table')
+      header_row = all_rows.first || raise(Error, 'no rows in table')
 
       all_rows.entries[1..-1].map do |row|
         cell_size_check(header_row, row)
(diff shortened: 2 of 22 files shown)


lib/watir/exception.rb
```diff
@@ -3,13 +3,13 @@ module Exception
     class Error < StandardError; end
 
     # TODO: rename Object -> Element?
+    class WatirException < Error; end
```
p0deje: I don't see why `Watir::Exception::WatirException` is better than `Watir::Exception::Error`.

lib/watir/elements/table.rb
```diff
@@ -69,7 +69,7 @@ def cell_size_check(header_row, cell_row)
 
       index = cell_row.selector[:index]
       row_id = index ? "row at index #{index - 1}" : 'designated row'
-      raise Exception::Error, "#{row_id} has #{row_size} cells, while header row has #{header_size}"
+      raise WatirException, "#{row_id} has #{row_size} cells, while header row has #{header_size}"
```
p0deje: The people can be rescuing those, what's the need for the change?

p0deje (changes requested):

lib/watir/elements/table.rb
titusfortner: yeah, you are right. I thought it read a little better in the code, but when it gets thrown, the whole namespacing will be there, so not really an improvement. Changed it back.

lib/watir/elements/select.rb
```diff
@@ -1,13 +1,11 @@
...
   class Select < HTMLElement
-    include Watir::Exception
-
     #
     # Clears all selected options.
     #
 
     def clear
-      raise Error, 'you can only clear multi-selects' unless multiple?
+      raise WatirException, 'you can only clear multi-selects' unless multiple?
```
jkotests: Missed reverting `WatirException` back to `Error`.

lib/watir/elements/select.rb
```diff
@@ -236,7 +234,7 @@ def matches_regexp?(how, element, exp)
       when :value
         element.value =~ exp
       else
-        raise Error, "unknown how: #{how.inspect}"
+        raise WatirException, "unknown how: #{how.inspect}"
```
jkotests: Missed reverting `WatirException` back to `Error`.

spec/watirspec/user_editable_spec.rb
```diff
@@ -188,14 +188,14 @@
       element = browser.text_field(id: 'new_user_password')
       allow(element).to receive(:value).and_return('wrong')
       msg = "#set! value: 'wrong' does not match expected input: 'secret'"
-      expect { element.set!('secret') }.to raise_exception(Watir::Exception::Error, msg)
+      expect { element.set!('secret') }.to raise_exception(Watir::Exception::WatirException, msg)
```
jkotests: Missed reverting `WatirException` back to `Error`.

spec/watirspec/user_editable_spec.rb
```diff
@@ -188,14 +188,14 @@
...
     end
 
     it "raises Exception if the text of content editable element doesn't match" do
       element = browser.div(id: 'contenteditable')
       allow(element).to receive(:text).and_return('wrong')
       msg = "#set! text: 'wrong' does not match expected input: 'secret'"
-      expect { element.set!('secret') }.to raise_exception(Watir::Exception::Error, msg)
+      expect { element.set!('secret') }.to raise_exception(Watir::Exception::WatirException, msg)
```
jkotests: Missed reverting `WatirException` back to `Error`.

jkotests (changes requested):

PR #779, opened 2018-09-14
deprecate #wait_while_present and #wait_until_present

titusfortner:
Right now we have slightly different behavior between `wait_until(&:present?)` and `wait_until_present` because I didn't think we could do in the former what is necessary for special cases requiring the latter. Because when we use this in the automatic waiting code we always do a check on the cached element before getting into the wait loop, this shouldn't affect performance or functionality. It just means that people using `wait_while_present` and `wait_until_present` can use the to_proc methods again, and we can deprecate these methods.

Also, nesting blocks in procs is very Inception and I like it a lot. :)

### DIFF
--- a/spec/watirspec/wait_spec.rb
+++ b/spec/watirspec/wait_spec.rb
@@ -107,48 +107,63 @@ describe Watir::Element do
   describe '#wait_until_present' do
     it 'waits until the element appears' do
       browser.a(id: 'show_bar').click
-      expect { browser.div(id: 'bar').wait_until_present(timeout: 5) }.to_not raise_exception
+      expect {
+        expect { browser.div(id: 'bar').wait_until_present(timeout: 5) }.to_not raise_exception
+      }.to have_deprecated_wait_until_present
     end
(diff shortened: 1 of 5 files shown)


p0deje (approved):

spec/watirspec/wait_spec.rb
```diff
@@ -303,6 +322,34 @@ def present?
         expect { element.wait_while(custom: '') }.to_not raise_exception
       end
 
+      it 'accepts keywords and block' do
+        element = browser.div(id: 'foo')
+        browser.a(id: 'hide_foo').click
+        expect { element.wait_while(custom: '', &:present?) }.to_not raise_exception
```
kapoorlakshya, line 328: `wait_while(custom: '', &:present?)` will wait while `custom` is blank AND while the element is present. Correct?

titusfortner: Yes, exactly

jkotests (approved):

PR #780, opened 2018-09-15
move predicate regexp conversion into xpath classes

titusfortner:
This PR was prompted by my seeing a bunch of code that only applies to XPath built selectors in the Locator & Selector Builder classes directly. This moves all XPath building code into the SelectorBuilder XPath classes.

In the process I realized we have a number of overloaded / confusing terms.

Locators are the ways elements can be located. One or more can be combined into a selector. For a given element, there can be multiple locators, but only one selector.

We overloaded the term `filter`. This code only uses it to mean either `first` or `all`. So `filter_elements` is now `matching_elements` and `filter_selector` is now: `values_to_match`, etc.


lib/watir/locators/cell/selector_builder/xpath.rb
```diff
@@ -0,0 +1,27 @@
+module Watir
+  module Locators
+    class Cell
+      class SelectorBuilder < Element::SelectorBuilder
```
p0deje: It doesn't look great that you need to extend the class all the time.

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -3,32 +3,67 @@ module Locators
...
-            @should_use_label_element = should_use_label_element
-          end
-
-          def build(selectors)
-            adjacent = selectors.delete :adjacent
-            xpath = adjacent ? process_adjacent(adjacent) : './/*'
+          # Regular expressions that can be reliably converted to xpath `contains`
+          # expressions in order to optimize the locator.
+          CONVERTABLE_REGEXP = /
+          \A
```
p0deje: This and below should have +1 indentation level.

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -3,32 +3,67 @@ module Locators
...
 
-            p xpath: xpath, selectors: selectors if $DEBUG
+            xpath = add_regexp_predicates(xpath, values_to_match)
+
+            Watir.logger.debug(xpath: xpath, selector: selector)
 
             [:xpath, xpath]
           end
 
+          def default_start
```
p0deje, line 38: Can it be a constant?

p0deje (changes requested):

lib/watir/locators/cell/selector_builder/xpath.rb
titusfortner: the alternative is to keep an empty selector_builder.rb file with just the subclass defined... well, I guess that might be better. 👍

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: the point of the method is that this value is different depending on the subclass, which constants are not good for.

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: man, I was hoping rubocop would automatically fix this for me when I ran it. :) Copied over the precise way it was in the other

p0deje (approved):

jkotests (approved):

PR #781, opened 2018-09-15
improve coverage by adding specs and deleting vestigial and impossibl…

titusfortner:
I'm starting to go through all of the "uncovered" lines of code. Some of this code isn't needed, some of it needs specs added. I'm going to try not to wait until I've overhauled everything to get things merged in, but would appreciate another set of eyes on it.

### DIFF
--- a/lib/watir/cookies.rb
+++ b/lib/watir/cookies.rb
@@ -18,7 +18,7 @@ module Watir
 
     def to_a
       @control.all_cookies.map do |e|
-        e.merge(expires: e[:expires] ? to_time(e[:expires]) : nil)
+        e.merge(expires: e[:expires] ? e[:expires].to_time : nil)
       end
     end
 
@@ -117,16 +117,6 @@ module Watir
     def load(file = '.cookies')
       YAML.safe_load(IO.read(file), [::Symbol]).each do |c|
         add(c.delete(:name), c.delete(:value), c)
-      end
-    end
-
-    private
-
-    def to_time(time)
-      if time.respond_to?(:to_time)
-        time.to_time
-      else
-        ::Time.local time.year, time.month, time.day, time.hour, time.min, time.sec
       end
     end
   end # Cookies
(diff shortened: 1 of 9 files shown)


lib/watir/cookies.rb
```diff
@@ -119,15 +119,5 @@ def load(file = '.cookies')
         add(c.delete(:name), c.delete(:value), c)
       end
     end
-
-    private
-
-    def to_time(time)
-      if time.respond_to?(:to_time)
```
p0deje, line 126: Do you know why this was needed?

lib/watir/locators/element/selector_builder.rb
```diff
@@ -54,9 +54,9 @@ def build(selector)
           inspect = selector.inspect
           return given_xpath_or_css(selector) if selector.key?(:xpath) || selector.key?(:css)
 
-          built = build_wd_selector(selector)
-          Watir.logger.debug "Converted #{inspect} to #{built}"
-          built
+          build_wd_selector(selector).tap do |built|
```
p0deje: 🤷‍♂️

p0deje (approved):

lib/watir/cookies.rb
titusfortner: I spent a solid half hour trying to figure out what class instance would respond to year / month methods and not to_time. Then I realized that this is only acting on information coming back from selenium, so we know that what the object will be nil or date time

jkotests (approved):

lib/watir/cookies.rb
p0deje: It is possible this exists for backwards compatibility with older Ruby, but whatever.

titusfortner: I literally couldn't find a class for any version of Ruby where it responded to `#year` but not `#to_time`. Maybe Selenium returned a custom something at one point?

lib/watir/locators/element/selector_builder.rb
titusfortner: I know, this goes against my preference of not unnecessarily using blocks.

lib/watir/cookies.rb
p0deje: 🤷‍♂️

PR #782, opened 2018-09-16
fix generator code and skip unreachable title and html methods

titusfortner:
I broke the generation code in my refactor.
Does the change I made to remove #html & #title make sense?

You can now do `browser.head.title` to get the `Title` instance, and `browser.element` will provide an `Html` instance.

I'm not sure why piping `less` causes rake to hang, and I can't get SVG to finish, hopefully this isn't something I broke.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -41,7 +41,7 @@ end
 
       File.open(spec_path, 'w') do |io|
         io << "<!--  downloaded from #{spec_uri} on #{Time.now} -->\n"
-        io << data = URI.open(spec_uri).read
+        io << data = URI.parse(spec_uri).read
         downloaded_bytes = data.bytesize
       end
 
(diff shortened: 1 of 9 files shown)


p0deje (changes requested):
What's reason for those changes, I can't figure out from code? Do we really need to skip Html and Title?

titusfortner:
We're not skipping the classes, just the methods in container because they are inaccessible to the user.

The method `#title` in `html_elements.rb` can never be used because Browser class will return `@driver.title` and any instance of `Watir::HTMLElement` will return the attribute defined at the top of the page. I thought it would make sense to allow it to be accessed via `Head` object (`browser.head.title`). I mean, the alternative is to skip the title attribute. This seemed a better approach.

Similarly `#html` in `html_elements.rb` can never be used because Browser class will return `@driver.page_source` and no other element can have a nested `Html` element.

p0deje:
Ah, I see what you mean. Still, those methods can be used indirectly :) Watirsome for example iterates over all `Watir::Container` methods to understand what elements Watir supports.

titusfortner:
Agreed, which is why I manually added them to `@tag_to_class`: https://github.com/watir/watir/pull/782/files#diff-198926258bcae3fde9d5ba349da2e206R153

Again, not sure this is the best way to do it, probably no harm in keeping them there as is, just felt weird to have methods that can't be used.

p0deje:
Let's keep them. Watir::Container is just a module so can be included in other places.

PR #783, opened 2018-09-18
More specs and removal of inaccessible code

titusfortner:
A couple interesting things in here...

`Element#value` is useless because the methods for which `attribute_value('value')` makes sense, there is already an attribute 'value' defined on the element. So this method is never actually used in our specs, and were it to ever be used, it would have to return an empty string.

`Option#text` is fascinating. The `getAttribute` atom that Selenium uses returns the text content when calling `attribute_value('label')` or `attribute_value('text')`. It can't be overridden in the DOM (I tried adding labels and text attributes without any change), so there is no point in overriding the superclass. Also better to use `#text` directly than call in an atom.

We're currently doing a selector type check in both Locator class and SelectorBuilder class. Seemed to make more sense to do it in SelectorBuilder, so I removed the check from Locator class.

Plus a bunch more specs to hit places in the code that are currently ignored.

### DIFF
--- a/lib/watir/adjacent.rb
+++ b/lib/watir/adjacent.rb
@@ -59,7 +59,7 @@ module Watir
     #
     # @example
     #   browser.text_field(name: "new_user_first_name").following_siblings.size
-    #   #=> 53
+    #   #=> 55
     #
 
     def following_siblings(opt = {})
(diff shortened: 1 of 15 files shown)


p0deje (approved):

jkotests (approved):

titusfortner:
fyi, I was wrong. HTML works for a label attribute in the option tag, without any text between in the element tag. That being said, just doing the label attribute should provide everything needed here. Updated code.

PR #788, opened 2018-09-19
automatically retry examples for specific exceptions typical on Travis

titusfortner:
I'm getting tired of hitting the re-run button on Travis when we have a IOError or Net::ReadTimeout. I'm not certain this actually fixes the problem, though, since each "it" statement isn't starting a new session; a dead session might be a dead session.


p0deje:
🤷‍♂️

Issue #786, opened 2018-09-18
Provide a way to search for text using XPath predicate


abotalov:
## Expected Behavior -
There should be an option (maybe `browser.element(class: x, all_text: y)`) that will be converted to XPath.

## Actual Behavior -
When `browser.element(class: x, text: y)` is called, Watir iterates through all elements with class `x` to find an element with text `y`. This is slow if there are many elements with class `x`. That's why I currently use xpath directly on some cases.

jkotests:
@abotalov , is `y` a String or a Regexp?

We should already be doing this for Strings:

~~~~~~~~
browser.element(class: 'a', text: 'y').exists?
#=>2018-09-18 16:44:51 DEBUG Watir Converted {:class=>"a", :text=>"y"} to 
  [:xpath, ".//*[contains(concat(' ', @class, ' '), ' a ') and normalize-space()='y']"]
~~~~~~~~

In the case of Regexps, we don't do anything. We could presumably make similar optimizations to what we did for attributes - ie trying to put part of the regexp into the XPath.

titusfortner:
The specific behavior requested here has been implemented. I believe the intended request is to allow users to leverage the Regexp optimizations we are currently using for attributes. This will require us to remove support for text_regexp deprecation in Watir 7, and will be a straightforward add with the new code.

PR #793, opened 2018-09-26
Redo selector builder

titusfortner:
The goal here is to be able to better leverage regular expressions to make more single `#find_element` calls instead of `#find_elements` calls with follow-on wire calls to match from the returned collection. This addresses #447 and #786 and was discussed in #331. 

To do this, we needed to be able to interact with all of the Regexp values in the Selector Builder implementation itself. Per something @jkotests brought up a while back about the concerns of our location classes not being well separated, I tried to identify what classes should be responsible for what things.

I determined that the Locator class should be responsible for anything with a wire call, including matching values that weren't able to be added to the selenium locator returned by the SelectorBuilder.

SelectorBuilder class should be responsible for normalizing the selector / locators and delegating actual building of the selenium locator to the implementation class, then returning the crafted locator independently of whatever pieces of the Watir selector were not used in building that locator.

I created a `selector_builder_spec.rb` unit test file to show what the conversions should look like. I found a few minor bugs in our current implementation, and there are a ton more I want to add here before actually merging this, but I'd love some eyes on this approach before I keep digging through it.

This refactor made it easier to do [this](https://github.com/watir/watir/commit/4d9699683a7db4c8922fc394201fd4171b9e7240) so as not have to match the regex attribute values that are entirely handled by the XPath `contains` parameter. [You can see the difference here](https://gist.github.com/titusfortner/9e75a497dd87f3e1a303a1975462263e) in how many wire calls are shifted from `find_elements` to `find_element`. 

This code "works" but can definitely be tidied up a lot more. It's amazing how complicated all of the corner cases makes this. :)

This sets us up to be able to use `contains` and `starts-with` for text & attributes. You can see @p0deje's ideas for this in [this comment](https://github.com/watir/watir/pull/331#issuecomment-89555559). We could even use XPath index if everything else in the selector is handled in the XPath building. It would be great if we only needed to do extra matching for things related to displayedness and especially complicated regular expressions.

Note that the `SelectorBuilder` class here is initialized differently, so this will break whatever projects are leveraging Watir's use of it. We may want to wait on this (and the follow-on implementations using this) to release with Watir 7 due to how it will interact with some of the things we're deprecating.

### DIFF
--- a/lib/watir/locators/cell/selector_builder/xpath.rb
+++ b/lib/watir/locators/cell/selector_builder/xpath.rb
@@ -3,7 +3,7 @@ module Watir
     class Cell
       class SelectorBuilder
         class XPath < Element::SelectorBuilder::XPath
-          def add_attributes(selector)
+          def add_attributes(selector, _scope_tag_name)
             attr_expr = attribute_expression(nil, selector)
 
             expressions = %w[./th ./td]
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -75,7 +75,7 @@ Metrics/CyclomaticComplexity:
 Metrics/AbcSize:
   Max: 21
   Exclude:
-    - 'lib/watir/locators/element/selector_builder.rb'
+    - 'lib/watir/locators/element/selector_builder/xpath.rb'
     - 'lib/watir/locators/element/locator.rb'
     - 'lib/watir/generator/base/generator.rb'
 
--- a/lib/watir/locators.rb
+++ b/lib/watir/locators.rb
@@ -60,7 +60,8 @@ module Watir
 
       def build_locator
         element_validator = element_validator_class.new
-        selector_builder = selector_builder_class.new(@query_scope, @selector.dup, element_class.attribute_list)
+        scope_tag_name = @query_scope.selector[:tag_name] if @query_scope.respond_to?(:selector)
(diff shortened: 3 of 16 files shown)


p0deje (approved):
The refactoring looks good to me, though I didn't dig into details. +1 for unit tests, those are definitely valuable.

jkotests:
I'll have to dig into the details, but this looks great.

Can we de-couple Watir 7 from Selenium 4? It'd be nice to get these performance improvements and build on the idea without having to wait.

lib/watir/locators/cell/selector_builder/xpath.rb
```diff
@@ -3,7 +3,7 @@ module Locators
     class Cell
       class SelectorBuilder
         class XPath < Element::SelectorBuilder::XPath
-          def add_attributes(selector)
+          def add_attributes(selector, _scope_tag_name)
```
kapoorlakshya, line 6: I'm curious what purpose `_scope_tag_name` serves here and on line 11 in `button/selector_builder/xpath.rb` if it is not being used in the methods. I'm assuming the `_` before the parameter name implies that it won't be referenced.

titusfortner: It's necessarily for one of the subclasses. We used to pass the entire query scope which isn't necessary and we want all wire calls to be in locator class

titusfortner:
Closing in favor of #796

PR #795, opened 2018-10-01
Locate visible link text via wd link locators

jkotests:
When locating links via :visible_text, if possible, we will convert it to Webdrivers's link_text or partial_link_text. This should improve performance for those with a lot of links on the page.

Note:
* I only implemented this for #link and #links. This will not get applied when using #element and #elements.

### DIFF
--- a/lib/watir/locators.rb
+++ b/lib/watir/locators.rb
@@ -3,6 +3,8 @@ require 'watir/locators/element/selector_builder'
 require 'watir/locators/element/selector_builder/xpath'
 require 'watir/locators/element/validator'
 
+require 'watir/locators/anchor/selector_builder'
+
 require 'watir/locators/button/locator'
 require 'watir/locators/button/selector_builder'
 require 'watir/locators/button/selector_builder/xpath'
--- a/lib/watir/locators/anchor/selector_builder.rb
+++ b/lib/watir/locators/anchor/selector_builder.rb
@@ -1,0 +1,37 @@
+module Watir
+  module Locators
+    class Anchor
+      class SelectorBuilder < Element::SelectorBuilder
+        private
+
+        def build_wd_selector(selector)
+          build_link_text(selector) || build_partial_link_text(selector) || super
+        end
+
+        def build_link_text(selector)
+          return unless can_convert_to_link_text?(selector)
+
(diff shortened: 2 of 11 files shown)


titusfortner:
So, I've spent a several dozen hours refactoring locators this past week. I'm ok with adding this for now, especially if we decide to hold off on my refactor #796. Unlike #793 this one is functionally backward compatible, so I don't think there is a reason to wait for Watir 7 to merge it in.

titusfortner (approved):

jkotests:
I'll have to make an update to skip Regexp with casefold. Will try to update tonight.

jkotests:
@titusfortner , when do you think we'll release #796? I mostly did this out of curiosity. If #796 is going to go out soonish, I'm fine to throw this out and redo it then. Especially given the simplicity of the change, it's easy for someone to take this and monkey-patch for the short-term.

titusfortner:
I'm on the fence about releasing #796 as is.
I've tidied it up so I think it is now mergeable, but I'm reluctant because it is a backward step as far as performance right now.

Let's... merge this PR (hacktoberfest), then you can create a new PR rebased off of #796 with the 
updated implementation.

Before releasing a version of Watir with #796, I'd like to see:
1. PR for direct support of link_text (and partial link text where possible)
2. PR for Simple Regex supported for Text
3. PR for direct index support where possible

Ideally we figure out how to work with more advanced Regexp as well. Capybara just implemented this: https://github.com/teamcapybara/capybara/blob/xpath_regexp/lib/capybara/selector/regexp_disassembler.rb#L12 which I believe is doing what we were trying to do with this: https://github.com/watir/watir/pull/796/files#diff-d08ef5e8146220999009367844ed48d6L10

jkotests:
Hacktoberfest only requires you to make a pull request... it doesn't have to be accepted. Just cannot be marked as spam.

titusfortner:
I merged #796 into master to make this easier to update

jkotests:
@titusfortner , this has been updated for the re-factored selector builder. I'm not quite sure why the code coverage is decreased.

titusfortner:
Weird weird on coverage thing. kept trying to get the numbers to add up and they just don't. :)

PR #796, opened 2018-10-01
Redo selector builder2

titusfortner:
Ok, this is an overhaul/replacement of #793. It does *not* include any additional functionality. In fact, it removes a performance tweak that has been shown to not be entirely accurate. It does manage the Separation of Concerns from the previous PR, and will make it very easy to improve text matching, regular expression matching, and indexes with XPath  to minimize unnecessary extra matching calls.

It does add a ton of specs for proper building of element Selectors. It then does an actual lookup using the XPath defined to ensure that it is finding the expected element (I added `data-locator` attributes to specify what I expected when writing the specs.

For instance, `id: /foob?/` should match `<div id='foo'>`, but our current implementation does not. Rather than fixing it in this PR, I removed it because it is only a performance tweak, and this refactor should allow for a better, more comprehensive solution.

I'd like to deprecate two things with this PR:
1. Using `:caption` locator to find text. This currently works for all elements, and I don't see why it is useful on any
2. Using `'text'` locator to find text, rather than `:text`

I'd also kind of like to prevent using symbols for tag_names, since that's kind of weird, but I didn't specifically add that this time.

Another issue is I can't figure out how to mock out the StaleElement code. The existing spec only works because the current code is making a completely unnecessary matching call after finding the right element with XPath. This is what caused the coverage drop. I should probably think more about it before merging.

Note that this does not pass rubocop as is. I split out all of the conditionals onto separate lines to make sure code coverage is 100% and no extraneous branches exist. Both selector_builder specs and our elements specs independently exercise 100% of this code, which is what I'm going for. I propose we keep the structure this way for any spec removal we do before merging this into master.

### DIFF
--- a/lib/watir/locators/element/selector_builder/xpath.rb
+++ b/lib/watir/locators/element/selector_builder/xpath.rb
@@ -5,111 +5,116 @@ module Watir
         class XPath
           include Exception
 
-          # Regular expressions that can be reliably converted to xpath `contains`
-          # expressions in order to optimize the locator.
-          CONVERTABLE_REGEXP = /
-            \A
-              ([^\[\]\\^$.|?*+()]*) # leading literal characters
-              [^|]*?                # do not try to convert expressions with alternates
-              (?<!\\)               # skip metacharacters - ie has preceding slash
-              ([^\[\]\\^$.|?*+()]*) # trailing literal characters
-            \z
-          /x
+          CAN_NOT_BUILD = %i[visible visible_text].freeze
+          LITERAL_REGEXP = /\A([^\[\]\\^$.|?*+()]*)\z/
 
-          def initialize(use_label_element)
-            @should_use_label_element = use_label_element
+          def build(selector)
+            @selector = selector
+
+            @requires_matches = (@selector.keys & CAN_NOT_BUILD).each_with_object({}) do |key, hash|
+              hash[key] = @selector.delete(key)
+            end
+
+            index = @selector.delete(:index)
+
+            start_string = default_start
+            adjacent_string = add_adjacent
+            tag_string = add_tag_name
+            class_string = add_class_predicates
+            attribute_string = add_attribute_predicates
+            converted_attribute_string = convert_attribute_predicates
+            text_string = add_text
+            index_string = if !adjacent_string.empty? && !index.nil? && !index.negative?
+                             "[#{index + 1}]"
+                           else
+                             @requires_matches[:index] = index if index
+                             ''
+                           end
+
+            xpath = "#{start_string}#{adjacent_string}#{tag_string}#{class_string}#{attribute_string}" \
+"#{converted_attribute_string}#{text_string}#{index_string}"
+
+            # TODO: Ideally everything gets put into @requires_matches
+            # class Array values make this difficult
+            @selector.merge!(@requires_matches)
+
+            {xpath: xpath}
           end
 
-          def build(selector, values_to_match)
-            adjacent = selector.delete :adjacent
-            xpath = adjacent.nil? ? default_start : process_adjacent(adjacent)
-            xpath << add_tag_name(selector)
-            index = selector.delete(:index)
+          protected
 
-            # the remaining entries should be attributes
-            xpath << add_attributes(selector)
+          def add_text
+            return '' unless @selector.key?(:text)
 
-            xpath << "[#{index + 1}]" if adjacent && index
-
-            xpath = add_regexp_predicates(xpath, values_to_match)
-
-            Watir.logger.debug(xpath: xpath, selector: selector)
-
-            [:xpath, xpath]
+            text = @selector.delete :text
+            if !text.is_a?(Regexp)
+              "[normalize-space()=#{XpathSupport.escape text}]"
+            else
+              @requires_matches[:text] = text
+              ''
+            end
           end
 
           def default_start
-            './/*'
-          end
-
-          def add_tag_name(selector)
-            tag_name = selector.delete(:tag_name).to_s
-            tag_name.empty? ? '' : "[local-name()='#{tag_name}']"
-          end
-
-          def add_attributes(selector)
-            element_attr_exp = attribute_expression(nil, selector)
-            element_attr_exp.empty? ? '' : "[#{element_attr_exp}]"
-          end
-
-          def add_regexp_predicates(what, selector)
-            return what unless convert_regexp_to_contains?
-
-            selector.each do |key, value|
-              next if %i[tag_name text visible_text visible index].include?(key)
-
-              predicates = regexp_selector_to_predicates(key, value)
-              what = "(#{what})[#{predicates.join(' and ')}]" unless predicates.empty?
+            if @selector.key?(:adjacent)
+              './'
+            else
+              './/*'
             end
-            what
           end
 
-          # @todo Get rid of building
-          def attribute_expression(building, selector)
-            f = selector.map do |key, val|
-              if val.is_a?(Array) && key == :class
-                '(' + val.map { |v| build_class_match(v) }.join(' and ') + ')'
-              elsif val.is_a?(Array)
-                '(' + val.map { |v| equal_pair(building, key, v) }.join(' or ') + ')'
-              elsif val.eql? true
-                attribute_presence(key)
-              elsif val.eql? false
-                attribute_absence(key)
-              else
-                equal_pair(building, key, val)
-              end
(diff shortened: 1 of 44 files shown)


titusfortner:
Ok, let me break down this implementation a little better... (You know when you have to explain something to someone it shows you how you can make things more clear? I added another commit that should make this more clear.)

Big Picture:

Locator classes:
Take shortcut by locating elements with Selenium where possible
Sends all the locators to the Selector Builder
The Selector Builder creates a Selenium locator (XPath by default) and deletes all of the keys from `@selector` that have been completely addressed.
If there are remaining keys in the selector, the Locator class does a `#find_all` with the Selenium locator, then iterates through to find element(s) that match all of those remaining keys

SelectorBuilder classes
Normalizes locator & checks types
Then delegates the actual building of the Selenium locator (XPath by default)

XPath Builder classes
All of the heavy lifting is here in the `#build` method.

For XPath there is no difference between `[@id and @value]` and `[@id][@value]`
So I've split up the locator into 8 sections, each of which will be put into brackets. Multiple parameters within those sections will be separated by `and`.

The created XPath looks like this:
```
"#{start_string}#{adjacent_string}#{tag_string}#{class_string}#{attribute_string}" \
"#{converted_attribute_string}#{text_string}#{index_string}"
```

Each string is created by a class that returns `''` if not applicable, deletes the applicable key from `@selector` if it has processed it, and returns the composite attribute inside of brackets if applicable.

The XPath Builder subclasses have a different `#build` method to process the things that are unique to that Element type. Usually the subclasses pull out the items special to the Element, make a call on the superclass, figure out what the common attributes are and pipe (` | `) those amongst multiple tag types, etc.

More clear?

titusfortner:
I should add that I split things up this way, so I could test the logic within each bracket separately, then test that they can be added together. Theoretically this will make it more straightforward to see what is going on.

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -5,111 +5,113 @@ class SelectorBuilder
...
+              './/*'
+            end
+          end
 
-              predicates = regexp_selector_to_predicates(key, value)
-              what = "(#{what})[#{predicates.join(' and ')}]" unless predicates.empty?
+          def add_tag_name
+            tag_name = @selector.delete(:tag_name)
+            if tag_name.is_a?(Regexp)
+              "[contains(local-name(), '#{tag_name.source}')]"
```
jkotests, line 78: This only works if the source is only literal characters. For example, the following literally searches for "a|b" instead of "a" or "b":

~~~~~~~~
browser.elements(tag_name: /a|b/).count
#=> Watir Converted {:tag_name=>/a|b/} to {:xpath=>".//*[contains(local-name(), 'a|b')]"}, with {} to match
~~~~~~~~

Though we didn't actually support locating :tag_name with Regexp in 6.14:


~~~~~~~~
browser.elements(tag_name: /a|b/).count
#=> Watir Converted {} to [:xpath, ".//*"]
#=> Watir Iterated through 3 elements to locate all {}
#=> NoMethodError: undefined method `to_sym' for /a|b/:Regexp
~~~~~~~~

Ideally, complex regexps would get passed back for matching. However, since we didn't have support before, simply raising exception also seems reasonable.

spec/watirspec/selector_builder/cell_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      it 'attribute and text' do
+        @selector = {headers: /before_tax/, text: '5 934'}
+        @wd_locator = {xpath: "./*[local-name()='th' or local-name()='td']" \
+"[contains(@headers, 'before_tax')][normalize-space()='5 934']"}
+        @data_locator = 'before tax'
+      end
+    end
+
+    it 'delegates adjacent to Element SelectorBuilder' do
```
jkotests, line 49: Isn't this saying:

~~~~~~~~
browser.table.tr.cell(adjacent: :ancestor, index: 2)
#=> Watir::Table
~~~~~~~~

Given we are call `#cell`, wouldn't one expect to find a `th` or `td`? "cell" is supposed to be the immediate child of a given row, so does that make it weird to be going even further up? I haven't used :adjacent directly, so maybe I'm thinking of this wrong?

jkotests (approved):
Looks great. Definitely seems like it will be easier to add stuff.

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: I was thinking about this from the standpoint of being able to match any header element. I could verify it is simple regexp here... But I can't think of another valid use case, so maybe not worth the bother of supporting.

spec/watirspec/selector_builder/cell_spec.rb
titusfortner: No, you're right. I actually spent several hours working with adjacent specs and implementation ideas, until I had the "I'm an idiot moment" and realized what I was doing didn't make sense. All this spec is doing is code coverage to demonstrate that return super line is hit in the cell XPath class for adjacent tags.

lib/watir/locators/element/selector_builder/xpath.rb
jkotests: The any header element is a nice idea. Since you already got it working for simple regexps, I'd just do the quick thing and throw an error if a complex regexp. Presumably you could pull the #simple_regexp? method from the Button XPath class back to here..

spec/watirspec/selector_builder/cell_spec.rb
jkotests: I'd be tempted to just throw an exception if someone specified :adjacent. I think it'd allow you to simplify such that you don't have to manually decomponse and recompose the xpath. This re-building of the xpath was quite confusing to me. I would have thought that you could just re-define certain methods and #build would just work.

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: though if that's really the only use case I think I'd prefer a new like `#section_header` to match just h1-h6.
Really the only time I expect users to use `:tag_name` locator directly is for adjacent calls, which can use optimization if necessary....

I guess the best reason to keep it is to be consistent across our locators. `#simple_regexp?` is in the superclass already, so available to use right there.

spec/watirspec/selector_builder/cell_spec.rb
titusfortner: Well, I guess I'm reluctant to throw an exception when it is handled easily enough....
Something like?

```ruby
table.row(id: 'outer_last').child(index: -1)
```
`#row` matches any `tr` under `tbody`, `thead` or `tbody`, though now that I think about it, I'm not sure why we are jumping through those hoops, because `tr` by itself should do that just fine... I think we can investigate more optimization there...

Actually, maybe I should remove all index matching with adjacent locators in this PR, and then add back in a global solution in a follow-up...

titusfortner: Really these special SelectorBuilder classes are only used internally by Table & Row subclasses to support things like `#hashes` and `#headers`. Also, we leverage them to do `attribute_value(:rowIndex)` because the order that is in the DOM is not the order of what is displayed on the page.

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: Also, matching header tags won't work with this implementation. `browser.element(tag_name: /h/)` is what would be supported and that would match a lot of things.

abotalov: `tag_name` can be used for custom tag names.

spec/watirspec/selector_builder/cell_spec.rb
jkotests: `#row` and `#tr` are not the same:
* `#row` matches only `tr` elements in the current table (ignores nested tables)
* `#tr` matches all descendant `tr`, including those in nested tables

jkotests: I'm not sure I understand the intent of the `table.row(id: 'outer_last').child(index: -1)` example. That wouldn't end up calling the cell selector builder?

titusfortner: No, it will make an adjacent call on the Row builder.
and `child` will work with both `th` and `td` elements.

This isn't a great example, but I think it should be a valid one, and it isn't that difficult to support with this code.

PR #799, opened 2018-10-08
Text regexp

titusfortner:
So this one has to remove support for locating visible text with a regular expression 
I'm thinking it should wait until whenever we remove the deprecations in Watir 7?


p0deje (approved):

jkotests (approved):
Agreed that we'll have to wait until Watir 7.

titusfortner:
I'm going to hold off on this one and focus on improving attribute regexp matching. Current implementation doesn't support text matching, so no reason to rush this one through.

PR #800, opened 2018-10-08
Index

titusfortner:
If the index is positive, and there are no visibility requirements, this will use index as part of the XPath.

This also tidies up some of the poor code from the recent huge PR I made.

### DIFF
--- a/lib/watir/locators/element/selector_builder/xpath.rb
+++ b/lib/watir/locators/element/selector_builder/xpath.rb
@@ -23,24 +23,29 @@ module Watir
             attribute_string = add_attribute_predicates
             converted_attribute_string = convert_attribute_predicates
             text_string = add_text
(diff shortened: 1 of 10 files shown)


p0deje (approved):

jkotests (approved):

lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -214,27 +227,27 @@ def attribute_absence(attribute)
...
+              @selector.delete(key)
+              predicate
             end
 
             predicates.compact.empty? ? '' : "[#{predicates.compact.join(' and ')}]"
           end
 
-          # TODO: Ideally we leverage @requires_matches to not return multiple values here
-          # class Array values make this difficult
           def convert_predicate(key, regexp)
```
jkotests, line 232: TextArea::SelectorBuilder::XPath will need to be updated to match the changes here.

PR #802, opened 2018-10-11
Use javascript to get tag_name for ElementCollection elements

lmtierney:
When the `selector` does not contain `:tag_name`, we are currently querying the `tag_name` from the element. This causes `N` wire calls which, depending on how many elements there are, can take a while when testing remotely.

The idea is to us the javascript executor to retrieve all the tags once. In doing so, we do miss out on the added benefit of `element_call` which `#tag_name` uses, but that shouldn't be an issue as we have already located the elements.

The only concern I could think of was with elements going stale between retrieving them and then performing javascript on them, but this should be a rare scenario. This is mitigated by the rescue and retry. Theoretically there is a chance for an infinite loop on the retry, but the likelihood of that is near impossible in normal conditions.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -86,12 +86,11 @@ module Watir
     def to_a
       hash = {}
       @to_a ||=
-        elements.map.with_index do |e, idx|
+        elements_with_tags.map.with_index do |(e, tag_name), idx|
           selector = @selector.merge(element: e)
           selector[:index] = idx
           element = element_class.new(@query_scope, selector)
           if [HTMLElement, Input].include? element.class
-            tag_name = @selector[:tag_name] || element.tag_name
             hash[tag_name] ||= 0
             hash[tag_name] += 1
             selector[:index] = hash[tag_name] - 1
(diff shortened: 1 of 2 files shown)


lib/watir/element_collection.rb
```diff
@@ -162,6 +161,19 @@ def elements
...
 
+    def elements_with_tags
+      els = elements
+      if @selector[:tag_name]
+        els.map { |e| [e, @selector[:tag_name]] }
+      else
+        begin
+          els.zip(browser.wd.execute_script('return arguments[0].map(function(e) {return e.localName});', els))
+        rescue Selenium::WebDriver::Error::ObsoleteElementError
+          elements_with_tags
```
titusfortner, line 172: It's unlikely, but if it happened there's no useful information about what's going on. This is how I addressed it elsewhere: https://github.com/watir/watir/blob/master/lib/watir/locators/element/locator.rb#L200-L212
Whenever I get back around to the Watir Executor, this can be cleaned up.

titusfortner (changes requested):
Yeah, there are a few places in the code that we might be able to remove excessive iterating with some well-applied JS execution.

lib/watir/element_collection.rb
titusfortner: well this link is super unuseful  after the refactor

lmtierney: I'll take a look again after everything stops moving. I know what you're referring to https://github.com/watir/watir/blob/1e1f7f891bfce7aa4acae6bf36490f7994923fa7/lib/watir/locators/element/locator.rb#L39-L51

titusfortner: Ugh, really we want to wrap these in an `#element_call` type method that respects the global timeout. I think this is where the Watir::Executor ideally gets really handy.

PR #804, opened 2018-10-13
Add Element#obscured?

jkotests:
Adds a method to check if an element is obscured - ie covered by another element. This allows users to check whether clicking an element will generate an ElementClickInterceptedError (or similar error based on driver).

This means that retrying clicks due to things like temporary overlays, can be changed from:

~~~~~~~~
begin
  browser.link.click
rescue
  retry
end
~~~~~~~~

To:

~~~~~~~~
browser.link.wait_while(&:obscured?).click
~~~~~~~~

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -489,6 +489,21 @@ module Watir
     end
 
     #
+    # Returns true if the element's center point is covered by a non-descendant element.
+    #
+    # @return [Boolean]
+    #
+
+    def obscured?
+      element_call do
+        return true unless present?
+
+        scroll_into_view
+        execute_js(:elementObscured, self)
+      end
+    end
+
+    #
     # Returns given style property of this element.
     #
     # @example
--- a/lib/watir/js_snippets/elementObscured.js
+++ b/lib/watir/js_snippets/elementObscured.js
@@ -1,0 +1,14 @@
+// Original Author: Florent B.
+// Source: https://stackoverflow.com/a/45244889/1200545
+function() {
+    var elem = arguments[0],
+        box = elem.getBoundingClientRect(),
+        cx = box.left + box.width / 2,
+        cy = box.top + box.height / 2,
+        e = document.elementFromPoint(cx, cy);
+    for (; e; e = e.parentElement) {
+        if (e === elem)
+            return false;
+    }
+    return true;
+}
--- a/spec/watirspec/elements/element_spec.rb
+++ b/spec/watirspec/elements/element_spec.rb
@@ -874,4 +874,59 @@ describe 'Element' do
       expect { browser.div(id: 'no_such_id').classes }.to raise_unknown_object_exception
     end
   end
+
+  describe '#obscured?' do
+    before { browser.goto WatirSpec.url_for('obscured.html') }
+
+    it 'returns false if element\'s center is not covered' do
+      btn = browser.button(id: 'not_obscured')
+      expect(btn).not_to be_obscured
+      expect { btn.click }.not_to raise_exception
+    end
+
+    it 'returns false if element\'s center is covered by its descendant' do
+      btn = browser.button(id: 'has_descendant')
+      expect(btn).not_to be_obscured
+      expect { btn.click }.not_to raise_exception
+    end
+
+    it 'returns true if element\'s center is covered by a non-descendant' do
+      btn = browser.button(id: 'obscured')
+      expect(btn).to be_obscured
+      not_compliant_on :chrome do
+        expect { btn.click }.to raise_exception(Selenium::WebDriver::Error::ElementClickInterceptedError)
+      end
+      compliant_on :chrome do
+        expect { btn.click }.to raise_exception(Selenium::WebDriver::Error::UnknownError)
+      end
+    end
+
+    it 'returns false if element\'s center is surrounded by non-descendants' do
+      btn = browser.button(id: 'surrounded')
+      expect(btn).not_to be_obscured
+      expect { btn.click }.not_to raise_exception
+    end
+
+    it 'scrolls element into view before checking if obscured' do
+      btn = browser.button(id: 'requires_scrolling')
+      expect(btn).not_to be_obscured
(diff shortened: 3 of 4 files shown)


Rajagopalan-M:
That's an amazing idea.

Rajagopalan-M:
But I want this behavior to be for all elements like waiting behavior for
visibility checking, stale element checking has done for all the action.

titusfortner (changes requested):
This is great! Only caveat is that I'd like for us to test it on all the browsers before we merge it since it isn't a wire call and isn't a bang method. Looks like a couple are failing on Firefox on Travis. What have you run it on so far?

titusfortner:
It should be noted that this is likely to be a solid an alternate solution to #532 from #534.

@Rajagopalan-M yeah I would be in favor of it philosophically, but it would prevent users doing wheat @p0deje  is describing here: https://github.com/watir/watir/pull/534#issuecomment-283244188

The right answer is going to be getting the Watir::Executor working (eventually) so each user can pick and choose what gets added by default.

Rajagopalan-M:
Excited to see such a feature. But does that affect performance? Because we
can simply rescue as we are rescuing for other exception like
InvisibleElement, staleElement. What do you say? Firefox is giving standard
Error now ElementInterceptedError, likewise we can raise bug in chrome to
give standard error like Firefox is providing us. What do you say?

titusfortner:
By default Watir clicks the element and rescues exceptions and automatically deals with the results. It does not do that for element not clickable type exceptions.

Using `browser.link.wait_while(&:obscured?).click` will result in an extra `#execute_script` wire call before attempting to click the element. But it can also be used to do something like this (we'd have to find a way to test that this method actually rescues these exceptions to make sure this is true):

```ruby
def click_unobscured(element)
  element.click
rescue ElementNotInteractableError, ElementClickInterceptedError
  element. wait_while(&:obscured?).click
end
```

@jkotests do we know what exceptions this avoids? I'm not sure we've found a way to reliably reproduce either of these exceptions.

Rajagopalan-M:
That's the perfect way to solve!

jkotests:
@titusfortner , I'll look into the test failures. The problem seems to be with #scroll_into_view and hidden elements, but I'm not exactly sure why.

I'm thinking I'll need to check that the element is present. Do you think it's better to have #obscured? wait for the element to be present (ie `element_call(:wait_for_present)`) or just do a check (ie `return true unless present?`)?

I'll have to do more testing to see what exceptions it avoids in Firefox. For Chrome, last I checked, it avoided Selenium::WebDriver::Error::UnknownError, which is Chrome's lack of specific ElementClickInterceptedError.

titusfortner:
yeah, it's a pain that Chrome is the only non-w3c compliant driver at this point.
Probably best to wait_for_present; I can't think of any use case that doesn't include wanting to click the element.
Do we get an error / handle the error if the element doesn't exist?

Rajagopalan-M:
@titusfortner @jkotests 

I have some other idea to handle this when it's chrome and that's what I have done in my local project. Please look at this code and let me know whether it's okay

>       rescue Selenium::WebDriver::Error::NoSuchWindowError
>         raise Exception::NoMatchingWindowFoundException, 'browser window was closed'
>       rescue =>e
>          if e.message.include? "Other element would receive the click"
>               retry            
>         end
>         raise_present unless Wait.timer.remaining_time.positive?
>      end

We can do this arrangement until Chrome releases standarad exception like FireFox(ElementClickInterceptedError)

What do you say?

titusfortner:
This PR should allow you to do the same thing more elegantly, but it still can't be rolled in as a default.

p0deje (approved):
I think there will be plenty of cases when this snippet will return incorrect results in future, but let's start with it. 👍

p0deje:
Please, keep in mind that will not prevent all the cases for using `rescue` because element might become unclickable between `obscured?` and `click` calls. But of course, the number of exception catching should drop.

jkotests:
I made the following changes to address the questions/concerns:
* `#obscured?` now requires the element to be present. The UnknownObjectException will be raised if it's hidden or does not exist. 
* Expanded the tests to also check that we are approximating what happens in the click - ie `#obscured?` should be true if `#click` will raise exception.

Tests passed locally using Chrome, Firefox and IE.

jkotests:
Great, the relaxed locate tests are failing. I'll take a look.

jkotests:
Can you remind me the history of relaxed_locate? If relaxed_locate is `false`, we don't do any waiting for `element_call`? For example, I'm calling `Element#wait_for_present`, which exits before the waiting:

~~~~~~~~
    def wait_for_present
      p = present?
      return p if !Watir.relaxed_locate? || p
~~~~~~~~

Should I be using a different method in `#obscured?` to wait for the element to be present? Or are tests that rely on `#wait_for_present` just not compliant with relaxed_locate = false?

titusfortner:
the `wait_for_*` methods should theoretically only be used for when taking actions on an element via `#element_call`.

The reason for the `#relaxed_locate` toggle was because some users were clicking on an element and using the resulting exception as part of their flow control. Essentially they were intentionally calling `#click` on an element that they expected not be present. Setting `Watir.relaxed_locate = false` is a way for each of those calls to not take an additional 30 seconds when they wanted to get the exception immediately.

All predicate methods on element: (`#exists?`, `#enabled?`, `#stale?`, `#present?`, `#visible?`) should immediately return a boolean value.

So we probably want something like:

```
def obscured?
    present? && element_call { execute_js(:obscured, @element) }
end
```

`element.wait_while(&:obscured?)` will poll until it goes away, and if the element might not be there in the first place, I think they should need to do: `element.wait_until(&:present?).wait_while(&:obscured?)` rather than us trying to figure out what that race condition might look like.

titusfortner:
Just noticed that `#focused?` really should be using `:assert_exist` in the `element_call` but that slipped through...

jkotests:
@titusfortner , the tests are passing now, except for on AppVeyor. That's a general problem we have right now, right?

titusfortner (approved):

PR #806, opened 2018-10-13
Fix Appveyor Failures

titusfortner:
Is there a reason we need to delete this file? Or does the filespace not get cleared out on Travis/Appveyor?


p0deje:
We need to delete it because otherwise, it will keep file descriptor open, won't it? It's not only about CI, but also running tests locally.

If you try to deal with permission denied error, please see https://github.com/SeleniumHQ/selenium/pull/6444/commits/058cd4a53b6b73674d741c9ba9d3ff60863ba660.

titusfortner:
Ok, the tests are running with no weird errors. These two look like legit issues.

https://ci.appveyor.com/project/titusfortner/watir/builds/21024449/job/pmqjga5cm0mx52j2

@jkotests Firefox on Windows doesn't seem to like your `#obscured?` code. Is it a new bug with Firefox/geckodriver?

@p0deje any thoughts on why the crossframe JavaScript error is happening? Should we file a bug with geckodriver?

titusfortner:
merged with https://github.com/watir/watir/commit/d06e06b9cfa2e522ff538ccdd2b843434640222a

PR #807, opened 2018-10-14
Unsub xpath

titusfortner:
This is a great refactor for simplifying the code I've been working on. I also figured out that we were doing some weird things with text field value attribute vs property.

### DIFF
--- a/lib/watir/locators/button/selector_builder/xpath.rb
+++ b/lib/watir/locators/button/selector_builder/xpath.rb
@@ -3,71 +3,82 @@ module Watir
     class Button
       class SelectorBuilder
         class XPath < Element::SelectorBuilder::XPath
-          def build(selector)
-            return super if selector.key?(:adjacent)
+          private
 
-            selector[:tag_name] = 'button'
+          def tag_string
+            return super if @adjacent
 
-            type = selector.delete :type
-            return super(selector) if type.eql?(false)
+            # Selector builder ignores tag name and builds for both button elements and input elements of type button
+            @selector.delete(:tag_name)
 
-            # both :value and :text selectors will locate elements by value attribute or text content
-            selector[:text] = selector.delete(:value) if selector.key?(:value)
+            type = @selector.delete(:type)
+            text = @selector.delete(:text)
 
-            wd_locator = super(selector)
-
-            start_string = default_start
-
-            button_string = "local-name()='button'"
-            common_string = wd_locator[:xpath].gsub(start_string, '').gsub("[#{button_string}]", '')
-
-            input_string = "(local-name()='input' and #{input_types(type)})"
-
-            tag_string = if type.nil?
-                           "[#{button_string} or #{input_string}]"
-                         else
-                           "[#{input_string}]"
-                         end
-
-            xpath = "#{start_string}#{tag_string}#{common_string}"
-
-            {xpath: xpath}
+            string = "(#{button_string(text: text, type: type)})"
+            string << " or (#{input_string(text: text, type: type)})" unless type.eql?(false)
+            "[#{string}]"
           end
 
-          protected
+          def button_string(text: nil, type: nil)
+            string = process_attribute(:tag_name, 'button')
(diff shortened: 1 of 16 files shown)


p0deje (approved):

lib/watir/locators/button/selector_builder/xpath.rb
```diff
@@ -3,68 +3,57 @@ module Locators
...
-            {xpath: xpath}
-          end
+            @selector.delete(:tag_name)
 
-          protected
+            type = @selector.delete :type
 
-          def use_index?
-            false
+            if type
```
jkotests: The type attribute is valid for `button` and `input` elements, so we cannot assume a tag name based on its presence.

For example, given:

~~~~~~~~
<button id="btn_type" type="submit">text</button>
<input id="id_type" type="submit" value="text">
~~~~~~~~

With the new changes:

~~~~~~~~
browser.buttons(type: 'submit').map(&:id)
#=> ["id_type"]
~~~~~~~~

In 6.14:

~~~~~~~~
browser.buttons(type: 'submit').map(&:id)
#=> ["btn_type", "id_type"]
~~~~~~~~

lib/watir/locators/button/selector_builder/xpath.rb
```diff
@@ -3,68 +3,57 @@ module Locators
...
+              "[local-name()='button' or (local-name()='input' and #{input_types(type)})]"
+            else
+              "[local-name()='button']"
+            end
           end
 
           # This is special because text locator for buttons match text or value
-          def add_text
-            return '' unless @selector.key?(:text)
+          def text_string
```
jkotests, line 41: We need to consider the tag name when deciding to check text or value:
* For button elements, `:text` only matches the element text
* For input elements, `:text` only matches the value attribute

Given the following button, which has different text and value attribute:

~~~~~~~~
<button id="btn_dif_value" value="btn_dif_value">btn_dif_text</button>
~~~~~~~~

With these changes, `:text` will now match the text and value:

~~~~~~~~
browser.buttons(text: 'btn_dif_value').map(&:id)
#=> ["btn_dif_value"]
browser.buttons(text: 'btn_dif_text').map(&:id)
#=> ["btn_dif_value"]
~~~~~~~~

Where as 6.14 would only match the text:

~~~~~~~~
browser.buttons(text: 'btn_dif_value').map(&:id)
#=> []
browser.buttons(text: 'btn_dif_text').map(&:id)
#=> ["btn_dif_value"]
~~~~~~~~

jkotests (dismissed):

lib/watir/locators/button/selector_builder/xpath.rb
titusfortner: grrr, good catch. Thanks.

lib/watir/locators/button/selector_builder/xpath.rb
titusfortner: Ah, if it is an actual button we need to check the text only, and if it is an input we need to check the value only. Got it.

titusfortner: @jkotests ok, so what is the right answer here:

```html
<button name="button_2" type="submit" value="button_2">Button 2</button>
```

```ruby
expect(browser.button(value: 'Button 2')).to exist
expect(browser.button(value: /Button 2/)).to exist
```

I kind of feel like this should only match `button_2` and not `Button 2`

I guess I should keep it backward compatible for now, though. :-/

jkotests: Oh. I don't think I ever noticed that. Yeah, I agree, it should be only matching the value attribute. More stuff for the list of things to deprecate.

titusfortner: I made it backward compatible. I'll add it to the list of things to deprecate. :)

titusfortner: It's like this logic is complicated or something...

PR #809, opened 2018-10-16
Implements Capybara's advanced Regexp Matching for XPath generation

titusfortner:
This restores the functionality Watir had in 6.14 that was removed in previous PRs. The previous implementation handled some of the wild cards incorrectly. This does a more thorough job of parsing partial matches for adding into the XPath


p0deje (approved):
My eyes are bleeding, but as long as it works, I approve.

titusfortner:
Right? I'm mostly treating that class like a black box. :)

titusfortner:
Ok, Capybara changed implementations. the `regexp_parser` gem does everything the previous implementation did, plus better support for escaped characters.

Rajagopalan-M:
This is something amazing, we get knowledge from people from cabypara since
it is also sit on the top of Ruby Selenium Binding.

titusfortner:
@Rajagopalan-M well, Capybara was stealing stuff from Watir and then made it much better in the process, so Watir is just stealing it back now. (seriously, though, we've been discussing ways to approach these common challenges, and I agree it is a very good thing).

Rajagopalan-M:
Ah! Okay!

jkotests (approved):

titusfortner:
whoops, didn't squash this.

PR #815, opened 2018-11-07
improve type check with constant instead of multiple conditionals

titusfortner:
This uses a Hash instead of a Case Statement. 
Also improves error messages, in my opinion.
Fixnum/Integer stuff can get removed when we deprecate 2.3

Also threw in update to run on latest Ruby point releases on Travis.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,6 +35,10 @@ Lint/HandleExceptions:
   Exclude:
     - 'lib/watirspec.rb'
 
+Lint/UnifiedInteger:
+  Exclude:
+  - 'lib/watir/locators/element/selector_builder.rb'
+
 # Configuration parameters: CountComments, ExcludedMethods.
 # ExcludedMethods: refine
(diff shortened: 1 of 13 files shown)


p0deje (approved):
Not sure it's much cleaner now, but up to you.

jkotests (approved):

titusfortner:
I think that when assignment values do not require additional logic to evaluate, then a Hash should be preferred over a "large" case statement. I'm finding more patterns in the code I've been writing in various languages where things like lookups from files containing configuration data make things significantly more clean than large case statements (think Java where the additional boiler plate increases the number of lines over Ruby). Also, when Rubocop complains about method length, this is one way to minimize the size of a method. This solution inarguably reduced the lines of code (though there are likely alternatives to this solution that could have done something similar without the Hash).

Article referencing this type of refactoring: https://franzejr.github.io/best-ruby/refactorings/case_with_hashes.html

Issue #813, opened 2018-11-06
Execute_script doesn't wait for argument elements to exist


abotalov:
## Expected Behavior -
`Browser#execute_script` waits for argument elements to exist.

## Actual Behavior -
`Browser#execute_script` doesn't seem to wait for its argument elements to exist.
It calls `wd` that calls `assert_exists`. `assert_exists` raises an error if element isn't found.
I think it would make sense to wait for elements to exist.

titusfortner:
This seems reasonable. Any chance you want to make a PR? :)

reach2jeyan:
I'll take a look at this and try to raise a PR.

PR #820, opened 2018-11-12
BugFix for the issue #813. execute_script does not wait for element

reach2jeyan:
@titusfortner This is my first PR. Can you please provide your comments?

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -699,7 +699,7 @@ module Watir
 
     # Locates if not previously found; does not check for staleness for performance reasons
     def assert_exists
-      wait_for_exists #Added method for waiting for element to exist
+      
       locate unless located?
       return if located?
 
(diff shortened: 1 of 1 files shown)


titusfortner:
Ok, the first step is to add a test that fails under current code and passes with new code.
The second step is to run the existing specs to make sure that they pass with the new code. In this case, `#wait_for_exists` calls `#assert_exist`, so you're going to get a Stack Level error: https://travis-ci.org/watir/watir/jobs/453874195

There are two places that call `#execute_script` and the idea is to wait for any elements that are passed in; so like inside the block here: https://github.com/watir/watir/blob/master/lib/watir/browser.rb#L216

titusfortner:
@reach2jeyan are you still interested in working on this? What help do you need?

reach2jeyan:
@titusfortner Can you let me know if we can jump into a quick slack call?

titusfortner:
@reach2jeyan what is your timezone again? Message me on slack and we'll figure something out.

titusfortner:
So... #813 fix was too simple to walk through the PR for it. Essentially specs already exist for the functionality we needed to use. We'll walk through a different issue tomorrow. Fortunately(?) there are 29 more open issues we can go through to find something to work on.

PR #822, opened 2018-11-18
Nested elements

titusfortner:
Ok, so this is where things start getting interesting...

`#cache=` is supposed to allow for distinguishing between
1. > This element is defined as a wrapping of a Selenium Element and can not be relocated
2. > This element can be relocated by the provided Selector, but the Selenium Element is already located, so we add it as cached.

This PR builds the selector when Element is initialized rather than when it is located.
This allows us to determine on initialization if a single XPath can be created for nested elements, and allows us to bypass recursion during location.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -797,7 +797,7 @@ module Watir
 
       begin
         check_condition(precondition, caller)
-        Watir.logger.info "-> `Executing #{inspect}##{caller}`"
+        Watir.logger.debug "-> `Executing #{inspect}##{caller}`"
         yield
       rescue unknown_exception => ex
         element_call(:wait_for_exists, &block) if precondition.nil?
@@ -822,7 +822,7 @@ module Watir
       rescue Selenium::WebDriver::Error::NoSuchWindowError
         raise NoMatchingWindowFoundException, 'browser window was closed'
       ensure
-        Watir.logger.info "<- `Completed #{inspect}##{caller}`"
+        Watir.logger.debug "<- `Completed #{inspect}##{caller}`"
         Wait.timer.reset! unless already_locked
       end
     end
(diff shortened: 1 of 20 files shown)


p0deje (commented):
I don't fully follow the changes, so let me withdraw from reviewing. I fully trust your judgement :)

lib/watir/elements/element.rb
```diff
@@ -616,6 +631,16 @@ def located?
       !!@element
     end
 
+    #
+    # @api private
+    #
+    # Set the cached element. For use when
```
jkotests: Sentence ends abruptly.

lib/watir/elements/radio.rb
```diff
@@ -1,8 +1,12 @@
 module Watir
   class Radio < Input
-    def initialize(query_scope, selector)
-      super
+    def build
       @selector[:label] = @selector.delete(:text) if @selector.key?(:text)
+      super
+    end
+
+    def driver_scope
```
jkotests: What is different about `Radio` that requires this? I'm still trying to work through the code, so maybe I'm just missing something. Just seems weird that other places that you added `#driver_scope` were in the `Locator`, where as this is the element itself.

titusfortner: Getting `RadioSet` to work was a pain. All that code made sense when I created it, but it was very confusing digging back into it. :-D I'm guessing I needed to have it here to get that to work, but I don't recall. I'll try removing it and see what happens. I should add more comments for these things in general.

jkotests:
@titusfortner , is there still value in reviewing this or should we wait until it can be merged with the overhaul (#825)?

titusfortner:
Hold off, I got halfway through a rebase before getting pulled away.

titusfortner:
@jkotests ok now everything is passing with the latest code merged in

titusfortner:
Ugh. `#merge` and `#merge!` are different. This one is temporarily no longer ready for review. :(

titusfortner:
@jkotests It works!!!

This is based on #826; we can do them at the same time or separately.

This won't work for IFrames, Tables, or Adjacent location, which are probably the 3 main use cases, but for the ever-popular Page Sections, this should be a huge performance bump.

Also, I kind of don't want to touch location again in 2019. :)

jkotests (approved):

Issue #824, opened 2018-11-22
Locators with hyphens and conditional regexps are not working on watir 6.15


danielfsbarreto:
## Meta -
Watir Version: 6.15
Selenium Version: 3.141.0
Browser Version: 70.0.3538.77 (Official Build) (64-bit)
Browser Driver Version: ChromeDriver 2.41.578700 (2f1ed5f9343c13f73144538f15c00b370eda6706)
OS Version: Ubuntu 18.04.1 LTS

## Expected Behavior -
Locators containing hyphens and conditionals like `row.divs(class: /(task-description|is-task_description)/).find(&:present?)` should work just fine, as it is composed by an absolutely valid regular expression. This snippet behaves as expected on 6.14.

## Actual Behavior -
```ruby
row.divs(class: /(task-description|is-task_description)/).find(&:present?)
```
```
RegexpError: empty range in char class: /[\/(task-description|is-task_description)\/]/
from /home/danielfsbarreto/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/watir-6.15.0/lib/watir/locators/element/locator.rb:171:in `block in matches_values?'
```

## Steps to reproduce -
```ruby
gem 'watir', '=6.15'
require 'watir'

browser = Watir::Browser.new
browser.goto 'https://jsfiddle.net/hx0tp3jg/'
browser.iframe(name: 'result').div(class: /(task-description|is-task_description)/).text
```
```
Traceback (most recent call last):
       16: from /home/danielfsbarreto/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/watir-6.15.0/lib/watir/elements/element.rb:603:in `locate'
       15: from /home/danielfsbarreto/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/watir-6.15.0/lib/watir/elements/element.rb:715:in `locate_in_context'
       14: from /home/danielfsbarreto/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/watir-6.15.0/lib/watir/locators/element/locator.rb:27:in `locate'
       13: from /home/danielfsbarreto/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/watir-6.15.0/lib/watir/locators/element/locator.rb:67:in `using_watir'
[...]

PR #825, opened 2018-11-25
Overhaul Locator Matching

titusfortner:
So, yeah, I did this over the weekend...

This pretty much came about because I couldn't figure out the right place to put a spec for #824. 
So, I completely redid the unit tests, and reorganized to code to make unit testing easier.

Instead of testing the whole thing with locator unit tests, we break it into 3 independent parts

1. Selector Builder (Hash input, Hash output)
    * I moved specs from Watirspecs into Unit Tests to focus on just Selection Building components
2. Locator (find_element & find_elements calls)
3. Matcher
    * This rolls in the Validator code as well. since you shouldn't need to validate anything if you are properly parsing all of the SelectorBuilder output
    * This allows us to mock out the incoming elements and just focus on the matching code
    * This won't replace all of the watirspecs because I do want Browser testing, and external projects will need to execute against those, but we should be able to slim it down a bunch.

Implementation wise I got rid of the `use_selenium` and `use_watir` since there's no good reason to shortcut since no extra wire calls are getting made. 

I have no idea how this will merge with #822 but I think this is a major step in improving the separation of concerns we've been working toward.


lib/watir/locators/element/matcher.rb
```diff
@@ -0,0 +1,128 @@
...
+        end
+
+        def elements_match?(element, values_to_match)
+          matches = values_to_match.all? do |how, expected|
+            if how == :tag_name
+              validate_tag(element, expected)
+            # TODO: Can this be class_name here or does that get converted?
+            elsif %i[class class_name].include?(how)
+              [expected].flatten.all? do |match|
+                fetch_value(element, how).split.any? do |class_value|
```
jkotests: Pulling `fetch_value` out of the loop will save some wire calls.

lib/watir/locators/element/selector_builder.rb
```diff
@@ -132,6 +164,18 @@ def raise_unless(what, types)
 
           raise TypeError, "expected one of #{types}, got #{what.inspect}:#{what.class}"
         end
+
+        def deprecated_locators
+          %i[partial_link_text link_text link].each do |locator|
+            next unless @selector.key?(locator)
+
+            Watir.logger.deprecate(":#{loctor} locator", ':visible_text', ids: [:link_text])
```
jkotests: Typo in "loctor" variable - missing "a". 

Given we know that some people are using these locators, we should consider adding a test for them. Right now, there only seem to be tests for converting :visible_text to this?

lib/watir/locators/element/selector_builder.rb
```diff
@@ -80,16 +109,19 @@ def normalize_locator(how, what)
           when :tag_name
             what = what.to_s if what.is_a?(::Symbol)
             [how, what]
-          when :text, :xpath, :index, :class, :css, :visible, :visible_text, :adjacent
+          when :text, :xpath, :index, :css, :visible, :visible_text, :adjacent
+            [how, what]
+          when :class
+            what = false if what.tap { |arr| arr.delete('') }.empty?
```
jkotests, line 114: I'm undecided on whether this matters, but this causes a change in behaviour:

~~~~~~~~
# 6.15
p browser.ul(class: ['']).exists?
#=> false
p browser.ul(class: []).exists?
#=> Can not locate elements with an empty Array for :class (Watir::Exception::LocatorException)

# Proposed code
p browser.ul(class: ['']).exists?
#=> false
p browser.ul(class: []).exists?
#=> false
~~~~~~~~

I'm honestly not sure what I would expect in either of these cases. Both seem like a mistake, so raising an exception makes sense. On the other hand, I guess it's possible someone is dynamically generating the Array, so they might want empty to be accepted.

lib/watir/locators/element/selector_builder.rb
```diff
@@ -132,6 +164,18 @@ def raise_unless(what, types)
...
           raise TypeError, "expected one of #{types}, got #{what.inspect}:#{what.class}"
         end
+
+        def deprecated_locators
+          %i[partial_link_text link_text link].each do |locator|
+            next unless @selector.key?(locator)
+
+            Watir.logger.deprecate(":#{loctor} locator", ':visible_text', ids: [:link_text])
+            tag = @selector[:tag_name]
+            next if tag.nil? || [:a, :link].include?(tag.to_sym)
```
jkotests: Are you sure `:link` should be here? I thought link tags are only for external resources.

jkotests (changes requested):
I like where this is going. The code is definitely getting more clear.

lib/watir/locators/element/selector_builder.rb
titusfortner: So, the change was intentional. I believe an empty array should act like an empty string had always acted. I don't think it counts as breaking backward compatibility if you are just removing an exception. :)

jkotests: Fair enough.

titusfortner:
This needed the `#cache=` implementation to pass the RadioSet tests, so I added that.

lib/watir/locators/element/selector_builder.rb
titusfortner: Yeah, looks like I added it back with [this](https://github.com/watir/watir/commit/a933eb64354a5396a8d2fcdd80afc4cc2cbdfd34#diff-7bb35ecef9bd09742767229a044a1c9aR259), but since `#link` is just an alias of `#a`, it will always have a `:tag_name` of `a`.

titusfortner:
> I like where this is going. The code is definitely getting more clear.

@jkotests Yeah, well, you were the one who originally got me thinking about how the SOC could be improved. Then the idea of trying to create these things as distinct "services" got me thinking about how to implement it.

lib/watir/locators/element/matcher.rb
titusfortner: good call. fixed.

lib/watir/locators/element/selector_builder.rb
titusfortner: yeah, I didn't add coverage to things that were deprecated because the fix for bugs with deprecated code is to update to the suggested approach...
i.e. I'm not fixing it if it's broken because I'm just going to delete it all hopefully sooner than later. :-D

spec/watirspec/radio_set_spec.rb
```diff
@@ -61,6 +61,7 @@
 
   context 'without name specified' do
     it 'Finds specified radio' do
+      browser.link
```
jkotests: Seems like a strange thing to add?

titusfortner: I... still don't know what this is from. I need to investigate. :)

titusfortner: hah. That was me committing a file that I didn't mean to commit. All fixed now.

jkotests (approved):

PR #833, opened 2018-12-16
Use javascript to get tag_name for ElementCollection elements

lmtierney:
This replaces #802, updated due to rebase conflicts

-----

When the `selector` does not contain `:tag_name`, we are currently querying the `tag_name` from the element. This causes `N` wire calls which, depending on how many elements there are, can take a while when testing remotely.

The idea is to us the javascript executor to retrieve all the tags once. In doing so, we do miss out on the added benefit of `element_call` which `#tag_name` uses, but that shouldn't be an issue as we have already located the elements.

The only concern I could think of was with elements going stale between retrieving them and then performing javascript on them, but this should be a rare scenario. This is mitigated by the rescue and retry. Theoretically there is a chance for an infinite loop on the retry, but the likelihood of that is near impossible in normal conditions.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -5,6 +5,8 @@ module Watir
 
   class ElementCollection
     include Enumerable
+    include Exception
+    include JSSnippets
     include Locators::ClassHelpers
 
     def initialize(query_scope, selector)
@@ -92,12 +94,12 @@ module Watir
     def to_a
       hash = {}
       @to_a ||=
-        elements.map.with_index do |el, idx|
+        elements_with_tags.map.with_index do |(el, tag_name), idx|
           selector = @selector.dup
           selector[:index] = idx unless idx.zero?
           element = element_class.new(@query_scope, selector)
           if [HTMLElement, Input].include? element.class
-            construct_subtype(element, hash).tap { |e| e.cache = el }
+            construct_subtype(element, hash, tag_name).tap { |e| e.cache = el }
           else
             element.tap { |e| e.cache = el }
           end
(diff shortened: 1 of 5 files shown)


lib/watir/element_collection.rb
```diff
@@ -162,6 +163,23 @@ def elements
...
     end
 
+    def elements_with_tags
+      els = elements
+      if @selector[:tag_name]
+        els.map { |e| [e, @selector[:tag_name]] }
+      else
+        retries = 0
+        begin
+          els.zip(browser.wd.execute_script('return arguments[0].map(function(e) {return e.localName});', els))
```
titusfortner: This shouldn't have `#wd`

titusfortner (changes requested):
Just the `#wd` removal. Nice job on the specs.

lib/watir/element_collection.rb
lmtierney: It has to or it will add an extra wait which is unnecessary

lmtierney: Or rather, it calls `wrap_elements_in` which defeats the purpose of the PR

titusfortner: It shouldn't, it's passing in a Selenium element not a Watir element...

titusfortner: ohh, wrap, right...

lmtierney: You're right, it doesn't return elements anymore; just the `tag_name` so it won't wrap anything. Should be fine to remove it.

lmtierney:
Updated and moved the js to a snippet. Also added a new spec to ensure we're getting the right container back without specifying `tag_name`

titusfortner (approved):

PR #837, opened 2018-12-17
frame switching changes the DOM and should therefore execute after hooks

titusfortner:
I don't think After Hooks are very widely used, so I'm not sure how big of a difference this would make.

Anyone see any reason not to add this?

I need it for Watigiri since the DOM changes, so if we don't want to add it, I'll need to monkey patch that project.

### DIFF
--- a/lib/watir/browser.rb
+++ b/lib/watir/browser.rb
@@ -263,8 +263,11 @@ module Watir
     end
 
     def ensure_context
-      driver.switch_to.default_content unless @default_context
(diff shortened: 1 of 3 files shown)


p0deje (approved):
Not sure why DOM changes in your case, but it looks reasonable.

jkotests (approved):
I don't think I understand the use case, but don't see any harm in adding it.

lmtierney (approved):

PR #840, opened 2018-12-20
deprecate using a label locator with an invalid element type to find …

titusfortner:
I'm not sure why we are allowing for any element to use the `:label` locator so long as it isn't a valid attribute. This should be restricted to the elements that support it.

Not sure what the actual fixed implementation will look like, I figured hard coding an array with the allowed element tags would be sufficient for the deprecation notice.

### DIFF
--- a/lib/watir/locators/element/selector_builder.rb
+++ b/lib/watir/locators/element/selector_builder.rb
@@ -100,7 +100,10 @@ module Watir
         end
 
         def should_use_label_element?
-          !valid_attribute?(:label)
+          return false if valid_attribute?(:label) # This is only Option, OptGroup & Track
+
+          deprecate_label_location(@selector[:tag_name])
+          true
         end
 
         def normalize_locator(how, what)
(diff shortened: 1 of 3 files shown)


lib/watir/locators/element/selector_builder.rb
```diff
@@ -178,6 +181,17 @@ def deprecated_locators
             raise LocatorException, "Can not use #{locator} locator to find a #{tag} element"
           end
         end
+
+        def deprecate_label_location(tag_name)
+          return unless tag_name.nil? || !%w[input button meter select textarea].include?(tag_name)
```
p0deje, line 186: This should be extended: `https://html.spec.whatwg.org/#category-label`

p0deje (approved):
I see no harm in allowing to use label with any other element, but if we want to stick to specification, we should probably restrict that. I'd prefer us to extract the elements which can use label from specification though.

jkotests (approved):

titusfortner:
@p0deje you make a good point, I'm not sure what strictly following the spec here gains us. I'll think on it more before deciding if I want to merge it. Thanks for the feedback.

titusfortner:
yeah, there's no compelling reason to limit this. closing...

PR #846, opened 2018-12-28
changed desired implementation of select lists

titusfortner:
Update: After further consideration of the code, I don't want to support multiple arguments here. If you need multiple selections, put it in an Array; if you need one selection, put it in the single parameter and it will probably do what you want. If you need to specify it, use one of the keywords. We also need to decide if this goes in 6.18 or 7.0.

--------
After looking at what code I wanted in Watir 7, and the performance of the wire calls to obtain it, I realized that we don't need `Select#select_all`, which I created forever ago in July 2017. My apologies to everyone who updated their code to use it.

I'm not sure why we aren't checking the label for `#include?`

This is the proposed API:

```ruby
# Automatically does what you likely want
browser.select(select_locator).select(value)
browser.select(select_locator).select(text)
browser.select(select_locator).select(label)

# If you need to specify a specific thing to look for (or desperately need to save wire calls)
browser.select(select_locator).select(value: value)
browser.select(select_locator).select(text: text)
browser.select(select_locator).select(label: label)

# To select multiple matching on multi-select lists, you need to put it in an array, or multiple params
browser.select(select_locator).select([/opt/])
browser.select(select_locator).select(option1, option2)
browser.select(select_locator).select([option1, option2]])
browser.select(select_locator).select(value:[value1, value2])
browser.select(select_locator).select(text: [text1, text2])
browser.select(select_locator).select(label: [label1, label2])
```

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -18,7 +18,7 @@ module Watir
     #
 
     def include?(str_or_rx)
-      option(text: str_or_rx).exist? || option(label: str_or_rx).exist?
+      option(text: str_or_rx).exist? || option(label: str_or_rx).exist? || option(value: str_or_rx).exist?
     end
 
     #
@@ -29,11 +29,13 @@ module Watir
     # @return [String] The text of the option selected. If multiple options match, returns the first match.
     #
 
-    def select(*str_or_rx)
-      if str_or_rx.size > 1 || str_or_rx.first.is_a?(Array)
-        str_or_rx.flatten.map { |v| select_all_by v }.first
+    def select(*str_or_rx, text: nil, value: nil, label: nil)
+      key, value = parse_select_args(str_or_rx, text, value, label)
+
+      if value.size > 1 || value.first.is_a?(Array)
+        value.flatten.map { |v| select_all_by key, v }.first
       else
-        select_matching([find_option(str_or_rx.flatten.first)])
+        select_matching([find_option(key, value.flatten.first)])
       end
     end
     alias set select
@@ -45,11 +47,13 @@ module Watir
     # @raise [Watir::Exception::NoValueFoundException] if the value does not exist.
     #
 
-    def select!(*str_or_rx)
-      if str_or_rx.size > 1 || str_or_rx.first.is_a?(Array)
-        str_or_rx.flatten.map { |v| select_by! v, :multiple }.first
+    def select!(*str_or_rx, text: nil, value: nil, label: nil)
+      key, value = parse_select_args(str_or_rx, text, value, label)
+
+      if value.size > 1 || value.first.is_a?(Array)
+        value.flatten.map { |v| select_by! key, v, :multiple }.first
       else
-        str_or_rx.flatten.map { |v| select_by! v, :single }.first
+        value.flatten.map { |v| select_by! key, v, :single }.first
       end
     end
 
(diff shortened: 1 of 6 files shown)


lib/watir/elements/select.rb
```diff
@@ -29,9 +29,16 @@ def include?(str_or_rx)
...
     #
 
-    def select(*str_or_rx)
-      results = str_or_rx.flatten.map { |v| select_by v }
-      results.first
+    def select(*both, text: nil, value: nil, label: nil)
+      selection = {both: both, text: text, value: value, label: label}.select do |_k, v|
+        !v.nil? && ![v].flatten.empty?
+      end
+      raise 'Can not select by more than one method' if selection.size > 1
```
jkotests: Can we include the result of `selection` in the error message? Just in case some one runs into the error message and asks what to do, the error message will reduce the back-and-forth questions.

lib/watir/elements/select.rb
```diff
@@ -179,6 +190,19 @@ def process_str_or_rx(str_or_rx)
       end
     end
 
+    def type_check(value)
+      msg = "expected String, Number or Regexp, got #{value.inspect}:#{value.class}"
```
jkotests: Do we actually support Numeric? As much as it's in the old `#find_options` case statement, I don't think we actually support it.

lib/watir/elements/select.rb
```diff
@@ -54,9 +62,16 @@ def select_all(*str_or_rx)
     # @raise [Watir::Exception::NoValueFoundException] if the value does not exist.
     #
 
-    def select!(*str_or_rx)
-      results = str_or_rx.flatten.map { |v| select_by!(v, :single) }
-      results.first
+    def select!(*both, text: nil, value: nil, label: nil)
+      selection = {both: both, text: text, value: value, label: label}.select do |_k, v|
```
jkotests: `#select` and `#select!` should always accept the same types of arguments, so it'd be nice to extract this to be shared code.

lib/watir/elements/select.rb
```diff
@@ -54,9 +62,16 @@ def select_all(*str_or_rx)
...
+        !v.nil? && ![v].flatten.empty?
+      end
+      raise 'Can not select by more than one method' if selection.size > 1
+
+      value = selection.values.first
+      type_check(value)
+
+      [value].flatten.map { |v| select_by! v, :multiple }.first
```
jkotests: Same as for `#select`, we don't seem to actually do anything with the `selection.keys.first`.

lib/watir/elements/select.rb
```diff
@@ -29,9 +29,16 @@ def include?(str_or_rx)
...
+    def select(*both, text: nil, value: nil, label: nil)
+      selection = {both: both, text: text, value: value, label: label}.select do |_k, v|
+        !v.nil? && ![v].flatten.empty?
+      end
+      raise 'Can not select by more than one method' if selection.size > 1
+
+      value = selection.values.first
+      type_check(value)
+
+      [value].flatten.map { |v| select_by v }.first
```
jkotests: I like the idea of being able to specify how to locate the option. I know I've had at least one time where this would have helped.

However... we don't actually seem to be doing anything with it - ie shouldn't selection.keys.first get based along/used?

jkotests (changes requested):

lib/watir/elements/select.rb
titusfortner: No, we don't. I removed it from the code and forgot to remove it from the message.

lib/watir/elements/select.rb
titusfortner: Yeah, optional keywords for the win again.
`#first` is returning the value of the first item selected. Not certain why that's what we want returned, but that's the backward compatible thing.

lib/watir/elements/select.rb
titusfortner: The refactor for Watir 7 is slightly different since we need to support single, multiple and normal. This was going to be the easy/working copy/paste in the meantime.

p0deje (approved):

lib/watir/elements/select.rb
titusfortner: Actually, I went ahead and added support for Numeric because it was pretty trivial, and why not.

titusfortner:
Ok, I updated everything, mostly just supporting numeric, and added a few tests to make sure everything is correct. @jkotests what do you think?

lib/watir/elements/select.rb
jkotests: I meant that `#select_by` does not know to select by `:text`, `:value`, vs `:label`. The specified keyword gets ignored.

For example, given the page:

~~~~~~~~
<html>
  <body>
    <select>
      <option value="a">1</option>
      <option value="b">2</option>
      <option value="c">3</option>
    <select>
  </body>
</html>
~~~~~~~~

If I try to to pick the `:value` of "3":

~~~~~~~~
browser.select_list.select(value: '3')
p browser.select_list.selected_options.map(&:text)
#=> ["3"]
~~~~~~~~

The option with `:text` of "3" has been selected (unexpected). I would only expect to pick options with the matching value attribute.

jkotests (changes requested):

titusfortner:
I've updated based on the feedback, and I think this does all the things we want it to do now. Specs added as well.

jkotests (approved):

Issue #848, opened 2019-01-08
Watir::Wait::TimeoutError when trying to interact with the original window


Blackbaud-KimberlyHammer:
## Meta -
Watir Version: 6.16.5
Selenium Version:  3.141.0
Browser Version: Firefox 57.0
Browser Driver Version: Using webdrivers 3.6.0 to get the latest driver
OS Version: Ubuntu

## Expected Behavior - 
We should still be able to locate and interact with elements in the original open window after closing a window we no longer need.

## Actual Behavior -
We often have to open up additional windows, run tests in those windows and then go back to the original window to continue running tests. When we use watir 6.10.2 in headless mode everything works great. When we try upgrading to 6.14.0 or 6.16.5, after we close a second window using browser.windows.last.close, we're unable to get the tests to interact with the original window in headless mode. We keep seeing Watir::Wait::TimeoutError when trying to locate elements in the original window.  I've tried the following, but nothing seems to work:
browser.windows.first.use
browser.windows.last.use
browser.driver.switch_to.window(browser.driver.window_handles[0]) 
browser.window(:title => /eTapestry/).use

Is there something else I need to do to be able to located and interact with elements in the window that is still open?

The original window we're trying to interact with does contain an iFrame browser.iframe(:id=>'etapContentIframe')

The only workaround I can find is to perform browser.refresh on the window I want to interact with, but this is not ideal.

titusfortner:
Hi @Blackbaud-KimberlyHammer does this work Firefox non-headless? Are you using Headless gem or `ENV['MOZ_HEADLESS']`?

Blackbaud-KimberlyHammer:
Thank you for the quick response @titusfortner. When I run the tests in Firefox non-headless it works if I only have the new window open for a very brief time. I'm able to reproduce the timeout error if the test is in the new window for a while before going back to the original window.  We are using the Headless gem.

When we use watir 6.10.2 it worked in both headless and non-headless modes regardless of how long we were in the new window.

titusfortner:
Hmm, we haven't really changed anything windows related since 6.10.
Also, there really isn't anything that is affected by how long a window is open.

Are you using this with the watirmark gem, or is this a direct watir implementation?

It is possible (for 'reasons') that this will not do the right thing: `browser.windows.last.close`
For 2 windows my current recommendation is:
```
browser.window(new_window_opts).use
do_things_in_new_window
browser.window.close
browser.original_window.use
```

[...]

PR #849, opened 2019-01-09
Window switch

titusfortner:
#848 reminded me that I don't like the current window switching code.

First, when working with safari driver recently I definitively came across a case where a newly opened window was indexed first in `#window_handles` call, so this really needs to get deprecated. 

If we aren't using indexes, then we need a way for people to avoid having to know title or url, since 90% of the time, users who are working with windows are only going to be dealing with two windows at most; they really just want: "give me the other one."

Right now, and no longer effective: `browser.windows.last.use`
Only option with current code: `browser.window(title_or_url_opt).use`
Proposed: `browser.switch_window`

This then allows us to get back to: `browser.original_window.use`

I have code for allowing switch_window to accept parameters, but it would essentially be the same thing as #use, so I'm on the fence as to whether it should be included.
```
browser.switch_window(title_or_url_opt)
# vs
browser.window(title_or_url_opt).use
```

### DIFF
--- a/lib/watir/has_window.rb
+++ b/lib/watir/has_window.rb
@@ -51,6 +51,26 @@ module Watir
       @original_window ||= window
     end
 
+    #
+    # Waits for and returns second window if present
+    # See Window#use
+    #
+    # @example
+    #   browser.switch_window
+    #
+    # @return [Window]
+    #
+
+    def switch_window
+      current_window = window
+      wins = windows
+      wait_until { (wins = windows) && wins.size > 1 } if wins.size == 1
+      raise StandardError, 'Unable to determine which window to switch to' if wins.size > 2
+
+      wins.find { |w| w != current_window }.use
+      window
+    end
+
     private
 
     def filter_windows(selector, windows)
--- a/spec/watirspec/html/window_switching.html
+++ b/spec/watirspec/html/window_switching.html
@@ -2,11 +2,21 @@
 <html>
   <head>
     <title>window switching</title>
+    <script>
+        function windowOpenDelayed(timeout) {
+            setTimeout(function() {
+                window.open("closeable.html");
+            }, timeout);
+        }
+    </script>
   </head>
 
   <body>
     <p>
       Click <a id="open" href="#" onclick='window.open("closeable.html")'>here</a> to open a new window.
+    </p>
+    <p>
+      Click <a id="delayed" href="#" onclick=windowOpenDelayed(1000)>here</a> to open a new window after a delay.
     </p>
   </body>
 </html>
--- a/spec/watirspec/window_switching_spec.rb
+++ b/spec/watirspec/window_switching_spec.rb
@@ -92,6 +92,40 @@ describe 'Browser' do
       expect { browser.window(handle: 'bar').use }.to raise_no_matching_window_exception
     end
   end
+
+  describe '#switch_window' do
+    it 'switches to second window' do
+      original_window = browser.window
+      browser.switch_window
+      new_window = browser.window
+
+      expect(original_window).to_not eq new_window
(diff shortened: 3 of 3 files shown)


titusfortner:
I guess it is also worth noting that this also works as expected, and returns the user to the original window:

```ruby
do_something_to_open_window
browser.switch_window
do_things_in_new_window
browser.switch_window
do_things_in_original_window
```

jkotests:
Looking at the comments in #755, my thought is still:
* `#switch_window` is a useful for most cases.
* We should still support :index.

Would it be hard for us to track the order of window handles? My thought would be, every time we call window handles:
* Append any new handles to our cache (ie new windows)
* Remove any handles that are no longer returned (ie closed windows)
* Return the windows handles sorted by the cache.

The only problem would be in the case where someone opened multiple windows in between calls for window handles. 

Maybe that's too much work?

titusfortner:
I have actually reproduced a situation when using index does not do what a user would expect. We can't maintain a feature that only intermittently works, and when it doesn't work it doesn't throw an exception. This puts us in the position of allowing Type 2 errors, which we need to be much more concerned about preventing. 

I'd love to be able to track all open windows, but we have no way of knowing if something opened a new window. We'd either have to make the assumption that a user always immediately switches to the new window, or constantly call the window handles endpoint, neither of which I think we want to do.

jkotests:
I suppose. Just seems quite unfortunate to those that have been using index so long without issue.

Do we need to deprecate `#to_a` or allow that as a dirty hack - ie `browser.windows.to_a[-1]`?

titusfortner:
Well, we don't need to deprecate `#to_a` since it can't be called right now. I'm on the fence about allowing it as a dirty workaround in Watir 7.

Another idea would be to create an optional extension that injects a window handles call as an AfterHook, and stores the results, and allows indexing Watir Collections. One advantage to this would be to allow users to toggle support on and off only for the tests that needed to handle more than 2 windows...

kapoorlakshya:
Hey @titusfortner, I like the `browser.switch_window(title_or_url_opt)` approach which allows us to pass the `title` or `url` locator. This will allow us to interact with more than just 2 windows in those rare cases. We have a few test cases that use 3 windows, and since `index` is being pulled out for good reasons we still need a way to reach the third window.

For example:

1. Search user
2. Click user to open a user specific page in a new window
3. Click a help icon on window 2, which opens window 3
4. Verify the correct Help page is opened in window 3
5. Close window 3
6. Switch back to window 2

Will this still be possible with your change?

jkotests:
@kapoorlakshya , switching to a window by title or url would continue to work as it does today:

~~~~~~~~
browser.window(title: 'your_title')
browser.window(url: 'the/url')
~~~~~~~~

jkotests:
@titusfortner , maybe you're right and we should just remove :index. Presumably there should always be something that the person can use to uniquely identify each window. Maybe not the title or url, but presumably at least the window content - eg `browser.windows.find { |w| w.text =~ /something unique/ }`. If people do this a lot, we would expand the supported locators - eg `browser.window(text: /something unique/)`

jkotests (approved):

kapoorlakshya:
Thanks, @jkotests. And I would find this useful - `browser.window(text: /something unique/)`.

titusfortner:
I'm pulling the deprecation code out of this... This PR is *just adding the `#switch_window` method.
The deprecations got more complicated, so I'll implement those elsewhere and reference the conversation here.

titusfortner:
fyi @kapoorlakshya & @jkotests instead of `:text` locator, I'm proposing `:element` locator, as it is much more flexible. 
```ruby
browser.window(element: browser.element(text: /something unique/))
```
implementation in #913

PR #850, opened 2019-01-14
select.rb

Will-Rowan:
I use this function quite often for clearing a precise amount of text as opposed to clearing all text from a text field. If any reviewers have a similar and better function that does the same thing, please feel free add it in place of this and in a possibly a more suitable location.

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -9,6 +9,19 @@ module Watir
 
       selected_options.each(&:click)
     end
+    
+    #
+    # Remove an exact amount of text from a highlighted area.
+    # 
+    # @example
+    #   backspace(4)
+    # 
+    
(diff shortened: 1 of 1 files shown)


titusfortner (changes requested):
Thanks for the submission, Will! We request that people write tests for new code like this, both because it makes it obvious what the code does and doesn't do, and makes it easier to verify it is in the correct place. As it stands, your PR says text fields, the comments say highlighted area, and the code is in the class that handles Select lists. I think what you actually want is for this to be in the user editable module. If you'd like help writing the specs, feel free to reach out to us on slack.

jkotests:
@Will-Rowan, I am going to close this due to inactivity. If you would still like to see this added, please re-submit with specs. Thanks.

PR #856, opened 2019-02-09
Respect/ignore case on certain attributes defined by HTML spec

### LINKED ISSUE #507: Case sensivity for attribute selectors
There is a difference between CSS and XPath in case sensivity of attributes.

Having these elements on the page:

```html
<form method="post" id="new_user" />
<form method="POST" id="delete_user" />
```

you can query them in browser:

```
> document.querySelectorAll('form[method="post"]')
< NodeList [<form id="new_user">, <form id="delete_user">] (2)

> document.querySelectorAll('form[method="POST"]')
< NodeList [<form id="new_user">, <form id="delete_user">] (2)

> $x('.//form[@method="post"]')
< [<form id="new_user">] (1)

> $x('.//form[@method="POST"]')
< [<form id="delete_user">] (1)
```

This comes from the fact that XPath considers attributes case sensitive while CSS does not.

Since now Watir only uses XPath selectors, using `method: 'post'` selector would return only one element. But when Watir is used with watir_css or watizzle gems, same selector would return two elements.

Also compare plain Selenium:

```ruby
driver.find_elements(xpath: './/form[@method="post"]').count # returns 1
driver.find_elements(css: 'form[method="post"]').count       # returns 2
```

Since two behaviours cannot co-exist, I'm not sure what to do about it.

Related links:

* [CSS3 Selectors Case Sensiviity](https://www.w3.org/TR/css3-selectors/#casesens)

p0deje:
Per https://www.w3.org/TR/html52/single-page.html#casesensitivity
Closes #507


lib/watir/locators/element/selector_builder/xpath.rb
```diff
@@ -13,6 +13,9 @@ class XPath
           def build(selector)
             @selector = selector
 
+            @valid_attributes = Watir::HTMLElement.attribute_list
+            @valid_attributes += Watir.tag_to_class[@selector[:tag_name].to_sym]&.attribute_list if @selector[:tag_name].is_a?(String)
+
```
titusfortner, line 16: I'm not sure this is the right code, but the logic we want is:
```
@valid_attributes = if @selector[:tag_name].is_a?(Regexp)
                      Watir::HTMLElement.attribute_list
                    else
                      Watir.tag_to_class[@selector[:tag_name]&.to_sym]&.attribute_list ||
                          Watir::HTMLElement.attribute_list
                    end
```

lib/watir/elements/element.rb
```diff
@@ -18,6 +18,17 @@ class Element
     attr_accessor :keyword
     attr_reader :selector
 
+    # https://www.w3.org/TR/html52/single-page.html#casesensitivity
+    CASE_INSENSITIVE_ATTRIBUTES = %i[accept accept_charset align alink axis
```
titusfortner, line 22: pedantic, but should we create a separate `class HTMLElement` for this code, since that's the class everything else is in?

titusfortner (changes requested):
@p0deje yes, this is structured the way I would expect

The one piece of weirdness is that this isn't working for dashes like with `@http-equiv`
This works:
```
".//*[local-name()='meta'][translate(@http-equiv,'-AB','-ab')='Content-Type']"
```
But this does not:
```
".//*[local-name()='meta'][translate(@http-equiv,'-ABC','-abc')='Content-Type']"
```
I'm assuming it's a bug somewhere. @twalpole any ideas?

Otherwise maybe we should just remove `http-equiv` from the list of things that get translated?

lib/watir/locators/element/selector_builder/xpath.rb
titusfortner: (`+=` results in a bunch of duplicates, and sometimes `@valid_attributes` is nil with code as is`)

twalpole:
> @p0deje yes, this is structured the way I would expect
> 
> The one piece of weirdness is that this isn't working for dashes like with `@http-equiv`
> This works:
> 
> ```
> ".//*[local-name()='meta'][translate(@http-equiv,'-AB','-ab')='Content-Type']"
> ```
> But this does not:
> 
> ```
> ".//*[local-name()='meta'][translate(@http-equiv,'-ABC','-abc')='Content-Type']"
> ```
> I'm assuming it's a bug somewhere. @twalpole any ideas?
> 
> Otherwise maybe we should just remove `http-equiv` from the list of things that get translated?

I would expect the second one to fail -- it's translating 'Content-Type' to 'content-Type' and then checking if it's equal to 'Content-Type'.  For this to work the value after the `=` also needs to be downcast

lib/watir/locators/element/selector_builder/xpath.rb
p0deje: My assumption is that we want it to work for all attributes defined on an element, meaning attributes defined on the element and all its parent (`HTMLElement` in our case). Hence, `+=`. There should be no duplicates at all (at least according to `html_elements.rb`) but I can use set instead of an array if you want.

Thanks for tag name using Regexp case, I've totally forgotten about it.

titusfortner: The attribute list is inherited. I verified with a debug point.

You had Regexp effectively covered when checking for String, I just went the other way (not Regexp)

lib/watir/elements/element.rb
p0deje: I wanted originally to put it into `html_elements`, but then I've thought of `svg_elements` that might also benefit from it. On the other side, I haven't found a place in SVG spec which talks about case-insensitive attributes at all. Not sure if we should do the same translations for SVG elements at all.

titusfortner: Hmm, yeah, I guess we'd at least have to define `CASE_INSENSITIVE_ATTRIBUTES` for `class SVGElement` if we move it. I could go either way, though I'm also trying to keep the size of `element.rb` to a minimum. :) Either way, though.

p0deje: Let's keep like that for now.

lib/watir/locators/element/selector_builder/xpath.rb
p0deje: You are right. I've pushed a new version, please check it out.

p0deje:
> I would expect the second one to fail -- it's translating 'Content-Type' to 'content-Type' and then checking if it's equal to 'Content-Type'. For this to work the value after the `=` also needs to be downcast

I've pushed a newer version which downcases both left-hand side and right-hand side, making sure cases with `meta` work correctly.

p0deje:
Seems like CI is green, so @titusfortner please have another look when you have time.

titusfortner (approved):

titusfortner:
Nice, I didn't even screw up the log with that. I resolved conflicts via github then squashed & merged all tidy-like.

PR #857, opened 2019-02-12
add Waitable to Collections

### LINKED ISSUE #854: How to deal with waiting when fetching multiple elements?
## Meta -
Watir Version: `6.16.5`
Selenium Version: `3.141.0`
Browser Version: Chrome `72.0.3626.96`
Browser Driver Version: ChromeDriver `2.45.615291`
OS Version: Windows 7 Enterprise SP1 (`7601`)

## Behavior -
I'm dealing with a website that uses [Apache Wicket](https://wicket.apache.org/) to load a table with an AJAX request after the user performs a search. Since I'd like to loop through each row of this table I presumed I'd be able to use something like this:
```ruby
@browser.trs(class: 'example').each do |tr|
  puts tr.html
end
```
However, it seems the waiting behaviour varies between selecting one element and more than one element as I don't get any output with the above code. Whereas, when I select the first row I do get its HTML:
```ruby
puts @browser.tr(class: 'example').html
```
Is there a preferred way to deal with this?

## Steps to reproduce -

The page appears on a licenced IP telephony portal, I wouldn't know where to start creating an example page that demonstrates this.

titusfortner:
This fixes #854


jkotests:
When I tried this the other day, the solution was quite limited and/or confusing.

Unlike individual elements, element collections do not get re-located. Each iteration of the wait loop will keep looking at the exact same elements. This means that:
* The wait can work if all of the elements you want to inspect already exist.
* The wait won't work if you are waiting for additional elements to be added or certain elements to be removed.

To wait for changes in the collection, I had to manually reset the collection (ie set `@to_a` back to nil). For example, to wait for a collection of elements to be created:

~~~~~~~~
browser.elements.wait_until { |c| c.reset!; c.count > 0) }
~~~~~~~~

Without the `reset!`, the loop would always timeout - ie it kept inspecting the original collection results, which started empty.

At a minimum, I'd say we would need to add `reset!`. Ideally, we'd find something that didn't need `reset!` as it'll be awkward to explain.

titusfortner:
@jkotests what would you think if we did:
```
    def each(&blk)
      @to_a = nil  # aka `#reset!`
      to_a.each(&blk)
    end
```
That would re-locate the collection for enumerable methods & aliases:
`#length`, `#count`, `#size`, `#empty?`, `#none?`

jkotests:
@titusfortner I was originally concerned that resetting the collection each time could lead to confusion. However, I didn't find a scenario that it would be unreasonable. So I think re-locating the collection each time is reasonable.

titusfortner:
Yeah, it wouldn't affect the methods that rely on `#[]` and the enumerable methods seem like they should only need to get called once when using them normally. I'll probably also define `#exist?` as `!empty?`

titusfortner:
I had to update the wait specs before adding tests for this.

This does the reset as discussed & adds the aliases for `#exist?` and `#exists?`

I included tests for your example above, let me know if I need any others

jkotests (approved):

woacademy:
@titusfortner thanks for getting this in, can finally start using it 🙇‍♂️

Issue #855, opened 2019-02-08
Can't upload file in Firefox 


peacebytes:
## Meta -
Watir Version: 6.16.5
Selenium Version: 3.141.0
Browser Version: Firefox 61 (64-bit)
Browser Driver Version: geckodriver-v0.24.0-linux64.
OS Version: Ubuntu 18.04

## Expected Behavior -
Firefox should allow to upload local files with absolute path
## Actual Behavior -
Error occured:
```
element located, but timed out after 30 seconds, waiting for #<Watir::FileField: located: true; 
{:tag_name=>"input", :type=>"file"}> to be present (Watir::Exception::UnknownObjectException)
```
## Steps to reproduce -
Trying to upload with code
```
file_name = "#{File.expand_path('.')}/inventory/uploadtest/photo1.jpg"
Watir.relaxed_locate = false
@browser.file_field.send_keys file_name
Watir.relaxed_locate = true
```
Stack trace
```
["/var/lib/gems/2.5.0/gems/watir-6.16.5/lib/watir/elements/element.rb:752:in `raise_present'", 
"/var/lib/gems/2.5.0/gems/watir-6.16.5/lib/watir/elements/element.rb:803:in `rescue in element_call'", 
"/var/lib/gems/2.5.0/gems/watir-6.16.5/lib/watir/elements/element.rb:787:in `element_call'", 
"/var/lib/gems/2.5.0/gems/watir-6.16.5/lib/watir/elements/element.rb:338:in `send_keys'", 
"/home/{personal_folder}/features/step_definitions/uploadpopup_steps.rb:45:in `block in <top 
(required)>'", "/var/lib/gems/2.5.0/gems/cucumber-3.1.2/lib/cucumber/glue/invoke_in_world.rb:39:in 
`instance_exec'", "/var/lib/gems/2.5.0/gems/cucumber-3.1.2/lib/cucumber/glue/invoke_in_world.rb:39:in 
`block in cucumber_instance_exec_in'", "/var/lib/gems/2.5.0/gems/cucumber-
3.1.2/lib/cucumber/glue/invoke_in_world.rb:54:in `cucumber_run_with_backtrace_filtering'", 
"/var/lib/gems/2.5.0/gems/cucumber-3.1.2/lib/cucumber/glue/invoke_in_world.rb:27:in 
`cucumber_instance_exec_in'", "/var/lib/gems/2.5.0/gems/cucumber-
....
```

Tried with `set file_name`
```
Watir.relaxed_locate = false
[...]

PR #858, opened 2019-02-12
implement FileField#upload as alias of #set

titusfortner:
So, `__FILE__` is giving me a full path, is there a reason we had `File.expand_path` before? I'm not even sure `#expand_path` is actually correct for what we wanted, anyway?

Also, I very much like the idea of `browser.file_field.upload(file)` (instead of just `#set`)

Clarity inspired by #855

### DIFF
--- a/lib/watir/elements/file_field.rb
+++ b/lib/watir/elements/file_field.rb
@@ -12,6 +12,7 @@ module Watir
 
       self.value = path
     end
+    alias upload set
 
     #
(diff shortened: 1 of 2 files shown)


p0deje (approved):

jkotests (approved):

titusfortner:
merged: https://github.com/watir/watir/commit/516b6a5935defe74aed316a8c0986c644b14060f

PR #861, opened 2019-03-13
Element#right_click now accepts modifiers.

kapoorlakshya:
This fulfills a requirement that I have to test certain "special fields" that display debug info when a right click with `:control` is performed.

**Example**
`browser.element(id: 'my-el').right_click(:shift, :alt)`

@titusfortner and I had a discussion on this a while ago, and we had agreed that since `#click` accepts modifiers, `#right_click` could as well.

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -196,15 +196,34 @@ module Watir
     end
 
     #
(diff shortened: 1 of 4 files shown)


jkotests:
Thanks @kapoorlakshya

PR #890, opened 2020-02-14
Fix failing unit tests

jkotests:
WIP: The failing tests are not reproducible locally. Using this branch to try to debug the issues.

Notes:
* There is a bug in Chromedriver with the input of contenteditable elements (https://crbug.com/chromedriver/3214). As none of the tests appear to require this attribute on the text field, removing it.
* Firefox no longer executes custom JavaScript during a shift + right-click, which means the right-click logger no longer receives the event. Switched the page to use a different modifier. See https://bugzilla.mozilla.org/show_bug.cgi?id=1322967


titusfortner:
Oh shoot. I should have looked at this first. I see that we both made some of the same decisions on fixing things. Only difference is the content-editable which appears to be fixed since I'm not seeing a problem currently.

PR #896, opened 2020-08-26
Fix right click

titusfortner:
Fixing tests related to context clicking

### DIFF
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -214,8 +214,9 @@ module Watir
         action = driver.action
         if modifiers.any?
           modifiers.each { |mod| action.key_down mod }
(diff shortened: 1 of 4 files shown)


lib/watir/elements/element.rb
```diff
@@ -214,8 +214,9 @@ def right_click(*modifiers)
         action = driver.action
         if modifiers.any?
           modifiers.each { |mod| action.key_down mod }
-          action.context_click(@element).perform
+          action.context_click(@element)
           modifiers.each { |mod| action.key_up mod }
+          action.perform
```
kapoorlakshya, line 219: Thanks for fixing this!

titusfortner:
Well, Travis is having issues, but I've verified this is working locally, so merging it.

PR #902, opened 2020-12-02
Less strict version check for `regexp_parser` gem

ShockwaveNN:
Current version of rubocop require `regexp_parser` >= 2.0
https://github.com/rubocop-hq/rubocop/blob/master/rubocop.gemspec#L36
Couldn't user both latest watir gem and latest rubocop

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+### Unreleased
+
+* Less strict version check for `regexp_parser` gem
+
 ### 6.17.0 (2020-08-27)
 * Require Ruby > 2.5
 * Implement Logger#selenium= to set selenium level from Watir
--- a/watir.gemspec
+++ b/watir.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.add_dependency 'selenium-webdriver', '~> 3.6'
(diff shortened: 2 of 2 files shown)


ShockwaveNN:
AppVeyor failures seems be the same on master and not related to my change

Waiting for travis status

ShockwaveNN:
Travis failures looks similar with current master, so I think this merge didn't break something

ShockwaveNN:
@titusfortner Could you take a look at this? Is this project alive?

Currently `rubocop` will allow usage of older version of `regex_parser`, but in general I think it's a good idead to make version check less strict

titusfortner:
Looks like latest version of Rubocop already relaxed the requirement so it isn't an urgent issue?
My non-work open-source time is getting Selenium 4 beta released; Watir will get a major update for Selenium 4.

ShockwaveNN:
@titusfortner Yes, rubocop is allow usage of `regexp_parser` v1

But what I should do if I want to use regexp_parser v2 and latest watir 6.17 in same project?

If i specify regexp_parser v2 - bundler will resolve watir to 6.14.0 - latest version that not containing this commit https://github.com/watir/watir/commit/06db65d40b24980e6f43133d1398dc7cb3834105

ShockwaveNN:
@titusfortner Thanks!

PR #906, opened 2021-02-19
Watir 7 Beta 2

titusfortner:
This is based on #919 now

It further removes all deprecated code and tidies up things that are no longer necessary with Selenium 4


titusfortner:
accidental branch deletion based on bad tracking branch. :-/

PR #910, opened 2021-02-22
Improve selects

titusfortner:
Edit: I guess we need to figure out disposition of #846 before moving forward with this. It might make sense to update that and deprecate `#select_all` and multiple arguments in `#select` and release another 6.x version. It would make the code much more simple for sure.

-------------
The build shortcut is what allows element to bypass locating query_scope element if a single XPath can be build for the child element. Current `#ensure_context` method will always locate `@query_scope` because `!@query_scope.located?` will always be true. Elements that need to locate the scope, now have a `:scope` key/value in `@build` so it is only necessary to ensure context in this situation if the query scope is an `IFrame`.

As for the Select List optimization, this builds an XPath that combines the 3 approaches value, text & label. We can wait for the element to be present rather than constantly looping through all the options and then exiting as necessary.

I ran the numbers for the pre-PR `select_list_spec.rb` running just the `#select` and `#select_all` tests:

| Command  | Before PR | After PR | Improvement |
| ------------- | -------- | -------- | -------- |
| get_element_css_value  | 96  | 79 | 17 |
| find_element  | 54  | 27 | -18 |
| find_elements  | 99  | 7 | 92 |
| #select multiple  | 34  | 10 | 24 |
| #select_all multiple  | 21  | 11 | 10 |
| #select_all label  | 28  | 33 | -5 |
| #select_all value  | 0  | 7 | -7 |

* The additional find_element calls are essentially replacing the css_value calls which check for staleness
* reducing `#find_elements` calls is the huge difference here; primarily in #select method because we aren't concerned about having to deal with multiple lists any more
* `#select` similarly no longer requires checking for `multiple` value unless an array is passed in (likely rare)
* `#select_all` has only half of the previous `multiple` calls because this PR caches it.
* `#select_all` has more label and value attribute calls, and I'm not certain why, but it seems like a solid tradeoff without digging even more into it.

### DIFF
--- a/lib/watir/element_collection.rb
+++ b/lib/watir/element_collection.rb
@@ -182,7 +182,11 @@ module Watir
     end
 
     def ensure_context
-      @query_scope.locate if @query_scope.is_a?(Browser) || @query_scope.located? && @query_scope.stale?
+      if @query_scope.is_a?(Browser) || !@query_scope.located? && @query_scope.is_a?(IFrame)
+        @query_scope.browser.locate
+      elsif @query_scope.located? && @query_scope.stale?
+        @query_scope.locate
+      end
       @query_scope.switch_to! if @query_scope.is_a?(IFrame)
     end
 
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -704,7 +704,9 @@ module Watir
     end
 
     def ensure_context
-      if @query_scope.is_a?(Browser) || !@query_scope.located? || @query_scope.located? && @query_scope.stale?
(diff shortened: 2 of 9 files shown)


titusfortner:
I guess this is a more moderate alternative to #846 
This PR still uses `#select_all` to differentiate between single and multiple select lists.

One of the issues, though is that the current implementation of `#select` is that it supports passing in an Array to indicate that all matching values in the select list should be selected as well as multiple arguments. 

I think that supporting both Arrays in `#select` as well as `#select_all` is unwieldy, and one or the other should be deprecated. I'm leaning toward deprecating `#select_all` like in #846 because there is something elegant about allowing everything to go through a single method (makes automatic form filling easier). I don't really want to release another 6.x, but maybe it makes sense in this case?

PR #908, opened 2021-02-22
Fix Ruby Debug Mode Warnings

titusfortner:
Edit: Moved this into draft status because I don't want to remove the methods in the first commit. At the very least they need to be implemented with `WebElement#property` instead of `WebElement#attribute`.

Based on #906 
Fixes #898 

`Option#selected?` and `Select#value` are essentially re-defining a call to an existing attribute method. We either keep the redefinition for better documentation purposes, or we get the verbose mode warning.

The YAML warning can't be fixed until we no longer support ruby 2.5

I aliased a few `#exist?` to `#exists?` so that the RSpec matcher wouldn't complain, and I don't see a downside to it.

PR #911, opened 2021-02-24
Fix Ruby warnings

titusfortner:
This supersedes #908 

* Ruby complained about our RSpec matcher for `#exists?`, I fixed it by having everything alias to `#exist?` as well
* Attribute list doesn't need to keep adding the same methods in `#attribute`, so there's a guard for it now.
* Defining instance variables in constructors where possible, guarding with `#method_defined?` conditionals when not
* `Select#value` is doing extra work, when the simple default attribute call always returns the same value (added spec to demonstrate this)
* Pulled `:selected` attribute from getting added to `html_elements.rb` so that the webdriver endpoint can handle determining `#selected?` instead of the attribute call (same behavior, just no Ruby warning about it)
* Can delegate `Watir::Logger#level=` by not supporting 2.3 or below

There are still some warnings in 2.7, but they can't be fixed until we no longer support 2.5 (which I think will likely be the case for Watir 7)

### DIFF
--- a/.github/actions/enable-safari/action.yml
+++ b/.github/actions/enable-safari/action.yml
@@ -1,0 +1,11 @@
+name: 'Enable Safari Driver'
+description: 'Set safaridriver to run in automation mode'
+runs:
+  using: composite
+  steps:
+    - run: |
+        defaults write com.apple.Safari IncludeDevelopMenu YES
+        defaults write com.apple.Safari AllowRemoteAutomation 1
+        sudo safaridriver --enable
+        safaridriver -p 0 &
+      shell: bash
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -36,3 +36,20 @@ jobs:
       - run: bundle install
       - run: |
           bundle exec rake spec:${{ matrix.task }}
+
+  safari-test:
+    name: Safari Test
+    runs-on: macos-latest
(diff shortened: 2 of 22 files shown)


lib/watir/elements/select.rb
```diff
@@ -106,18 +106,6 @@ def selected?(str_or_rx)
...
-    # Returns the value of the first selected option in the select list.
-    # Returns nil if no option is selected.
-    #
-    # @return [String, nil]
-    #
-
-    def value
-      option = selected_options.first
-      option&.value
-    end
```
jkotests: Are you fine that this removal would technically be a change in behaviour for multi-selects? A multi-select with nothing selected would return `nil` in 6.17 but will now return `""` (empty string) for 6.18.

The change seems reasonable in the long term. It's just whether we should go through the whole deprecation warning for this.

titusfortner: erf, no, I don't want to do this, actually. I eventually want to implement this with `WebDriver::Element#property` which will return null instead of empty string. I don't want to go down that path, though until we also get `WebDriver::Element#dom_attribute` which won't be until Selenium 4 beta 2, so I figured I'd wait on all of it until Watir 7. I'm pulling the last commit and keeping this method. I added a spec for this as well.

jkotests (approved):

PR #912, opened 2021-02-24
deprecate #select_all in favor of using an Array with #select

titusfortner:
The plan going forward is to treat everything submitted to `Select#select` that is Array-like (multiple args count), to be treated as a `#select_all` that applies only to multiple selects.

There are a bunch of things to optimize still (ideas from #846 & #910), and this should provide the groundwork to do that. I think this is all I want to do in 6.x and then work on the rest with a "clean" slate in 7.x.

### DIFF
--- a/lib/watir/elements/select.rb
+++ b/lib/watir/elements/select.rb
@@ -30,8 +30,11 @@ module Watir
     #
 
     def select(*str_or_rx)
-      results = str_or_rx.flatten.map { |v| select_by v }
-      results.first
+      if str_or_rx.size > 1 || str_or_rx.first.is_a?(Array)
+        str_or_rx.flatten.map { |v| select_all_by v }.first
+      else
+        str_or_rx.flatten.map { |v| select_by v }.first
+      end
     end
 
     #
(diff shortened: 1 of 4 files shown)


lib/watir/elements/select.rb
```diff
@@ -55,8 +62,11 @@ def select_all(*str_or_rx)
     #
 
     def select!(*str_or_rx)
-      results = str_or_rx.flatten.map { |v| select_by!(v, :single) }
-      results.first
+      if str_or_rx.size > 1 || str_or_rx.first.is_a?(Array)
+        str_or_rx.flatten.map { |v| select_by! v, :multiple }.first
+      else
+        str_or_rx.flatten.map { |v| select_by! v, :first }.first
```
jkotests: Is this supposed to be changing the second argument from `:single` to `:first`?

If I am looking at the related JS snippets correctly, they are only stopping the selection if the second argument is "single". So with this change `#select!` will always be selecting all instead of just the first.

jkotests (changes requested):

lib/watir/elements/select.rb
titusfortner: Ah, I actually couldn't find where that was  being used to know what the right keyword was to put there. Good catch.

jkotests (approved):

PR #913, opened 2021-02-25
Update Window Locators

titusfortner:
The deprecations are based on the conversation from #849 

The reason for `WindowCollection` is to be able to enforce the deprecation of indexing.

But... Once we have it, we can do some powerful things with it, starting with including `Waitable` similar to how we just did in #857 
Implementing Waitable meant making it relocatable, which is a much better implementation than a simple store of windows.

I have some more ideas for Windows, but this is the next step to make sure on.


titusfortner:
I just added an element locator to Window because it was so much easier than I thought it would be, and it's a nice replacement if we're taking away index.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    alias length count
+    alias size count
+    alias empty? none?
+
+    #
+    # First window of the collection
+    #
+    # @return [Watir::Window] Returns an instance of a Watir::Element subclass
```
jkotests: Should be `Watir::Window` instead of `Watir::Element`.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,98 @@
...
+    #
+
+    def first
+      self[0]
+    end
+
+    #
+    # Last element of the collection
+    #
+    # @return [Watir::Element] Returns an instance of a Watir::Element subclass
```
jkotests: Both lines should be referring to windows instead of element.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,98 @@
...
+    #
+    # Get the element at the given index or range.
+    #
+    # Any call to an ElementCollection that includes an adjacent selector
+    # can not be lazy loaded because it must store the correct type
+    #
+    # Ranges can not be lazy loaded
+    #
+    # @param [Integer, Range] value Index (0-based) or Range of desired element(s)
+    # @return [Watir::Element, Watir::ElementCollection] Returns an instance of a Watir::Element subclass
```
jkotests: Needs to be updated for windows.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,98 @@
+module Watir
+  class WindowCollection
+    include Enumerable
+    include Waitable
+
+    def initialize(browser, selector = {})
+      unless selector.keys.all? { |k| %i[title url].include? k }
```
jkotests: Would it be hard to support `:element` in WindowCollection? It seems useful for Window, so would be nice here too.

jkotests (changes requested):

lib/watir/window_collection.rb
titusfortner: Oh not at all, this was just an oversight; and an indication that I don't have the right tests, either. :/
I squashed in some tests for `ElementCollection#new` and added element to this list.

titusfortner:
It's like I just copy/pasted that other class over and didn't even look at the documentation commends. :)
Should all be fixed now, thanks.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,99 @@
...
+      self[-1]
+    end
+
+    #
+    # Get the element at the given index or range.
+    #
+    # @note windows in a collection are not ordered so this is not reliably
+    # @deprecated use Browser#switch_window or a better Window locator
+    # @param [Integer, Range] value Index (0-based) or Range of desired element(s)
+    # @return [Watir::Element, Watir::ElementCollection] Returns an instance of a Watir::Element subclass
```
jkotests: This documentation still needs an update.

lib/watir/window_collection.rb
titusfortner: and... adding that test showed that my implementation was broken, so... yay tests.

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,99 @@
...
+    alias eql? ==
+
+    def to_a
+      @to_a ||= begin
+        all = @browser.driver.window_handles.map { |wh| Window.new(@browser, handle: wh) }
+        if @selector.empty?
+          all
+        else
+          all.select do |win|
+            @selector.all? { |key, value| win.send(key) =~ /#{value}/ }
```
jkotests: As per the spec failure, this doesn't know how to handle the `:element` selector.

jkotests (changes requested):

lib/watir/window_collection.rb
titusfortner: Yeah, took me a couple tries and several more specs to make sure I had this right. :)

lib/watir/window_collection.rb
```diff
@@ -0,0 +1,110 @@
...
+    # @deprecated use Browser#switch_window or a better Window locator
+    # @return [Watir::Window] Returns an instance of a Watir::Window
+    #
+
+    def last
+      self[-1]
+    end
+
+    #
+    # Get the element at the given index or range.
```
jkotests: Not to be picky... but still referencing "element" here.

jkotests (approved):

lib/watir/window_collection.rb
titusfortner: gah. thanks. again.

PR #914, opened 2021-02-25
Window keywords

titusfortner:
This is based on #913 

Since there are only a handful of valid selectors for Windows, they should be specified directly.
We're removing support for ordered pairs for elements, so we should do the same for windows.

I'm marking this as draft because there are a few sloppy things left, and it is based on another PR still, but I want to get buy in on the idea before I put more time into it.


jkotests:
I haven't used keywords much, so sorry if this is a dumb question. What's the benefit of using keywords for this specific scenario?

I think my only objection to keywords here is that it would be inconsistent with elements. For elements, I can specify my selector as a Hash and pass it in:

~~~~~~~~
hsh = {text: 'a', class: 'b'}
browser.element(hsh).exists?
 ~~~~~~~~

Doing the same for windows will require the double-splat:

~~~~~~~~
hsh = {url: 'a', title: 'b'}
browser.window(**hsh).exists?
 ~~~~~~~~

A very minor inconsistency, but it's always better to be consistent unless there's some other gain.

titusfortner:
So, we're getting rid of `#extract_selector`, which means I need to remove allowing ordered pairs. Of note the deprecation notice here assumes elements not windows, so I need to fix that either way.

I think I went down this path because moving from`*args` to `selector = {}` is more complicated than moving to keywords. 
(I can toggle off of `args.empty?` in the latter, but have to do more analysis for the former since everything becomes an arg).

The main difference between Elements & Windows right now is that infinite locators are supported for the first, but only a handful for Windows...

Besides conversion convenience... The biggest advantages at end state is that you would get this code for free, and the implementation code would be more obvious
```ruby
unless selector.keys.all? { |k| %i[title url index element].include? k }
  raise ArgumentError, "invalid window selector: #{selector_string}"
end
```

Except... well, why don't we open it up more...
```ruby
def initialize(selector = {}, handle: nil)
  @handle = handle
  @handle ||= current_window if selector.empty?

  @title = selector.delete(:title)
  @url = selector.delete(:url)
  @element = selector.delete(:element)
  raise if @element && !selector.empty?
  @element ||= browser.element(selector)  
end
```

titusfortner:
Ok, we've agreed not to go the keyword route for this

PR #916, opened 2021-03-07
Baseline Capabilities Specs for Watir 6.18

titusfortner:
This code sets it up so I can show what Capabilities are doing now in v6.18, what the plan is for that functionality in 7.0 and what I intend to do about it for 6.19.

* 142 tests of what *should be working with Capabilities in 6.18
* 86 are passing in 6.18
* 14 don't work in 6.18 and won't be a part of Watir 7 API, so I'm not going to fix them for 6.19
* Some of these work in 6.18, but I'm going to add deprecations in Watir 6.19 and remove in 7.0
* Some of these work in 6.18, but with a different underlying implementation from how they will in 6.19 so have different assertions based on version

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -20,6 +20,7 @@ require 'watir/browser'
 require 'watir/screenshot'
 require 'watir/after_hooks'
 require 'watir/logger'
+require 'watir/version'
 
 module Watir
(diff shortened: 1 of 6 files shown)


titusfortner:
Closing this as a PR since it is subsumed by #917

PR #917, opened 2021-03-07
Watir 6.19 - Capability Deprecations

titusfortner:
This is the 6.19 implementation for the specs created in #916 (which is why this is a draft PR)

This is functionally completely backward compatible (some items are put in options instead of capabilities, etc, but the driver will do the same thing with it).

It implements the broken things that need to be supported & deprecates the working things that should no longer be supported.

The documentation will need updating again.


titusfortner:
Not sure why the Actions results aren't showing up here, but:
https://github.com/titusfortner/watir/actions?query=branch%3Adepr_caps

Of note, the Ruby 3.0 unit tests are failing, but Selenium 3 isn't Ruby 3 compatible, so neither is Watir 6 explicitly.

jkotests (approved):

titusfortner:
This was merged

PR #918, opened 2021-03-07
deprecate WindowCollection#to_a

titusfortner:
Further discouraging the use of indexing with window collections


titusfortner:
closing this as a separate PR as it is subsumed by #917

PR #919, opened 2021-03-08
Watir 7 Beta 1 - Remove deprecated caps 

titusfortner:
This moves from 6.x to 7.x by requiring Selenium 4 (Beta 2 which is currently unreleased).
It shows working code after all of the deprecated capabilities code has been cleared out.


titusfortner:
everything in this PR is in #906 , so I'll merge from that PR

Issue #405, opened 2016-02-16
Watir should provide a standard interface to HTML elements


abaird:
We need two standard methods that:
- interacts with the element
- gets the element's representation

For example, let's say there is an interaction method "set". Calling `set` would have the following effects:

```
link.set - clicks on the link
checkbox.set - checks the checkbox
radio.set(true) - clicks the radio button
```

Let's say that there is a method "value" that gets the element's representation. Calling `value` would have the following effects:

```
link.value - returns the html
checkbox.value - returns whether or not the checkbox is checked
radio.value - returns whether or not the radio is selected
```

A code example can be found [here](https://github.com/bret/watircraft/blob/master/lib/extensions/watir.rb). The point of this is to provide frameworks a single method to interact with an element instead of having to alias all of Watir's different methods for interacting with elements.

This suggestion was raised during an Austin Homebrew Testers meeting.

p0deje:
This seems strange to me. Why would `A#set` be aliased to `A#click`? The naming is strange here. Why would `A#value` return HTML and not `href` attribute?

More general question is why would you want to unify all the element methods?

bret:
@titusfortner showed us some code that looked something like this (https://github.com/bret/watircraft/blob/master/lib/extensions/watir.rb) (Titus can you post your actual code?) and I said that I thought it was wrong that you had to write wrappers like this for watir. Instead Watir should provide a consistent methods that allowed you to set or get the value of an element.

Suggestion to use "value" doesn't actually work out because their is a collision with existing usage.

I agree that link.set is silly.

This proposal needs to be worked out in more detail.

abaird:
Clearly I chose bad names for these methods. I have re-worded the original post. The names are still unwieldy, but hopefully less confusing now. Also, I meant href and not html.

This is just a _possible_ implementation. The root of the conversation though was the discussion around why [this code in Titus's watir_drops framework](https://github.com/titusfortner/watir_drops/blob/master/lib/watir_drops/page_object.rb#L45-L64) was something we felt like was code that Watir needed to handle and not have it be re-implemented in every page object framework.

titusfortner:
[This is the code](https://gist.github.com/titusfortner/ad873ff38bfda98f52f3ea6acb618601) I was showing earlier this year. Instead of having to do a conditional to interact with a default action, we create them in the applicable classes.

I don't necessarily think we need to do this for *all elements, just the ones that make sense to use with a variable setter.

p0deje:
[...]

PR #920, opened 2021-03-16
provide standardized methods to get and set values for specific elements

titusfortner:
What do we think about this as the solution to #405 ?
Don't need to wait on #846 since all desired behavior is now supported in 6.19

Summary of Behavior:
----------
Checkbox:
`#set` —> boolean argument for whether it should be selected; clicks if not matching
`#get` —> boolean return for if selected

Radio:
`#set` —> boolean argument for whether it should be selected; clicks if true & not already selected
`#get` —> boolean return for if selected

RadioSet:
`#set` —> String or Regexp argument; first button with matching value or label is selected
`#get` —> String return with the attribute value for the selected Radio button

Text & TextArea:
`#set` —> String or Regexp argument; Clears & Sends Keys (current behavior)
`#get` —> String return with the property value of the element

Select
`#set` —> String or Regexp or Array argument; selects first or all matching values/labels (current behavior of #select)
`#get` —> String return of the attribute value for the selected option

Element (if subtype not specified in the element definition)
`#set` -->
* Calls to_subtype and uses one of the above behaviors if appropriate
    * note if it is a Radio, it determines if the argument is a String/Regexp or a Boolean to determine if it uses Radio or RadioSet behavior 
* If "content-editable" attribute set, sets user editable and clear/send_keys (current behavior)
* Does not click element if provided argument is “false”
* Otherwise clicks element with provided modifiers

`#get` --> 
* Calls to_subtype and uses one of the above behaviors if appropriate
* Radio button always returns boolean (never RadioSet behavior)
* Otherwise returns text value (including for content-editable elements since `value` returns null, oddly)

### DIFF
--- a/lib/watir/elements/radio.rb
+++ b/lib/watir/elements/radio.rb
@@ -9,8 +9,8 @@ module Watir
     # Selects this radio button.
     #
 
-    def set
-      click unless set?
+    def set(bool = true)
+      click if bool && !set?
     end
     alias select set
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,17 +19,12 @@ Lint/UnifiedInteger:
 
 # Default: 17
 Metrics/AbcSize:
-  Max: 20
+  Max: 22
   Exclude:
-    - 'lib/watir/capabilities.rb'
     - 'lib/watir/locators/element/selector_builder.rb'
-    - 'lib/watir/locators/element/selector_builder/regexp_disassembler.rb'
-    - 'lib/watir/locators/element/selector_builder/xpath.rb'
-    - 'lib/watir/locators/element/locator.rb'
+    - 'lib/watir/locators/element/selector_builder/*.rb'
     - 'lib/watir/generator/base/generator.rb'
-    - 'lib/watir/generator/base/visitor.rb'
     - 'spec/locator_spec_helper.rb'
-    - 'spec/watirspec_helper.rb'
 
 Metrics/BlockLength:
   Exclude:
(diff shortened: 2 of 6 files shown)


lib/watir/elements/radio.rb
```diff
@@ -9,8 +9,8 @@ def build
     # Selects this radio button.
     #
 
-    def set
-      click unless set?
+    def set(bool = true)
+      click if bool && !set?
```
jkotests, line 13: Are people going to get confused why `radio.set(false)` doesn't clear the radio button? Is it worth throwing an error if `!bool && set?`?

jkotests (approved):

lib/watir/elements/radio.rb
titusfortner: I'm thinking of `#set` and `#get` as things that will primarily be used by Page Object implementations.
If you use `#populate_page_with` the `Hash` object should be able to dynamically set it to true/false. I think that's preferable to having to delete the key before passing it in.

Then again... I'd hope people would use RadioSet instead of Radio for these in general.

titusfortner: and "clearing a radio button" isn't something a user does with a page, so I don't think a user should reasonably expect that to be the behavior here.

titusfortner:
We agreed via chat that there is no compelling reason for a standardized `#get`. The implementation ideas I have for one only involve using `#text`. Also this didn't make it in Watir 6, so it'll get rebased into Watir 7.

titusfortner:
I pulled out the `#get` implementation and annotated in #405 why we are doing that.
I'm going to merge this as soon as I verify the tests are passing against current main branch

Issue #829, opened 2018-12-12
Wrap Selenium Actions Class


titusfortner:
Based on articles like this one: https://mark0203.blogspot.com/2018/12/ta-mobile-browser-its-all-about.html

Watir probably needs to better wrap actions class. We have a few specific methods for hover, drag & drop, right click, but we probably should provide a more comprehensive way for people to do everything that can be one in Selenium

titusfortner:
I'm just going to call this out of scope for the time being.
Selenium 4 no longer uses additional methods. Everything is in `#action` which means dropping down into selenium is probably the best bet for anything more complicated than what we provide by default.

PR #926, opened 2021-03-24
allow users to access Selenium Actions class directly from Browser in…

titusfortner:
addresses #829 

This is generally less of an issue now that everything has to go through actions class instead of mouse or keyboard methods, but this seems like a minor add to allow the same functionality.

### DIFF
--- a/lib/watir/browser.rb
+++ b/lib/watir/browser.rb
@@ -179,6 +179,16 @@ module Watir
     end
 
     #
+    # Allows using Selenium Actions Class like Selenium
+    #
+    # @return [Selenium::WebDriver::ActionBuilder]
+    #
+
+    def action
+      wd.action
+    end
+
+    #
     # Waits until readyState of document is complete.
     #
     # @example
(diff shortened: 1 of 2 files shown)


jkotests (commented):
I'm a bit concerned that this would cause confusion. Having `Browser#action` makes me think that I can do actions purely in Watir. However, I would still need to remember that the elements passed to the actions need to be `Selenium::WebDriver::Element`.

For example, I would expect that I would now be able to do:

~~~~~~~~
my_watir_element = browser.div
browser.actions.click(my_watir_element)
~~~~~~~~

Where as I would still need to manually converted the `Watir::Element` to `Selenium::WebDriver::Element`:

~~~~~~~~
my_watir_element = browser.div
browser.actions.click(my_watir_element.wd)
~~~~~~~~

Until we can support automatically converting the element, might be better to wait on this.

titusfortner:
Yup, I agree, I should have actually written a spec before adding this.

PR #943, opened 2021-08-10
Shadow dom support

jkotests:
@titusfortner , given:

- The request for shadow DOM support has come up a couple of times and
- We don't know when browsers will add support

How do you feel about having a hacky and limited workaround in the mean time?

This draft branch shows that we can provide at least some support in the short term. There are at least a few limitations:
- Only Chrome is supported by this approach.
- `#click` doesn't work for elements in the shadow DOM, but `#click!` does.
- The performance of complicated locators in a large shadow DOMs may be poor. `:xpath` locators cannot be applied to the shadow root, which means our XPath selector builder cannot be used. As a hack (rather than having to build a CSS selector builder) was to apply a simple `css: '*'` and force all of the locators through the more expensive filtering approach.
- To get things working quickly, the `ShadowRoot` inherits from `Element`. This means that the `ShadowRoot` has a lot of methods that don't work and/or make sense.

Again the idea here is simply to try to help some people from being blocked. If you're okay with taking this short-term approach, I can clean up this code and further investigate any other limitations.

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -99,5 +99,6 @@ require 'watir/elements/text_area'
 require 'watir/elements/text_field'
 require 'watir/elements/input'
 require 'watir/radio_set'
+require 'watir/shadow_root'
 
 require 'watir/aliases'
--- a/lib/watir/elements/element.rb
+++ b/lib/watir/elements/element.rb
@@ -491,6 +491,16 @@ module Watir
     end
 
     #
+    # Returns shadow root of element
+    #
+    # @return [Watir::ShadowRoot]
+    #
+
+    def shadow_root
+      ShadowRoot.new(self, {})
+    end
+
+    #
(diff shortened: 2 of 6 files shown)


titusfortner:
What if we implement this as `#shadow_root!` and when direct driver support comes out, we can implement the same API as `#shadow_root` and deprecate `#shadow_root!`?

I don't really know what this is doing on the JS side, but I think this is the correct API.

jkotests:
We may be able to skip this hack as Chromedriver support is in the works - https://chromium-review.googlesource.com/c/chromium/src/+/3033190 .

titusfortner:
Presumably it will be available in Chrome 95 or 96: https://wpt.fyi/results/webdriver/tests?label=experimental&label=master&aligned
If it gets delayed or it doesn't do what we need, we can revisit

PR #947, opened 2021-10-19
add tests for remote execution to github actions with Selenium 4.0.1

titusfortner:
PR to make sure all the remote tests are passing in all the platforms


titusfortner:
I'll rewrite if/when I do this.

PR #949, opened 2021-10-27
Can specify scroll position on hover

osadasami:


titusfortner:
This is great, thanks for the submission. 

Thinking through this, what do you think about doing this as a keyword? That way we can have the same API across all the action class wrappers, and we don't have to do as much work in `#right_click`.

```
double_click(scroll: :top)
right_click(:alt, :control, scroll: top)
```

It makes sense to take an automatic action when attempting to act on an element, I don't think it makes sense when you want to query an element's state. There are reasons you might want to see if an element is obscured without it automatically scrolling somewhere first.

I wonder if we should let people opt out of the automatic scroll so long as we're making it a parameter. (don't worry about this one, just musing).

osadasami:
> Thinking through this, what do you think about doing this as a keyword? That way we can have the same API across all the action class wrappers, and we don't have to do as much work in #right_click.

Using keyword looks much better, thanks. 

I have used `scroll_pos:` instead of `scroll:` becase `scroll:` conflicts with `scroll.to`

titusfortner:
Sorry for the delay, thanks for the submission! I'll get a release with this out soon.

PR #952, opened 2021-11-15
Update Scrolling Behavior

### LINKED ISSUE #948: Specify scroll position when using hover
When we use `el.hover`, Watir scrolls to the top of the element then hovers it.

But we don't always want to scroll to the top of the element when hovering.

Sometimes we need to scroll to the center or to the bottom of the elements when hovering.

it's impossible to do now.

If we use `el.scroll.to :center` and then `el.hover`, the scroll will go to the top of the element.

If we use `el.hover` and then `el.scroll.to :center` we will lose hover state.

I have found methods which use `scroll.to` inside:
- hover
- double_click
- right_click
- drag_and_drop_on
- drag_and_drop_by
- obscured?

So it would be great to specify scroll position in those methods.

Something like this:

```rb
def hover(scroll_position = :top)
  element_call(:wait_for_present) do
    scroll.to(scroll_position)
    driver.action.move_to(@element).perform
  end
end
```

titusfortner:
This completely fixes #948

* changes `browser.scroll.to :center` behavior
* Adds `Element#in_viewport?`
* Adds default scrolling to center for `#click` with modifiers
* Removes scroll position parameter from `#obscured?` (added in #949)
* Changes `#obscured?` to only scroll if element not in viewport, and then scrolls to bottom (correct behavior per w3c spec) instead of to the top


lib/watir/elements/element.rb
```diff
@@ -541,15 +542,25 @@ def present?
...
         return true unless present?
 
-        scroll.to(scroll_pos)
+        scroll.to :bottom unless in_viewport?
         execute_js(:elementObscured, self)
       end
     end
 
+    #
+    # Returns true if the element's center point is covered by a non-descendant element.
```
jkotests, line 555: Needs an updated description.

lib/watir/elements/element.rb
```diff
@@ -146,9 +146,10 @@ def tag_name
     # @param [:shift, :alt, :control, :command, :meta] modifiers to press while clicking.
     #
 
-    def click(*modifiers)
+    def click(*modifiers, scroll_to: :center)
```
jkotests, line 149: Is it inconsistent that we only apply the `:scroll_to` when modifiers are applied? `#click` without modifiers seems to be the only mouse action that doesn't centre the element first. Applying the `:scroll_to` would make `#obscured?` simpler too - ie could be made similar to the other click methods.

jkotests (changes requested):

lib/watir/elements/element.rb
jkotests: As discussed on Slack, we'll keep it this way for now. Always scrolling would generate a lot of additional commands.

titusfortner:
fixed & merged

PR #953, opened 2021-11-22
Shadow dom support

titusfortner:
This is a continuation of the code from #943

It's implemented with the Selenium methods, so right now it only works with Chrome v96.
Selenium shadow root is really no different from what you get back from execute script, so we could even rescue `rescue Selenium::WebDriver::Error::UnknownCommandError` and use the script to get a shadow root object.

The other major difference is that since a Shadow Root isn't an element, a lot of the methods in Element don't apply. As such, it doesn't really make sense to have it inherit from Element. So I've pulled out the methods applicable to ShadowRoot and Element and put them into a new module `SearchContext`.

Really I wish I already had a command executor in place, and I'm not sure how I feel about this code, hence the Draft PR.

### DIFF
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -99,5 +99,6 @@ require 'watir/elements/text_area'
 require 'watir/elements/text_field'
 require 'watir/elements/input'
 require 'watir/radio_set'
+require 'watir/shadow_root'
 
 require 'watir/aliases'
--- a/lib/watir.rb
+++ b/lib/watir.rb
@@ -1,6 +1,7 @@
 require 'selenium-webdriver'
 require 'time'
 
+require 'watir/search_context'
 require 'watir/scroll'
 require 'watir/wait'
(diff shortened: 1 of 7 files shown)


jkotests (approved):
I think this is a good first pass. It solves people's immediate problem of needing to work with shadow roots. We can make future improvements, particularly the locator performance, in future releases.

tobytang1987:
Is there any update on this? Agree with @jkotests that this could be the first version to support shadow roots requirements.

titusfortner:
The issue is that Chrome shadow root support doesn't work with XPath. There's more code that needs to be added here, but I've been more focused on Selenium code than watir recently. Hopefully soon.

titusfortner:
merged - 01a071116cb63a8ec3b8855876a57c7a7ceac835 & e7c80ffb26b58a7c315b776a2bf754f9d952122e

PR #962, opened 2022-08-01
Reduce rubocop offenses

ivodimitrov:
The aim of the PR is to bump rubocop version and to reduce to minimum rubocop offenses.

### DIFF
--- a/lib/watir/capabilities.rb
+++ b/lib/watir/capabilities.rb
@@ -16,7 +16,7 @@ module Watir
       end
       raise(ArgumentError, ':url and :service are not both allowed') if @options.key?(:service) && @options.key?(:url)
 
-      @browser = browser.nil? && infer_browser || browser
+      @browser = (browser.nil? && infer_browser) || browser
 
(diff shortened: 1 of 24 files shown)


titusfortner:
Looks like this broke the unit tests? Can you double check? Thanks.

ivodimitrov:
@titusfortner I've:
- fix one selector 
- left the offense about `ENV['SAUCE_USERNAME']` 

> Style/FetchEnvVar: Use ENV.fetch('SAUCE_USERNAME') or ENV.fetch('SAUCE_USERNAME', nil) instead of ENV['SAUCE_USERNAME'] 

so now all unit tests passed.

Now:

> 229 files inspected, 11 offenses detected, 7 offenses autocorrectable

We should discuss detected offenses one by one, as well as `.rubocop_todo.yml` and they could be fixed in another PR.

Could you please review thе PR again. 
Thank you in advance!

titusfortner:
merged, thanks - 255caf9b09414bcebb5b92ad57401363530fdebe

PR #968, opened 2023-03-15
Cookie: add httpOnly and SameSite attributes

### LINKED ISSUE #967: Cookie httpOnly
[Cookie add method](https://rubydoc.info/gems/watir/Watir/Cookies#add-instance_method) as an option `secure` that I believe will set the `secure` attribute on the cookie, but there is no option to control the `httpOnly` attribute.

Ref. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

noraj:
fix #967

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,7 +19,7 @@ Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
 Metrics/AbcSize:
-  Max: 22
+  Max: 23
   Exclude:
     - 'lib/watir/locators/element/selector_builder.rb'
     - 'lib/watir/locators/element/selector_builder/*.rb'
--- a/lib/watir/cookies.rb
+++ b/lib/watir/cookies.rb
@@ -43,16 +43,19 @@ module Watir
     # Adds new cookie.
     #
     # @example
-    #   browser.cookies.add 'my_session', 'BAh7B0kiD3Nlc3Npb25faWQGOgZFRkk', secure: true
+    #   browser.cookies.add 'my_session', 'BAh7B0kiD3Nlc3Npb25faWQGOgZFRkk', secure: true, http_only: true
     #
     # @param [String] name
     # @param [String] value
     # @param [Hash] opts
     # @option opts [Boolean] :secure
+    # @option opts [Boolean] :http_only
     # @option opts [String] :path
+    # @option opts [String] :same_site
     # @option opts [Time, DateTime, NilClass] :expires
     # @option opts [String] :domain
     #
+    # @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
 
     def add(name, value, opts = {})
       cookie = {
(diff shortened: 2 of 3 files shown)


titusfortner:
Awesome, thanks! Since it is `httpOnly` with a camel case, we need to use `http_only` so it will be converted. Same thing with `sameSite`, which is kind of weird, but /shrug (https://w3c.github.io/webdriver/#cookies)

titusfortner:
Also would be appreciated if we add a test for it to make sure it is actually working in the browser (would catch the syntax issue).

noraj:
> Awesome, thanks! Since it is `httpOnly` with a camel case, we need to use `http_only` so it will be converted. Same thing with `sameSite`, which is kind of weird, but /shrug (https://w3c.github.io/webdriver/#cookies)

Done ✅ 

> Also would be appreciated if we add a test for it to make sure it is actually working in the browser (would catch the syntax issue).

I usually write test with minitest and I'm not familiar with rspec.

I'm not sure how to update https://github.com/watir/watir/blob/1dbb3c77b1bf8ae0fc98aec7f5896589c226d9ed/spec/watirspec/cookies_spec.rb#L99-L110

or add test for http_only and same_site.

titusfortner:
Hmm, we probably don't need separate tests for each parameters, but can essentially copy the [path test](https://github.com/watir/watir/blob/main/spec/watirspec/cookies_spec.rb#L66) for the two parameters; we just want to make sure that when we add the cookie for Chrome/Firefox, the actual cookies have the values we set.

This is what is done in Selenium: https://github.com/SeleniumHQ/selenium/blob/trunk/rb/spec/integration/selenium/webdriver/manager_spec.rb#L30

oh, hmm, I wonder if the reason the `samesite` test is guarded is because it should be `same_site`... That would be interesting. If we show it works in Watir, I'll have to get it fixed in Selenium. 😆 

Thanks for your help.

noraj:
## HttpOnly

It should look like this but I don't know how to simulate JS access:

```ruby
    it 'is not visible to javascript when http_only is true',
      browser.goto WatirSpec.url_for 'index.html'

      options = {http_only: true}

      browser.cookies.add 'httponly', 'true', options
      cookie = browser.cookies.to_a.find { |e| e[:name] == 'httponly' }
      
      # add a test to check it equals empty when access via JavaScript (document.cookie)
      expect(cookie[:name]).to eq 'httponly'
      expect(cookie[:value]).to eq true
    end
```

cf. https://github.com/SeleniumHQ/selenium/blob/5374cf84b74ae36758b9022b5f229e7311b776a3/rb/spec/integration/selenium/webdriver/manager_spec.rb#L114C13-L121

## Secure

It is currently tested only over HTTP

https://github.com/watir/watir/blob/1dbb3c77b1bf8ae0fc98aec7f5896589c226d9ed/spec/watirspec/cookies_spec.rb#L98-L110

but it should be tested over HTTPS too cf. https://github.com/SeleniumHQ/selenium/blob/5374cf84b74ae36758b9022b5f229e7311b776a3/rb/spec/integration/selenium/webdriver/manager_spec.rb#L123-L145

## SameSite

I will be harder to test since there are 4-5 cases to test.

cf. https://github.com/SeleniumHQ/selenium/blob/5374cf84b74ae36758b9022b5f229e7311b776a3/rb/spec/integration/selenium/webdriver/manager_spec.rb#L147-L187 and https://github.com/SeleniumHQ/selenium/blob/5374cf84b74ae36758b9022b5f229e7311b776a3/rb/spec/integration/selenium/webdriver/manager_spec.rb#L42-L52

## Help wanted

I never used RSpec and I don't know Watir internals required to test the complex cases showcased here. So if someone would help me getting the right documentation pointers or submit unit tests for me, it would be greatly appreciated.

titusfortner:
Rebased off main, added tests, changed to snake case, fixed linting failure

titusfortner:
Most of the test failures are window minimization weirdness and a test that passes and fails depending on the browser version on the OS, which is apparently different right now. It'll all work fine next week. 😂

PR #969, opened 2023-04-17
DateField and DateTimeField - Accept inputs that behave like Date

jkotests:
Setting a `DateField` or `DateTimeField` currently requires a `String`, `Date` or `Time`. This can be too strict. 

For example, it prevents using ActiveSupport with time zones:

~~~~
require 'active_support/time'

Time.zone = 'Stockholm'
browser.date_field.set(2.months.ago)
#=> ArgumentError (DateField#set only accepts instances of Date or Time)
~~~~

This PR proposes accepting any object that responds to `#strftime`.

### DIFF
--- a/lib/watir/elements/date_field.rb
+++ b/lib/watir/elements/date_field.rb
@@ -9,8 +9,8 @@ module Watir
     def set!(date)
       date = Date.parse date if date.is_a?(String)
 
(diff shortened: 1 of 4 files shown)


titusfortner (approved):
Agreed. Thanks.

